PR #39, opened 2026-01-17
Surface update status controls

OmarMcAdam:
## Summary
- add a sidebar version badge that flags when updates are available
- rework the Settings Updates panel into a compact header with expandable details

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4179,7 +4179,7 @@ export default function App() {
               <div class="pt-2 flex flex-wrap gap-2">
                 <Button variant="secondary" onClick={() => setDeveloperMode((v) => !v)}>
                   <Shield size={16} />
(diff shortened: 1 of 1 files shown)


benjaminshafii:
Automated review (bot): reviewed the diff (draft PR).

Nice direction for a more visible update surface. A couple notes from the patch:
- The diff appears to introduce a new Updates card and also modify an existing Updates section; make sure the final UI doesn’t end up with duplicated Updates blocks.
- `refreshUpdates()` sets `updateRefreshStatus` to success/error but doesn’t currently clear it on all paths (e.g. after “available” it returns null) — worth verifying the UX state machine.
- Consider extracting the Updates UI into a small component to reduce the App.tsx size/merge conflicts (App.tsx is already a hotspot).

Not related to #121 directly, but it touches settings UX so keeping merge conflicts low would help.

Issue #61, opened 2026-01-18
Session navigation breaks after leaving and re-entering


benjaminshafii:
## Problem
After opening a task, going back, then selecting another task, the session does not load correctly.

## Steps
1) Open any session.
2) Go back to dashboard.
3) Select another session from recents.

## Expected
Selected session loads its messages and state consistently every time.

## Notes
Likely selection and view state desync.

benjaminshafii:
Automated note (bot): there is an open PR that addresses this exact race.

PR #111 adds abort guards so that async loads from Session A can’t overwrite Session B if the selection changes mid-flight. That’s the right fix for the described repro.

If you’re reviewing/merging: please sanity-check that aborting doesn’t leave stale loading state behind.

benjaminshafii:
Automated message: tracking this — likely a session selection/state sync bug. We should reset session state and force a reload on re-entry to avoid stale view state. If you hit this on the latest build, please confirm version and steps so we can verify the fix.

benjaminshafii:
Thanks for the report — I’m a bot. Please share your OpenWork version, OS, and whether this repros on a fresh workspace. If you can, include a short screen recording and diagnostics logs.

benjaminshafii:
Thanks for the report — I am a bot. To move this forward, can you share: 1) OpenWork version, 2) OS + architecture, 3) exact steps and expected vs actual, and 4) any logs or screenshots? If possible, include the last 100 lines of app logs.

PR #111, opened 2026-01-19
#61

apoorvgarewal07:
The Problem:
When you open a task → go back → select another task quickly, the session doesn't load correctly because:

1.First selectsession(taskA) starts async loads (messages, todos, permissions) 
2.You switch to SelectSession(taskB).
3.But taskA's pending async calls finish and overwrite taskB's state with taskA's data → incorrect session displays
 
The Solution :
I added 4 abort guards at critical checkpoints that check if the selected session changed

### DIFF
--- a/src/app/session.ts
+++ b/src/app/session.ts
@@ -132,10 +132,20 @@ export function createSessionStore(options: {
     mark("calling session.messages");
     const msgs = unwrap(await withTimeout(c.session.messages({ sessionID }), 12_000, "session.messages"));
     mark("session.messages done");
+    // If the selected session changed while we were waiting, abort to avoid
+    // overwriting a newer selection's state.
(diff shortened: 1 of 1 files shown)


benjaminshafii:
Automated review (bot): looked through the diff.

The abort guards are the right fix for the race you described (async loads for A finishing after switching to B). The checkpoints are placed in sensible spots.

Minor suggestions:
- Consider using a single monotonically increasing “selection token” (run id) to reduce repeated `if (selectedSessionId() !== sessionID)` blocks.
- Verify that aborting doesn’t leave stale `busy`/loading indicators (if any) in a partially-loaded state.

Otherwise looks good.

benjaminshafii (approved):

benjaminshafii:
thx @apoorvgarewal07 merged!

PR #116, opened 2026-01-19
Fix workspace_guide skill loading issue during onboarding

### LINKED ISSUE #75: Error: workspace_guide that doesn't exist
Image

The Error:
The system is trying to load a skill called workspace_guide that doesn't exist. You can see the sequence:

The user (or system) requests to "Load the workspace_guide skill..."
The app responds: "Loading the workspace_guide skill to summarize local vs global contents..."
Then it hits the error: "Workspace guide skill not found"
The app attempts a fallback: "I'll scan for skills, plugins, and templates to summarize local versus global content"
Finally it acknowledges the failure directly: "I tried to load the workspace_guide skill, but it isn't present in this workspace"

What's Happening:
The onboarding flow appears to be hardcoded to look for a workspace_guide skill that should provide a welcoming overview, but that skill either:

Was never created/included in this workspace
Was removed or renamed
Is expected to be installed separately but wasn't

The app does gracefully degrade by providing a generic "Welcome snapshot" instead, but the error messages are visible to the user, which makes the onboarding feel broken rather than smooth.
The Fix Would Likely Involve:
Either ensuring the workspace_guide skill is bundled with the workspace by default, or updating the onboarding logic to check if the skill exists before attempting to load it (avoiding the visible error state).

opencolin:
This PR fixes the issue where the onboarding flow would fail with 'workspace_guide that doesn't exist' error.

## Changes
- Added retry logic and delay for skill loading in welcome session creation
- Gracefully handle missing workspace_guide skill with fallback prompt
- Improved error handling to prevent onboarding failures

## Problem
The onboarding flow was trying to use the workspace_guide skill immediately after workspace creation, but there was a timing issue where the skill wasn't yet indexed by the OpenCode engine.

## Solution
- Added 1-second delay before creating welcome session
- Implemented retry logic (up to 3 attempts) when loading workspace_guide skill  
- Added fallback prompt that works without skill reference if retries fail
- Better error handling that doesn't break the entire onboarding flow

Fixes #75

### DIFF
--- a/src/app/workspace.ts
+++ b/src/app/workspace.ts
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
           }
 
           if (!already) {
+            // Wait a bit for skills to be indexed before creating welcome session
+            await new Promise(resolve => setTimeout(resolve, 1000));
+
             const session = unwrap(
               await nextClient.session.create({ directory: wsRoot, title: "Welcome to OpenWork" }),
             );
-            await nextClient.session.promptAsync({
-              directory: wsRoot,
-              sessionID: session.id,
-              model: options.defaultModel(),
-              variant: options.modelVariant() ?? undefined,
-              parts: [
-                {
-                  type: "text",
-                  text:
-                    "Load the `workspace_guide` skill from this workspace and give a short, welcoming overview of what lives here (skills, plugins, templates) versus what's global. Avoid CLI language or raw file paths. Then suggest two friendly next actions to try inside OpenWork.",
-                },
-              ],
-            });
+
+            // Try to create the welcome prompt with retry logic for skill loading
+            let promptSuccess = false;
+            let retryCount = 0;
+            const maxRetries = 3;
+
+            while (!promptSuccess && retryCount < maxRetries) {
+              try {
+                await nextClient.session.promptAsync({
+                  directory: wsRoot,
+                  sessionID: session.id,
+                  model: options.defaultModel(),
+                  variant: options.modelVariant() ?? undefined,
+                  parts: [
+                    {
+                      type: "text",
+                      text:
+                        "Load the `workspace_guide` skill from this workspace and give a short, welcoming overview of what lives here (skills, plugins, templates) versus what's global. Avoid CLI language or raw file paths. Then suggest two friendly next actions to try inside OpenWork.",
+                    },
+                  ],
+                });
+                promptSuccess = true;
+              } catch (e) {
+                retryCount++;
+                if (retryCount < maxRetries) {
+                  // Wait before retrying
+                  await new Promise(resolve => setTimeout(resolve, 2000));
+                } else {
+                  // On final failure, try without the skill reference
+                  try {
+                    await nextClient.session.promptAsync({
+                      directory: wsRoot,
+                      sessionID: session.id,
+                      model: options.defaultModel(),
+                      variant: options.modelVariant() ?? undefined,
+                      parts: [
+                        {
+                          type: "text",
+                          text:
+                            "Give a short, welcoming overview of this OpenWork workspace. Explain what lives here (skills, plugins, templates) versus what's global. Avoid CLI language or raw file paths. Then suggest two friendly next actions to try inside OpenWork.",
+                        },
+                      ],
+                    });
+                    promptSuccess = true;
+                  } catch (fallbackError) {
+                    // If both attempts fail, just continue without the welcome session
+                    console.warn("Failed to create welcome session:", fallbackError);
+                  }
+                }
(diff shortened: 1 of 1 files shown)


src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
           }
 
           if (!already) {
+            // Wait a bit for skills to be indexed before creating welcome session
+            await new Promise(resolve => setTimeout(resolve, 1000));
+
             const session = unwrap(
               await nextClient.session.create({ directory: wsRoot, title: "Welcome to OpenWork" }),
             );
-            await nextClient.session.promptAsync({
```
Corbkali, line 257: await nextClient.session.promptAsync({

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-              directory: wsRoot,
```
Corbkali, line 258: directory: wsRoot,

Corbkali: sessionID: session.id,

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-              sessionID: session.id,
-              model: options.defaultModel(),
```
Corbkali, line 260: model: options.defaultModel(),

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-              variant: options.modelVariant() ?? undefined,
```
Corbkali, line 261: variant: options.modelVariant() ?? undefined,

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-              parts: [
```
Corbkali, line 262: parts: [

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-                {
```
Corbkali, line 263: {

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-                  type: "text",
```
Corbkali, line 264: type: "text",

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-                  text:
```
Corbkali, line 265: text:

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-                    "Load the `workspace_guide` skill from this workspace and give a short, welcoming overview of what lives here (skills, plugins, templates) versus what's global. Avoid CLI language or raw file paths. Then suggest two friendly next actions to try inside OpenWork.",
```
Corbkali, line 266: "Load the `workspace_guide` skill from this workspace and give a 
short, welcoming overview ...",

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-                },
```
Corbkali, line 267: },

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-              ],
```
Corbkali, line 268: ],

src/app/workspace.ts
```diff
@@ -251,22 +251,63 @@ export function createWorkspaceStore(options: {
...
-            });
```
Corbkali, line 269: });

benjaminshafii:
Automated review (bot): looked through the diff.

This fix makes sense directionally: onboarding can race skill indexing, so a delay + retry + fallback prompt should prevent the whole flow from breaking.

A couple suggestions before merge:
- The fixed `setTimeout(1000)` + retry loop is pragmatic, but it would be more robust to retry based on an explicit condition (e.g. `refreshSkills({force:true})` then check for `workspace_guide` existence) rather than sleeping.
- The retries currently swallow the root error; consider capturing the last error string and surfacing it in developer mode (or at least store it for debugging).
- `console.warn(...)` is fine for dev, but if you have a centralized logger, prefer that (or gate the warn behind developer mode).

This is adjacent to the broader startup/first-run reliability issues tracked in #121.

kmk142789420-stack (approved):

PR #129, opened 2026-01-20
Clarify docs references and add contributing notes

tersePrompts:


benjaminshafii:
thank you @tersePrompts

Issue #172, opened 2026-01-21
[Bug report] if there is no opencode, it will create opencode stub file, then user install opencode, it will always error.


caiqinghua:
1. there is no opencode
2. pnpm dev
3. install opencode
4. always error

<img width="572" height="777" alt="Image" src="https://github.com/user-attachments/assets/1b1b092f-c08f-44a8-9b4c-2f981d53da53" />

<img width="1172" height="807" alt="Image" src="https://github.com/user-attachments/assets/f02e24bc-fd01-47fb-b619-44c43321a803" />

Windswithyou:
Maybe install opencode firstly and then reboot openwork

caiqinghua:
@Windswithyou We should let user run openwork by 'pnpm dev' after user installed opencode

OmarMcAdam:
Is this when running dev or downloading and installing the built app from releases?

caiqinghua:
pnpm dev

caiqinghua:
Already make a PR

benjaminshafii:
Automated message: thanks for the repro. We’ve been removing the debug stub behavior so OpenWork won’t leave a fake `opencode` behind. If you can, please try the latest v0.3.5 (https://github.com/different-ai/openwork/releases/tag/v0.3.5) and confirm whether the error persists.

For local dev builds, deleting the generated stub under `packages/desktop/src-tauri/sidecars/opencode` (if present) and ensuring a real `opencode` is on PATH should clear the state.

benjaminshafii:
Thanks for the report — I’m a bot. Please share your OpenWork version, OS, and the exact error message (copy/paste). Also let us know where the stub file is created and the path to your opencode binary. Diagnostics logs would help too.

benjaminshafii:
Thanks for the report — I am a bot. To move this forward, can you share: 1) OpenWork version, 2) OS + architecture, 3) exact steps and expected vs actual, and 4) any logs or screenshots? If possible, include the last 100 lines of app logs.

benjaminshafii:
Thanks — I am the OpenWork enterprise bot. Can you share OS, OpenWork version, the exact error text, and the path of the stub file that was created?

PR #173, opened 2026-01-21
fix issues #172 create opencode stub file confuse user

caiqinghua:
1. there is no opencode
2. pnpm dev error
3. install opencode
4. always error


benjaminshafii:
Tried to repro locally from a fresh worktree. With PATH stripped of opencode,  fails before build.rs runs because node_modules/tauri aren’t installed in this environment, so I couldn’t fully validate runtime behavior yet.

Reviewing the change set, I’d like a couple adjustments before merge:

1) Avoid mutating  in-place. The prepare/restore pair is brittle if  fails or is interrupted, leaving a dirty config. Prefer a release-only config (ex:  passed via ) or a single wrapper script that runs build and restores in a  so restore is guaranteed.

2) If we keep externalBin out of the default config for dev, keep the release config as the single source of truth (no backups in-tree). That avoids committing transient diffs or leaving a stale  file on failure.

3) The  change to PATH in dev looks right; for release, the hard error is ok, but please keep the error message user-facing (“install opencode or set OPENCODE_BIN_PATH”) and fail only in release builds.

Let me know if you want me to wire up a clean release config + build wrapper; I can take it from here.

benjaminshafii:
Follow-up (previous comment got garbled by shell quoting).

Tried to repro locally from a fresh worktree. With PATH stripped of opencode, pnpm -C packages/desktop dev fails before build.rs runs because node_modules/tauri aren't installed in this environment, so I couldn't fully validate runtime behavior yet.

Reviewing the change set, I'd like a couple adjustments before merge:

1) Avoid mutating tauri.conf.json in-place. The prepare/restore pair is brittle if tauri build fails or is interrupted, leaving a dirty config. Prefer a release-only config (ex: tauri.conf.release.json passed via tauri build --config ...) or a single wrapper script that runs build and restores in a try/finally so restore is guaranteed.

2) If we keep externalBin out of the default config for dev, keep the release config as the single source of truth (no backups in-tree). That avoids committing transient diffs or leaving a stale .backup file on failure.

3) The engineSource change to PATH in dev looks right; for release, the hard error is ok, but please keep the error message user-facing (install opencode or set OPENCODE_BIN_PATH) and fail only in release builds.

Let me know if you want me to wire up a clean release config + build wrapper; I can take it from here.

caiqinghua:
@benjaminshafii Yes, keeping tauri.conf.json immutable is better design. We should only remove create_debug_stub() to avoid confusion.

benjaminshafii:
Repro update: on this PR branch, with PATH excluding opencode but including cargo (~/.cargo/bin), `pnpm -C packages/desktop dev` builds and launches successfully (no sidecar warning anymore). The process just keeps running as expected; my command timed out after the app launched. So dev no longer hard-requires opencode, which is good.

I still think we should avoid in-place tauri.conf.json edits for release builds (same notes as before). If you want me to wire up a release-only config or safer wrapper, I can do it.

caiqinghua:
Yes, agree. I have removed tauri.conf.json edits.

caiqinghua:
@benjaminshafii ben, is there any other problem?

benjaminshafii:
Automated message: removing the debug stub is the right direction, but this now hard‑fails `build.rs` whenever OpenCode is missing—even in dev. That blocks local dev and conflicts with the “graceful degradation” goal.

Suggestion: only hard error in release builds; in dev log a warning and continue with PATH mode. Also this PR is currently CONFLICTING with `dev`, so please rebase.

OmarMcAdam:
Closing as inactive

PR #175, opened 2026-01-21
fix: add rpm to release-macos-aarch64.yml

AdianKozlica:
I realised I also didn't add it to this file, apologies for the mistake.

### DIFF
--- a/.github/workflows/build-desktop.yml
+++ b/.github/workflows/build-desktop.yml
@@ -39,6 +39,7 @@ jobs:
             libsoup-3.0-dev \
             libwebkit2gtk-4.1-dev \
             libssl-dev \
+            rpm \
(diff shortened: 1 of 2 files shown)


benjaminshafii (commented):
Automated message: A couple checks before merge. 1) This switches the Linux target to aarch64 on ubuntu-22.04; do we already have aarch64 cross-compile/toolchain (or QEMU) set up? Without it, this will likely fail. 2) This replaces the x86_64 Linux artifact in this workflow; if there isn’t another workflow producing x86_64 builds, we’ll lose them. If the goal is just adding rpm, consider keeping x86_64 and only adding --bundles deb,rpm; otherwise a dedicated aarch64 Linux workflow may be clearer.

Issue #174, opened 2026-01-21
[Bug report]New task will work in old workspace after user switch workspace


caiqinghua:
<img width="1075" height="472" alt="Image" src="https://github.com/user-attachments/assets/78869929-5417-40a7-92cf-11d3c8b83c77" />

OmarMcAdam:
Thanks for reporting, will address this

benjaminshafii:
Automated message: thanks — this looks like a workspace switch state sync issue. We should ensure session creation uses the active workspace root and resets session state after switching.

Can you share the exact steps and version (and whether it’s local or remote workspace)? We’re prioritizing workspace switch UX now, so this will help confirm if it’s fixed in the latest build.

nodfans:
> Automated message: thanks — this looks like a workspace switch state sync issue. We should ensure session creation uses the active workspace root and resets session state after switching.
> 
> Can you share the exact steps and version (and whether it’s local or remote workspace)? We’re prioritizing workspace switch UX now, so this will help confirm if it’s fixed in the latest build.

I also tried reproducing this but couldn't. The session list had sync issues a few days ago that required reloading to fix, but this appears to have been resolved in recent updates.

benjaminshafii:
Thanks for the report — I’m a bot. Please share your OpenWork version, OS, and exact steps to reproduce (including how you switch workspaces). If you can, include a short screen recording and diagnostics logs.

benjaminshafii:
Thanks for the report — I am a bot. To move this forward, can you share: 1) OpenWork version, 2) OS + architecture, 3) exact steps and expected vs actual, and 4) any logs or screenshots? If possible, include the last 100 lines of app logs.

benjaminshafii:
Thanks — I am the OpenWork enterprise bot. Can you share exact steps to reproduce (including the workspace switch sequence), OpenWork version, and any logs?

PR #177, opened 2026-01-21
restart engine when user switch workspace. fix issuse #174

caiqinghua:
截屏2026-01-21 21 22 42


benjaminshafii:
thx @caiqinghua

benjaminshafii:
Automated message: thanks for the contribution! Let's get to the core of the change.

Core fix: restarting the engine on workspace switch in host mode makes sense because the engine is tied to a single projectDir; the reconnect + busy/error handling in `packages/desktop/src/app/context/workspace.ts` looks solid.

Simplification: the release config scripts feel heavier than needed. In opencode (see `vendor/opencode/packages/desktop/src-tauri/tauri.conf.json`), `bundle.externalBin` stays in the config and the sidecar is handled by copy scripts; no in-place config edits. Consider either:
- keep `externalBin` always in `tauri.conf.json` and rely on the `build.rs` release-only copy guard, then drop `prepare-release.mjs`/`restore-config.mjs` and revert the `build` script, or
- use a separate release config file (`tauri build --config src-tauri/tauri.release.conf.json`) instead of edit/restore.

Also, as written, `node scripts/restore-config.mjs` only runs if `tauri build` succeeds (because of `&&`), so a failed build can leave the config modified.

caiqinghua:
@benjaminshafii Only keep restart engine when switch workspace.

benjaminshafii:
@caiqinghua having a look at it now

PR #183, opened 2026-01-22
jump check of debug

Windswithyou:
In windows, it will use opencode.cmd in debug folder. And it will fail to start opencode. I remove the "target/debug" from the path.
On Windows, sidecar is not used - opencode is found via PATH at runtime. Create a minimal stub to satisfy Tauri's externalBin requirement

### DIFF
--- a/packages/desktop/src-tauri/build.rs
+++ b/packages/desktop/src-tauri/build.rs
@@ -30,6 +30,19 @@ fn ensure_opencode_sidecar() {
   }
   let dest_path = sidecar_dir.join(file_name);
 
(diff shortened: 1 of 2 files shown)


benjaminshafii:
Automated message: I’m concerned about writing a minimal PE stub in `build.rs` on Windows. It’s a hack that can ship a fake binary and is hard to reason about. Prefer a config-level fix (skip `externalBin` on Windows or use a real sidecar) and avoid generating executables in build scripts.

Also, `find_in_path('opencode')` may miss installs that only expose `opencode.exe` on Windows—can we keep the `.exe` lookup?

Mergeable, but I’d prefer these adjustments first.

PR #189, opened 2026-01-22
Harden installer and path handling

Golenspade:
## Summary
- Pipe the installer script into bash stdin to avoid shell pipelines and surface curl errors.
- Add workspace state file locking plus safer environment/config path handling.
- Canonicalize OpenCode binaries/sidecar paths and add skill name length guardrails.

## Testing
- `pnpm --filter @different-ai/openwork build:web`
- `pnpm --filter @different-ai/openwork test:e2e`
- `cargo check --manifest-path packages/desktop/src-tauri/Cargo.toml --locked`
- `cargo test --manifest-path packages/desktop/src-tauri/Cargo.toml --locked`

### DIFF
--- a/packages/desktop/src-tauri/Cargo.lock
+++ b/packages/desktop/src-tauri/Cargo.lock
@@ -989,6 +989,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs2"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9564fc758e15025b46aa6643b1b77d047d1a56a1aea6e01002ac0c7026876213"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "futf"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


benjaminshafii:
Automated message: the hardening changes look good and align with the security/least‑privilege goals (file locks, canonical paths, symlink guard, safer installer piping).

Two notes:
- PR is currently CONFLICTING; please rebase on `dev`.
- Optional: consider adding a short failure message if `bash` exits non‑zero so users see the installer error quickly (you already capture curl stderr, which is great).

Happy to re-review once rebased.

benjaminshafii:
@Golenspade thx.

we made a few changes over the last days to make sidecar the default path so we can control the experience a bit more. we're still giving PATH as an escape hatch.

What does this PR intend to bring to the experience?

Golenspade:
@benjaminshafii  thx for the review:Intent is UX hardening, not a path-precedence change. This PR makes install + binary resolution more deterministic and safer: we canonicalize candidate binaries, avoid symlink overwrites, validate env-derived paths, and lock workspace state to prevent corruption. The user impact is fewer mysterious install failures or “wrong binary” issues, while still keeping PATH as an escape hatch. The new sidecar-default policy you mentioned is compatible; this just adds guardrails around whatever path wins.

Golenspade:
@benjaminshafii  rebased and added the bash exit error alert

OmarMcAdam:
Closing as inactive

PR #191, opened 2026-01-22
Add session commands, auth picker, and completion

Golenspade:
## Summary
- Add `/model`, `/connect`, `/auth`, `/agent`, `/save`, and `/rename` command support in the session prompt.
- Provide a provider auth picker modal to avoid guessing provider IDs.
- Add command autocomplete with arrow/Tab navigation and Enter-to-run.
- Sort sessions by activity and support session renaming.

## Testing
- `pnpm --filter @different-ai/openwork typecheck`
- `pnpm --filter @different-ai/openwork test:e2e`

### DIFF
--- a/packages/app/src/app/app.tsx
+++ b/packages/app/src/app/app.tsx
@@ -195,6 +195,7 @@ export default function App() {
     loadSessions,
     refreshPendingPermissions,
     selectSession,
+    renameSession,
     respondPermission,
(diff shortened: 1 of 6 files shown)


benjaminshafii:
@Golenspade wow that looks great. Will look a bit deeper and test and report back. But really appreciate the direction.

PR #195, opened 2026-01-22
fix: improve chat input UX and IME composition handling

JiahaoZhu11:
## Summary

This PR improves the chat input component with two main enhancements:

1. **Better UX for the input bubble** - The entire input bubble now behaves as a cohesive input component
2. **Fixed IME composition handling** - Enter key now works correctly for Chinese/Japanese/Korean input

### Chat Input UX
- Entire input bubble is now clickable to focus the textarea (not just the textarea itself)
- Added `cursor-text` cursor style to indicate the bubble is interactive
- Send button no longer shows tooltip or responds to hover when hidden (empty input)
- Removed focus outline from textarea since the outer bubble already shows focus state via `focus-within`

### IME Composition Fix
- Added `compositionstart`/`compositionend` event handlers to track IME state
- Implemented `skipNextEnter` flag to differentiate between:
  - Enter to confirm character selection (should not send)
  - Enter to send message (should send)
- Fixes the issue where pressing Enter during IME composition would send the incomplete text instead of confirming the selected characters

## Testing

- [ ] Click anywhere on input bubble → textarea focuses
- [ ] Empty input → send button hidden, no tooltip on hover
- [ ] With text → send button visible, tooltip shows on hover
- [ ] Type Chinese/Japanese/Korean with IME → Enter confirms selection
- [ ] After confirming IME selection → Enter sends message
- [ ] Click to select IME candidate → next Enter sends message correctly

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@different-ai/openwork-ui",
   "private": true,
-  "version": "0.3.4",
+  "version": "0.3.5",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@different-ai/openwork",
   "private": true,
-  "version": "0.3.4",
+  "version": "0.3.5",
   "type": "module",
   "scripts": {
     "dev": "tauri dev",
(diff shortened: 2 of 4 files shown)


Golenspade:
can you deliver a snapshot,so we can align the visual design

benjaminshafii:
Automated message: IME handling + click-to-focus UX looks good and matches the polished input behavior we want.

Two items:
- PR is currently CONFLICTING with `dev`; please rebase.
- Please double-check `skipNextEnter` doesn’t swallow a real Enter when there was no composition event (should be covered, but worth a quick test with IME + normal typing).

I can’t provide a snapshot from here; a short gif/screenshot of the updated input would help design review.

Issue #232, opened 2026-01-24
PKGBUILD does not build


l3afyb0y:
I am working on a PKGBUILD because the current one doesn't build. I'm working out the issue, it's going to take a minute.
It's an Arch-Moment™ for sure, so I gotta read the wiki some and possibly spin up a new VM. 
I'll close this issue when I've submitted PR, but want to make sure it's known of.

l3afyb0y:
> I am working on a PKGBUILD because the current one doesn't build. I'm working out the issue, it's going to take a minute. It's an Arch-Moment™ for sure, so I gotta read the wiki some and possibly spin up a new VM. I'll close this issue when I've submitted PR, but want to make sure it's known of.

I need to chroot which could take some time, for whatever reason my chroot might be severely out of date. I gotta rebuild and retest this. Even if it's not broken, it's not clean.

benjaminshafii:
hey @l3afyb0y keep me posted!

l3afyb0y:
> hey [@l3afyb0y](https://github.com/l3afyb0y) keep me posted!

Will do. I'm gathering some info while this compiles. Arch has tools for this, so it's just a matter of running them until I hit something. The one I'm waiting for now is final testing, and unfortunately needs to compile the whole of the systems packages. Could be another 15-30 mins until it's done. But if all goes well, or even if it doesn't, I'll know for certain why it's not building on my system correctly. 
I think I have it figured out but don't know until I make sure. Last I checked AUR should be working, and they should be nearly identical. Playing it safe a little since I've had this installed before, it's possible it's building off caches when I run yay to test AUR.
The issues I'm having don't make perfect sense to me yet, but I'll get it shortly. 
Rather safe than sorry than in a month when someone else finds it lol.

In the meantime, I'm writing some scripts that'll make AUR/PKGBUILD management easier on your side as well as mine. More yours than mine, you'll get them later with explanation on how to use since it's a neat GitHub feature. My biggest issue was never bothering to learn bash and have spent too long on fish at this point. but PKGBUILD's are bash. It's like just different enough that I get a little confused.
I'll update again soon.

l3afyb0y:
Fixed by PR #239

benjaminshafii:
thx @l3afyb0y you're killing it!

benjaminshafii:
[...]

PR #239, opened 2026-01-24
fix: Cleaned PKGBUILD and fixed installation. Changes to README for Arch installs.

l3afyb0y:
In #232 I noticed some things about the adopted PKGBUILD, and upstream's and that it wasn't installing.
In the interest of allowing it to be seen how PKGBUILD works if anyone is ever troubleshooting, I sat and attempted to debug it.
The repo's PKGBUILD took me some digging in the wiki, and a lot of "what?"s to figure out, but I think I've got it.

- Changed ${url} to be correct.
- Fixed an Arch-Moment™ where makepkg runs in the wrong directory and hits permission issues
- Fixed an Arch-Moment™ where makepkg breaks install by not running in root of compressed folder
- Cleaned up bsdtar by using flags piped to avoid build artifacts
- Left notes for future troubleshooters in PKGBUILD (not present in AUR)
- Changed README to include how to install (confirmed to work) on Arch systems.

For those who may be troubleshooting:
Arch Wiki is always the best source, as I'm sure Arch users know, so if something does break and I, or anyone else can't get to it,
referencing that is going to be the way to go for anyone looking to fix theirs in the future.
Changing the PKGBUILD if you need updates or a reinstall , or it is broken, or even if you just don't wanna git pull, using makepkg -si will always install the latest version and skip yay, allowing install if for some reason your personal checksum doesn't match AUR (AUR has been live for a bit now so it's possible). It will always ultimately pull from the Releases tab for stability as long as pkgver changes.
That should keep things stable for Arch users. Even if mistakes get made, we can install pretty consistently now regardless of inconsistencies.


benjaminshafii:
thx @l3afyb0y

l3afyb0y:
> thx @l3afyb0y 

No problem. @ me if there are more issues.

PR #241, opened 2026-01-24
fix: support all file extensions in artifact path parsing

nodfans:
## Problem
Hardcoded extension whitelist caused `.jsonc` files to be parsed as `.json`, breaking file paths in the file manager. (See screenshot)

Xnip2026-01-24_23-55-05

As shown above, the AI correctly identifies `opencode.jsonc`, but clicking "Open" fails with "Could not open artifact" because the path was incorrectly transformed to `opencode.json`.

## Solution
- Remove extension whitelist, support any valid extension (2-10 chars)
- Add path-pattern validation to avoid false positives
- Normalize paths for cross-platform compatibility

## Testing
Verified with 40+ test cases including `.jsonc`, `.yaml`, `.toml`, `.graphql`, etc.

Xnip2026-01-25_00-18-11


benjaminshafii:
thx @nodfans !

PR #253, opened 2026-01-25
added script to build, install and uninstall openwork on linux and up…

LanzelotSniper:
…dated .gitignore and README.md


OmarMcAdam:
Prefer to using exising build systems to build for linux instead of custom scripts

PR #254, opened 2026-01-25
fix: prevent auto-scroll when toggling view steps

nodfans:
## 🐛 Problem
Clicking "View steps" causes the page to unexpectedly jump to the bottom.

## 🔧 Changes
- Modified `createEffect` to use `on()` with explicit dependency tracking
- Added comparison between previous and current counts
- Only triggers scroll when counts increase (new content added)


benjaminshafii:
thx @nodfans

PR #348, opened 2026-01-31
Fix Windows sidecar initialization exiting early during beforeDevCommand

### LINKED ISSUE #270: Windows: sidecar engine fails to connect on first launch
## Summary
OpenWork should run without a pre-installed OpenCode CLI by using the bundled sidecar. On Windows, the engine appears to start but the sidecar never connects.

## Expected
- First launch uses bundled sidecar and connects to the engine without requiring a system OpenCode install.

## Actual
- On Windows, onboarding/advanced settings show `serve` appears to run, but starting OpenWork fails and the app never connects to the engine.

## Repro
1. Install OpenWork on Windows with no prior OpenCode install.
2. Launch OpenWork and proceed through onboarding.
3. Start OpenWork (host mode).
4. Observe connection failure despite `serve` appearing to run.

## Notes
- Goal: OpenWork should work out-of-the-box on Windows (sidecar should “just work”).
- This is blocking Windows users from connecting on first launch.

sameerthawait:
### Summary
Fixes a Windows-specific issue where `prepare-sidecar.mjs` exited early
during Tauri `beforeDevCommand`, causing sidecar initialization to be
incomplete.

### Root Cause
On Windows, calling `process.exit(0)` inside `beforeDevCommand` can cause
Tauri to treat the command as failed, preventing the sidecar from being
ready on first launch.

### Fix
Removed the early `process.exit(0)` and allowed the script to complete
naturally, ensuring proper sidecar setup.

### Verification
- `prepare-sidecar.mjs` completes successfully on Windows
- `beforeDevCommand` no longer fails
- Sidecar is correctly detected on first launch

Fixes #270


sameerthawait:
Thanks! Happy to make any changes if needed.

jcllobet:
@sameerthawait I believe this has been addressed on the latest releases from what we could test internally. Can you update to the latest version and let us know?

PR #405, opened 2026-02-03
Improve session catch-up cues

benjaminshafii:
## Summary
- add an empty state message in the session sidebar when there are no sessions
- surface unread message count and a jump-to-latest affordance when auto-scroll is off


benjaminshafii:
Screenshot: https://files.catbox.moe/fmt7d5.jpg

PR #406, opened 2026-02-03
Improve scheduled tasks empty state

benjaminshafii:
## Summary
- add a call-to-action button that opens Plugins when there are no scheduled tasks
- clarify workspace picker search states with empty and no-results messaging


benjaminshafii:
Screenshot: https://files.catbox.moe/uf8l9q.jpg

PR #424, opened 2026-02-03
Add auto-update flow for new app releases

robdefeo:
## Summary
- add Tauri updater command wiring and release-check logic
- integrate updater state into app/system state and UI surfaces
- expose update controls/status in Settings and related pages

### DIFF
--- a/packages/app/src/app/app.tsx
+++ b/packages/app/src/app/app.tsx
@@ -2043,6 +2043,8 @@ export default function App() {
     setPendingUpdate,
     updateEnv,
     setUpdateEnv,
+    updateChannel,
+    setUpdateChannel,
     checkForUpdates,
     downloadUpdate,
     installUpdateAndRestart,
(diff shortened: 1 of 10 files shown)


benjaminshafii:
**Automated review from OpenWork enterprise bot 🤖**

**Requested changes**
1. Add a timeout + fallback for prerelease checks (GitHub API rate limits/latency can stall the UI). If the prerelease API fails, fall back to stable or surface a fast error state.
2. Invalidate pending updates on channel change (or store the channel with the pending update and block mismatched downloads). This avoids downloading a prerelease after switching to stable.
3. Avoid holding the full update payload in memory. If the updater API only supports byte installs, add a hard cap/guard and error message; otherwise stream to disk or use the updater’s built-in install path.

**Testing request**
- Include a screenshot of Settings → Updates showing the channel toggle and update status (checking/downloading/ready or restart-required).
- Note how you tested the flow.

benjaminshafii:
Thanks for the @robdefeo

robdefeo:
I'm going to close this, I now see there is already auto update checking. Thanks for the review

PR #562, opened 2026-02-14
fix(app): prevent session store races

### LINKED ISSUE #537: [Bug]: Windows 11 is completely unusable.
### Summary

I've tried many times and even reinstalled it. OpenCode works fine, but OpenWorks just won't run.

### Steps to reproduce

Image

Image

### Expected behavior

normal run

### Actual behavior

I've tried many times and even reinstalled it. OpenCode works fine, but OpenWorks just won't run.

JaiDodson:
- Replace loadSessions/refreshSidebarWorkspaceSessions with local store removal in deleteSessionById to prevent reconcile() from wiping unrelated sessions
- Remove loadSessionsWithReady calls from sendPrompt to eliminate the same race condition on message send
- Inject new sessions directly into store and sidebar in createSessionAndOpen instead of relying on server round-trip
- Make applySessionAgent auto-create a session when none is selected
- Add shell:true to bun spawns in prepare-sidecar.mjs for Windows

fixes #537

### DIFF
--- a/packages/app/src/app/app.tsx
+++ b/packages/app/src/app/app.tsx
@@ -891,9 +891,6 @@ export default function App() {
           }),
         );
 
-        await loadSessionsWithReady(workspaceStore.activeWorkspaceRoot().trim()).catch(
-          () => undefined
-        );
       } else {
         const result = await c.session.promptAsync({
           sessionID,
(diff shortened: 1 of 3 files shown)


benjaminshafii:
great stuff @JaiDodson

PR #687, opened 2026-02-26
Fix: Introduced a throttle on downloads to fix the ui freeze

### LINKED ISSUE #609: [Bug]: Application becomes unresponsive when downloading new version
### Summary

I've notices quite a lot of times that I am unable to use openwork when i download a newer version from settings

### Steps to reproduce

update openwork version

### Expected behavior

to be able to use openwork when the software is downloading an update

### Actual behavior

unusable when updating the versions

saifxyzyz:
Closes #609 

This PR puts a throttle on the download chunks so that it doesn't overload the cpu and cause ui freeze

Created a throttle function to rate limit on the downloads(updates), so that the ui is responsive and the load on cpu is reduced


benjaminshafii:
thx @saifxyzyz

PR #703, opened 2026-03-01
chore(session): add compaction loop diagnostics

### LINKED ISSUE #700: Track upstream auto-compaction loop causing repeated synthetic continue turns
## Problem
Some reports describe OpenWork repeatedly emitting synthetic "Continue..." turns in long sessions. 
This may not be only a UI visibility issue; there is also a likely **core loop** path in OpenCode compaction.

## Upstream evidence
OpenCode issue:
- anomalyco/opencode#15533 — "Auto-compaction infinite loop when assistant ended its turn"

Code path (OpenCode):
- `packages/opencode/src/session/compaction.ts` injects synthetic continue when `result === "continue" && input.auto`.
- `packages/opencode/src/session/prompt.ts` immediately re-enters loop and can trigger compaction again on overflow.

Potential failure mode:
1. assistant ends naturally
2. overflow triggers auto-compaction
3. synthetic continue is injected
4. follow-up turn overflows again
5. compaction repeats

## Why this matters for OpenWork
OpenWork ships an embedded OpenCode sidecar (`packages/desktop/package.json` `opencodeVersion`), so runtime compaction behavior is inherited from OpenCode.

Even after fixing UI filtering of synthetic text, looping compaction can still waste tokens and degrade UX.

## Expected behavior
- Auto-compaction should resume only when continuation is actually needed.
- Sessions should not oscillate between compaction + synthetic continue without new user intent.

## Suggested actions
1. Track and prioritize upstream fix from anomalyco/opencode#15533.
2. Once merged, bump OpenWork `opencodeVersion` and validate long-session behavior.
3. Add OpenWork integration regression coverage:
   - long context near overflow
   - assistant ends naturally
   - verify no repeated synthetic continue loops
4. Add temporary diagnostics in OpenWork logs (compaction count per session / minute) to detect runaway loops early.

## Notes
OpenWork should still fix the app-side rendering leak separately (synthetic control text visibility), tracked in #699.

gh-xj:
## Summary
- add app-side diagnostics for synthetic compaction control turns in the session event pipeline
- track per-session synthetic "Continue..." control text frequency in a 60s rolling window
- emit warning-level perf logs when repeated synthetic continue events exceed a loop-suspect threshold
- clear per-session diagnostic state when a session is deleted

## Why
Issue #700 tracks an upstream auto-compaction loop condition. Until upstream runtime fixes land, OpenWork needs local observability to detect potential runaway patterns early.

This patch does not change prompt/runtime behavior. It only adds diagnostics for repeated synthetic continue control text.

Closes #700.

## What is logged
When a `message.part.updated` event contains a synthetic text part matching the compaction control message:
- `session.compaction:synthetic-continue`
  - includes `sessionID`, `messageID`, `partID`, `countPerMinute`, `windowMs`

If count crosses threshold (>=3 within 60s), throttled warnings are emitted:
- `session.compaction:synthetic-continue-loop-suspected`
  - includes `sessionID`, `countPerMinute`, `threshold`, `windowMs`

### Ran
- `pnpm typecheck`

### Result
- Fails due pre-existing repo/toolchain issues unrelated to this patch:
  - `packages/app/src/app/context/global-sdk.tsx` (TS1005/TS1109/TS1128)
  - `packages/app/src/app/lib/openwork-server.ts` (TS1005)
  - `packages/app/tsconfig.json` (TS6046)
  - warning: local `package.json` exists but `node_modules` missing

No direct diagnostics were reported in changed file:
- `packages/app/src/app/context/session.ts`

## Manual verification
1. Enable developer mode.
2. Reproduce a compaction-heavy long session.
3. Confirm `[OWPERF] session.compaction:synthetic-continue` appears when synthetic continue control text is emitted.
4. If repeated quickly, confirm `[OWPERF] session.compaction:synthetic-continue-loop-suspected` appears with per-minute counts.

### DIFF
--- a/packages/app/src/app/context/session.ts
+++ b/packages/app/src/app/context/session.ts
@@ -60,6 +60,12 @@ const sortSessionsByActivity = (list: Session[]) =>
       return a.id.localeCompare(b.id);
     });
 
+const SYNTHETIC_CONTINUE_CONTROL_PATTERN =
+  /^\s*continue if you have next steps,\s*or stop and ask for clarification if you are unsure how to proceed\.?\s*$/i;
+const COMPACTION_DIAGNOSTIC_WINDOW_MS = 60_000;
+const COMPACTION_LOOP_WARN_THRESHOLD = 3;
+const COMPACTION_LOOP_WARN_MIN_INTERVAL_MS = 10_000;
+
 const createPlaceholderMessage = (part: Part): PlaceholderAssistantMessage => ({
   id: part.messageID,
   sessionID: part.sessionID,
@@ -150,6 +156,8 @@ export function createSessionStore(options: {
   const [permissionReplyBusy, setPermissionReplyBusy] = createSignal(false);
   const reloadDetectionSet = new Set<string>();
   const invalidToolDetectionSet = new Set<string>();
+  const syntheticContinueEventTimesBySession = new Map<string, number[]>();
+  const syntheticContinueLoopLastWarnAtBySession = new Map<string, number>();
 
   const skillPathPattern = /[\\/]\.opencode[\\/](skill|skills)[\\/]/i;
   const skillNamePattern = /[\\/]\.opencode[\\/](?:skill|skills)[\\/]+([^\\/]+)/i;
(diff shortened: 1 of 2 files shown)


gh-xj:
Addressed Copilot review items in commit `124d135`:
- added hot-event rate limiting for `session.compaction:synthetic-continue`
- switched loop-threshold diagnostic to warning-level log path (`sessionWarn` / `console.warn`)
- resolved Copilot review threads

PR #739, opened 2026-03-04
fix(app): prevent blank transcript in virtualized sessions

benjaminshafii:
## Summary
- gate session virtualization on scroll-container readiness so long sessions do not enter virtual mode before the viewport is mounted
- add a defensive fallback: if virtual mode is active but no virtual rows are produced, render the standard message list instead of showing a blank transcript
- trigger virtualizer measurement after activation to stabilize row measurement on mount and session switches

## Validation
- `pnpm --filter @different-ai/openwork-ui build` ✅
- `pnpm --filter @different-ai/openwork-ui typecheck` ⚠️ fails on existing repo issue: `src/app/lib/local-file-path.ts(1,70): Could not find a declaration file for module './local-file-path.impl.js'`
- Manual Chrome MCP verification ✅
  - `ses_348e6379bffewp4yT3FIMjt5le` (USDC fees thread, long session) now renders transcript content
  - `ses_3492f7816ffeR17b2SvoaRzWMT` (scheduling thread) still renders as expected

### DIFF
--- a/packages/app/src/app/components/session/message-list.tsx
+++ b/packages/app/src/app/components/session/message-list.tsx
@@ -22,6 +22,7 @@ export type MessageListProps = {
   searchHighlightQuery?: string;
   workspaceRoot?: string;
   scrollElement?: () => HTMLElement | null | undefined;
+  scrollReady?: boolean;
   virtualizationThreshold?: number;
   footer?: JSX.Element;
 };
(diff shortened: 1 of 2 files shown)


benjaminshafii:
Pushed directly to  per request in-thread (commit 5c4ebf5b). Closing this PR.

PR #753, opened 2026-03-06
Fix PKGBUILD - Remove opencode binary if present

### LINKED ISSUE #288: PKGBUILD depends - Opencode
Someone on AUR left a comment that made me realize that this package will install the Opencode binary. I didn't notice because I used a curl command to download Opencode.
So my first question is, is Openwork meant to be compatible with Opencode, and meant to allow for both to be installed?
Then my second, are you making modifications to, or keeping updated with the Opencode official binary?
If not, it matters to change the dependencies or change the package(), if updates are out of sync in the repos my planned fix could potentially break openwork in the future.
If it does keep updated with Opencode, then I have a PKGBUILD ready for PR.

l3afyb0y:
Remove opencode binary if it is already installed during package installation to avoid pacman conflicts.

## Summary
- Fixes PKGBUILD conflicts when Opencode is already installed.
- Matches current AUR version.

## Why
- When Opencode binary is installed via pacman, it becomes 'owned' by a package which breaks installation due to Openwork shipping with the Opencode binary.

## Issue
- Closes #288 

## Scope
- Fix PKGBUILD

## Out of scope
- Update checksums and .SRCINFO,  I'm gonna leave that to upstream this time.

### Ran
- This is the same as the AUR PKGBUILD, it's been tested via:
```bash
yay -S
```

## Manual verification
1. Run either 
```bash
makepkg -si
```
or
```bash
yay -S openwork
```

## Evidence
image

## Risk
- N/A

## Rollback
- N/A

### DIFF
--- a/packaging/aur/PKGBUILD
+++ b/packaging/aur/PKGBUILD
@@ -23,7 +23,7 @@ package() {
 
   # Extract the internal data archive directly to the cwd.
   bsdtar -O -xf "${pkgname}-${pkgver}.deb" 'data.tar*' | bsdtar -C "${pkgdir}" -xf -
-    if [ -f /usr/bin/opencode ]; then rm -f "${pkgdir}/usr/bin/opencode"; fi;
+  rm -f "${pkgdir}/usr/bin/opencode"
 }
 
 # .deb Internal Structure Reference:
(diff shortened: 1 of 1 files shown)


benjaminshafii:
thx @l3afyb0y !

PR #796, opened 2026-03-08
feat(server): add builtin past-conversations skill

### LINKED ISSUE #792: Add default `past-conversations` skill for recalling previous chats
## Problem
OpenWork can’t reliably recall prior user conversations unless prompted with ad-hoc instructions. This hurts continuity and makes memory-dependent requests brittle.

## Proposal
Add a built-in default skill named `past-conversations` that is strongly prioritized when the user asks to recall previous conversations. Triggering should be heavily biased toward recall-style prompts (e.g. “what did we discuss before?”, “find our earlier chat about X”).

Suggested skill content:

```yaml
---
name: past-conversations
description: Find and inspect past conversations. Use when the user asks you to recall a previous conversation.
---
```

### Discovery of past conversations
All past conversations exist in the sqlite database at `~/.local/share/opencode/opencode.db`.

Use `sqlite3` and query the `part` table:

```sql
CREATE TABLE `part` (
  `id` text PRIMARY KEY,
  `message_id` text NOT NULL,
  `session_id` text NOT NULL,
  `time_created` integer NOT NULL,
  `time_updated` integer NOT NULL,
  `data` text NOT NULL
);
```

Only search rows where `data` JSON has `type = "text"`, and search within the `text` field. Sort by `time_created`, then group by `session_id` (conversation-level grouping).

Reference implementation notes: https://gist.github.com/jlongster/059741f0d2384eb9d41836dfffcb9cc7

## Acceptance Criteria
- `past-conversations` exists as a first-class default skill.
- Skill trigger routing is biased toward recall/introspection prompts by default.
- Skill queries only text message parts (`type: "text"`) from `part.data`.
- Results are ordered by `time_created` and grouped by `session_id`.
- Output makes it easy to inspect relevant conversation snippets and sessions.
- Add test coverage for trigger routing and query behavior.

caoergou:
## Summary

Add a builtin `past-conversations` skill that allows the AI to recall previous conversations from the OpenCode SQLite database.

## Changes

- Add `packages/server/src/builtin-skills/past-conversations/SKILL.md`
- Modify `listSkills()` to include builtin skills with fallback mechanism
- Add `getSkillContent()` for reading skill content from filesystem or embedded module
- Add `script/generate-builtin-skills.ts` for single-file executable support
- Add `script/copy-builtin-skills.ts` for npm package distribution
- Add trigger phrases in skill description for AI routing
- Add test coverage for builtin skills functionality

### Build Process

Two build modes are supported:

1. **Normal build (`bun run build`)**: Uses `tsc` + copy script to include SKILL.md files in dist/
2. **Single-file executable (`bun run build:bin`)**: Generates TypeScript module with embedded skill content

### Single Source of Truth

All skill content lives in `SKILL.md` files. The generated TypeScript module is created during build and should not be committed to git.

## Test Plan

- [x] All 115 tests pass
- [x] TypeScript typecheck passes
- [x] Build scripts work correctly
- [ ] Manual testing: skill appears in skills list
- [ ] Manual testing: skill can query past conversations

Closes #792


caoergou:
Hi, just checking in on this PR — it's been about 11 days since submission.

Happy to address any feedback or make adjustments if the approach doesn't align with the project direction. Let me know if there's anything blocking review!

jcllobet:
Hey @caoergou , it's an interesting take to memory / recall, but we're not set on introducing this yet as we're in the middle of a larger redesign. 

Do you mind if we circle back on it in a week or two? Paging @OmarMcAdam so he can add more context.

Issue #814, opened 2026-03-09
[Bug]: No way to disconnect a provider


OmarMcAdam:
### Summary

Currently we can add providers
but there dosnt seem to be a way to disconnect providers

### Steps to reproduce

add provider
try to remove

### Expected behavior

to have a button to disconnect providers

### Actual behavior

no way to disconnect providers

PR #817, opened 2026-03-10
fix: allow disconnecting providers from settings

OmarMcAdam:
## Summary
- add Disconnect actions for connected providers in the settings Providers section
- remove provider auth with the same DELETE flow OpenCode uses, then dispose and refresh provider state so the list updates immediately
- show disconnect progress and success/error feedback in the settings UI

## Testing
- pnpm typecheck


OmarMcAdam:
fixes #814

Issue #813, opened 2026-03-09
[Bug]: Connecting to new provider timeout before response


OmarMcAdam:
### Summary

**bug when connecting new provider
dialogue oauth response times out prior to return
provider is correctly added to config
app needs to be restarted before you can properly access the new provider**

### Steps to reproduce

add new provider
open the page for auth
app shows "Failed to complete OAuth Request failed Request timed out."
continue flow
close and restart Openwork
provider is connected

### Expected behavior

to not timeout, instead keep checking if the provider has now been connected, and if yes, close modal and refresh provider list

### Actual behavior

shows error, dosnt refresh

OmarMcAdam:
closed by 818

PR #818, opened 2026-03-10
fix provider oauth polling flow

OmarMcAdam:
## Summary
- keep provider OAuth auto flows retrying in the background instead of showing a manual retry button
- treat pending OAuth callback states as waiting conditions while refreshing provider state from a disposed client
- refresh and close the provider modal automatically once the provider connection is detected

## Testing
- `pnpm typecheck`


OmarMcAdam:
fixes #813

PR #865, opened 2026-03-12
Refresh the Den landing page

jcllobet:
## Summary
- split the Den landing page into Den-specific components so the page file stays focused on composition
- replace the static capability grid with a full-width auto-scrolling carousel and make the How it works sequence span a single desktop row
- restore Den naming in the hero and keep the updated footer/support-grid structure aligned with the current landing site
- merge pricing and proof into one value section so the Den bargain is clearer without turning the page into a hard sell

## Verification
- `pnpm --filter @different-ai/openwork-landing build`
- captured a clean before shot from `origin/dev` at `http://127.0.0.1:3302/den`
- captured a clean after shot from this branch at `http://127.0.0.1:3303/den`

### Before
Before Den page

### After
After Den page


jcllobet:
Updated the Den pricing/value section to make the comparison explicitly human repetitive work vs a Den worker, and tightened the CTA to `Start with one worker`.

Verification:
- `pnpm --dir packages/landing build`

PR #866, opened 2026-03-12
fix(landing): update Den landing page CTAs and value section

jcllobet:
- Update pricing copy from 'Cancel anytime' to 'Free for a limited time'
- Remove target=_blank from deploy link
- Remove comparison animation and simplify value section
- Add hire CTA with mailto link and reorganize cost cards


jcllobet (commented):
lgtm

PR #886, opened 2026-03-13
Fix infinite auto compaction loop

casualjim:
## Summary
- Replace idle-only auto-compaction with a policy-based implementation that compacts only when the active model's context usage crosses a threshold.
- Add a small policy seam so compaction behavior can be customized without patching the main app flow.
- Update the Settings copy to reflect that compaction is now context-pressure based instead of running after every completed turn.
- 
## Why
- The previous implementation compacted on every `non-idle -> idle` transition, which could retrigger on its own compaction run and leak continuation summaries into the chat repeatedly.
- OpenCode already exposes model context limits and per-message token usage, so OpenWork should use the same signal instead of a fixed heuristic.

## Issue
- Closes #

## Scope
- `packages/app/src/app/app.tsx`
- `packages/app/src/app/lib/auto-compaction-policy.ts`
- `packages/app/src/app/lib/auto-compaction-policy.custom.ts`
- `packages/app/src/app/pages/settings.tsx`

## Out of scope
- Worker/container packaging
- Kubernetes manifests
- Approval flow behavior
- Existing `Todo` typing issues outside the compaction path

### Ran
- `pnpm --filter @different-ai/openwork-ui typecheck`

### Result
- pass/fail: fail
  - `packages/app/src/app/app.tsx:2308` - `Type 'Todo[]' is not assignable to type 'TodoItem[]'`
  - `packages/app/src/app/context/session.ts:794` - `Argument of type 'Todo[]' is not assignable to parameter of type 'StoreSetter<TodoItem[] ...>'`

## CI status
- pass:
- code-related failures:
  - local typecheck is blocked by pre-existing `Todo` vs `TodoItem` type errors unrelated to this branch
- external/env/auth blockers:
  - none for this change set

## Manual verification
1. Enable `Auto context compaction` in Settings.
2. Run a session that approaches the selected model's context limit but stays below the threshold and confirm no compaction occurs on idle.
3. Continue the same session past the threshold, confirm compaction runs once on idle, and verify it does not loop or dump repeated handoff summaries into the transcript.

## Evidence
- N/A

## Risk
- Low to medium: compaction timing changes from a simple idle trigger to provider/model metadata-driven behavior, so sessions with missing model limit metadata may stop auto-compacting until a custom policy is added.
## Rollback
- Revert `4db87ab9` and `573e0898` from `fix/auto-compaction` to restore the prior idle-based compaction behavior

### DIFF
--- a/packages/app/src/app/app.tsx
+++ b/packages/app/src/app/app.tsx
@@ -1723,6 +1723,7 @@ export default function App() {
       previousStatus: previous,
       status,
       messages: visibleMessages,
+      providers: providers(),
       lastAutoCompactedAt,
       now: Date.now(),
     });
--- a/packages/app/src/app/app.tsx
+++ b/packages/app/src/app/app.tsx
@@ -49,6 +49,7 @@ import {
   shellInSession,
   listCommands as listCommandsTyped,
 } from "./lib/opencode-session";
+import { shouldAutoCompact } from "./lib/auto-compaction-policy";
 import { clearPerfLogs, finishPerf, perfNow, recordPerfLog } from "./lib/perf-log";
 import {
   AUTO_COMPACT_CONTEXT_PREF_KEY,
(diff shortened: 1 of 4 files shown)


src-opn:
resolved

PR #890, opened 2026-03-13
fix: sync user global opencode config into managed config dir

### LINKED ISSUE #870: [Bug]: Permissions in opencode.jsonc are ignored on every launch
### Summary

I have been fighting trying to force Openwork into 'Ask' mode by default. Despite having the below permissions in `~/.config/opencode/opencode.jsonc` and also the workspace's `opencode.jsonc,` the 'active' workspace in openwork upon launch simply ignores these permission rules. 

If I then switch to another worker and start a new chat, the permissions are suddenly adhered to. I can't for the life of me figure out how to make 'Ask' the default behavior. 

```
"permission": {
    "bash": {
      "*": "ask",
      "git status *": "allow",
      "git diff *": "allow",
      "git log *": "allow",
      "git branch *": "allow",
      "grep *": "allow",
      "cat *": "allow",
      "ls *": "allow",
      "find *": "allow",
      "wc *": "allow",
      "head *": "allow",
      "tail *": "allow",
      "echo *": "allow"
    }
  }
```

### Steps to reproduce

1. Set Permissions in opencode.jsonc (global or workspace level)
2. Launch openwork
3. In the default open chat, prompt something that should require `ask` permissions (in my case anything `bash` that is not a list/read/find etc)
4. Command will succeed without prompting to approve
5. Create a new workspace or switch to another existing one, and open a new chat
6. Perform the same restricted operation and you should get prompted for approval this time

This is not workspace specific, if you perform the above and then close and re-launch Openwork, the same workspace that was just working correctly will now ignore permissions. You have to switch to a non-active workspace to pick up on the permissions again. 

### Expected behavior

Global permission should always be adhered to unless specifically overridden at the workspace level. Workspace level permissions should always apply.

### Actual behavior

OpenWork is always defaulting to Allow mode on the active workspace, regardless of defined permissions.

ZeusCraft10:
## Summary
- Sync user's global opencode config files into the orchestrator's managed config directory on startup, so permission rules and other settings from `~/.config/opencode/opencode.jsonc` are respected.

## Why
- The orchestrator sets `OPENCODE_CONFIG_DIR` to a managed directory containing only managed tool scripts, which shadows the user's global config. This causes permission rules defined in `opencode.jsonc` to be silently ignored on the active workspace at launch.

## Issue
- Closes #870

## Scope
- Added `syncUserGlobalOpencodeConfig()` to copy global config files (`opencode.jsonc`, `opencode.json`, `config.json`, `AGENTS.md`) into the managed config dir on daemon and run startup.
- Uses mtime comparison to avoid overwriting workspace-specific edits or SDK-written state.
- Extracted `OPENCODE_GLOBAL_CONFIG_FILES` constant shared between `resolveHostOpencodeGlobalConfigDir` and the new sync function.

## Out of scope
- Filesystem watcher for live config reloading (daemon restart required for mid-session changes, matching sandbox behavior)
- Removing `OPENCODE_CONFIG_DIR` override entirely (managed tools depend on it)

### Ran
- `pnpm -F openwork-orchestrator build`

### Result
- pass
- TypeScript compilation succeeds with no errors

## CI status
- pass: N/A (not yet run)

## Manual verification
1. Create `~/.config/opencode/opencode.jsonc` with custom permission rules (e.g., `"permissions": { "bash": { "*": "ask" } }`)
2. Launch OpenWork — the active workspace should respect those permissions immediately
3. Restart OpenWork — permissions should persist on the initially active workspace
4. Update global config — changes should propagate on next startup (mtime-based sync)

## Evidence
- N/A (requires manual testing with desktop app)

## Risk
- Low. Single file changed (`packages/orchestrator/src/cli.ts`), 30 lines added, 2 removed. Only affects startup paths. Mtime guard prevents accidental overwrites.

## Rollback
- Revert commit. The managed config dir returns to its previous state (no user config files), restoring the pre-fix behavior.

### DIFF
--- a/packages/orchestrator/src/cli.ts
+++ b/packages/orchestrator/src/cli.ts
@@ -95,6 +95,8 @@ const SANDBOX_INTERNAL_OPENCODE_ROUTER_HEALTH_PORT = 3005;
 const SANDBOX_OPENCODE_GLOBAL_CONFIG_CONTAINER_PATH = "/persist/.config/opencode";
 const SANDBOX_OPENCODE_GLOBAL_DATA_IMPORT_CONTAINER_PATH = "/persist/.openwork-host-opencode-data";
 
+const OPENCODE_GLOBAL_CONFIG_FILES = ["opencode.jsonc", "opencode.json", "config.json", "AGENTS.md"];
+
 type ParsedArgs = {
   positionals: string[];
   flags: Map<string, string | boolean>;
@@ -519,10 +521,9 @@ async function resolveHostOpencodeGlobalConfigDir(): Promise<string | null> {
     candidates.push(join(homedir(), "Library", "Application Support", "opencode"));
   }
 
-  const files = ["opencode.jsonc", "opencode.json", "config.json", "AGENTS.md"];
   for (const candidate of Array.from(new Set(candidates.map((item) => resolve(expandTildePath(item)))))) {
     if (!(await isDir(candidate))) continue;
-    for (const file of files) {
+    for (const file of OPENCODE_GLOBAL_CONFIG_FILES) {
       try {
         await access(join(candidate, file));
         return candidate;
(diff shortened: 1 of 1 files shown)


OmarMcAdam:
Prefer this to be opt in rather than enforced

PR #901, opened 2026-03-13
Fix share preview wrapping for long pasted skills

jcllobet:
## Problem
- Pasting long skill text into the share preview kept the preview layers in `white-space: pre`, so the preview pane scrolled horizontally instead of soft-wrapping on desktop and mobile.
- While editing an already-classified preview, the filename dot temporarily fell back to the generic config tone between preview refreshes, which caused a visible flicker.
- The preview still needs to preserve the original pasted text exactly, including real newline boundaries, because we only want visual wrapping.

## Fix
- Added a service-local Playwright regression that pastes the provided Mom Test skill fixture and asserts no horizontal overflow on desktop and mobile, exact textarea value preservation, and visible paragraph/newline separation in the preview.
- Added a second browser regression that delays an in-flight preview refresh and asserts the preview filename dot keeps the prior `dot-skill` classification while the next preview response is pending.
- Updated the shared preview styles used by both the editable home preview and the read-only bundle preview to use `pre-wrap` plus long-token wrapping, and hid horizontal overflow at the preview container.
- Replaced the per-keystroke placeholder recomputation in `share-home-client` with a stable predicted preview item that only resets on explicit source changes, and is refreshed by real preview responses when they arrive.

## Verification
- `pnpm --dir services/openwork-share test`
- `pnpm --dir services/openwork-share test:e2e e2e/preview-wrap.spec.ts`
- `pnpm --dir services/openwork-share build`
- `curl -sS http://127.0.0.1:3101/health`
- Manual Chrome MCP validation on `http://127.0.0.1:3101/` after pasting the long fixture confirmed `wrap`, `highlight`, and `editor` all reported matching `clientWidth` and `scrollWidth`.

## Screenshots
After
CleanShot 2026-03-13 at 12 16 03@2x


jcllobet:
Proof uploaded to the PR:\n\n- Before/after screenshots are embedded in the PR body from committed artifacts in `services/openwork-share/artifacts/`.\n- Browser proof: the new Playwright regression now covers both wrapping and the preview-dot stability case on desktop and mobile.\n- Local verification run on the final HEAD `c862a7d4`:\n  - `pnpm --dir services/openwork-share test:e2e e2e/preview-wrap.spec.ts`\n  - `pnpm --dir services/openwork-share test`\n  - `pnpm --dir services/openwork-share build`

PR #960, opened 2026-03-16
Fix share OG PNG rendering on Vercel

jcllobet:
## Summary
- stop generating public OG PNGs via `sharp` rasterization of SVG
- serve `/og/:id` PNGs with `ImageResponse` instead so Vercel renders readable text for social crawlers
- keep `?format=svg` as a debug fallback and fix the OG continuation URL to use the public share domain

## Root cause
The live page metadata was correct and `/og/:id` was returning a valid `image/png`, but the PNG itself was already broken before WhatsApp saw it. The raw SVG output from the same route was correct. That isolated the failure to the `SVG -> sharp/librsvg -> PNG` rasterization path on Vercel.

## Verification
- `cd services/openwork-share && pnpm build`
- `LOCAL_BLOB_DIR=/tmp/openwork-share-blobs PORT=3002 pnpm start`
- published a local skill with `POST /api/v1/package`
- loaded the share page and extracted `og:image` from the page HTML
- fetched that `og:image` URL directly and verified:
  - `200 OK`
  - `content-type: image/png`
  - `PNG image data, 1200 x 630`

## Evidence
Broken production PNG rendered from the previous pipeline:

Broken production PNG

Fixed local PNG fetched from the share page's `og:image` meta tag after this change:

Fixed local PNG

### DIFF
--- a/services/openwork-share/app/og/[id]/route.ts
+++ b/services/openwork-share/app/og/[id]/route.ts
@@ -1,6 +1,6 @@
-import sharp from "sharp";
 import { fetchBundleJsonById } from "../../../server/_lib/blob-store.ts";
-import { renderBundleOgImage, renderRootOgImage } from "../../../server/_lib/render-og-image.ts";
+import { buildBundleOgImageModel, buildRootOgImageModel, renderBundleOgImage, renderRootOgImage } from "../../../server/_lib/render-og-image.ts";
+import { renderOgPngResponse } from "../../../server/_lib/render-og-image-response.tsx";
 
 export const runtime = "nodejs";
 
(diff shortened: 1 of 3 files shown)


jcllobet:
Verified locally via the crawler path:

- share page: http://127.0.0.1:3002/b/01KKWANJBTHZCNN2VWJWBAABM9
- og:image extracted from page HTML: http://127.0.0.1:3002/og/01KKWANJBTHZCNN2VWJWBAABM9

Broken production PNG from the previous pipeline:

Broken production PNG

Fixed local PNG fetched from the page metadata after this change:

Fixed local PNG

PR #976, opened 2026-03-17
fix den-v2 desktop auth handoff parity

OmarMcAdam:
## Summary
- port the desktop auth handoff endpoints from `services/den` into `services/den-v2` so desktop sign-in works on the Daytona-first control plane too
- register the new `/v1/auth` router in `services/den-v2/src/index.ts`
- add the `desktop_handoff_grant` SQL migration required by the new handoff exchange flow

## Validation
- compared `services/den/src/http/desktop-auth.ts` against den-v2 and ported the missing flow
- `npm run build` in `services/den-v2`


OmarMcAdam:
Root cause note for reviewers:\n\n-  was originally introduced in  by  ().\n- During the later shared-schema refactor that moved Den DB ownership into , that table definition did not get carried over.\n- The desktop auth routes still expected the table to exist, so  missed the handoff flow entirely and the shared schema no longer matched the route implementation.\n- This PR restores  in , updates both Den services to consume the shared definition, and keeps the  handoff migration in place so the database can be brought back in sync.\n- I also disabled the  and  GitHub workflows here because Render is now the source of truth for Den builds/deploys.

OmarMcAdam:
Root cause note for reviewers:

- `DesktopHandoffGrantTable` was originally introduced in `services/den/src/db/schema.ts` by commit `6143397e` (`feat(den): hand off desktop auth through the web`).
- During the later shared-schema refactor that moved Den DB ownership into `packages/den-db`, that table definition did not get carried over.
- The desktop auth routes still expected the table to exist, so `den-v2` missed the handoff flow entirely and the shared schema no longer matched the route implementation.
- This PR restores `DesktopHandoffGrantTable` in `packages/den-db`, updates both Den services to consume the shared definition, and keeps the `den-v2` handoff migration in place so the database can be brought back in sync.
- I also disabled `.github/workflows/deploy-den.yml` and `.github/workflows/deploy-den-v2.yml` here because Render is now the source of truth for Den builds and deploys.

PR #978, opened 2026-03-17
fix(session): render the reload-required prompt inline

benjaminshafii:
## Summary
- render the session reload-required prompt as a full-width inline surface instead of an inset modal-style card
- remove the rounded card chrome, backdrop treatment, and drop shadow so the prompt reads like part of the page layout
- keep the existing copy, actions, and blocked-task warning behavior intact

## Testing
- `pnpm typecheck`
- `pnpm --filter @different-ai/openwork-ui build`

## Notes
- I did not run the full Docker + Chrome MCP flow or capture screenshots/video for this PR.
- Reviewer repro for the full UI gate: run `packaging/docker/dev-up.sh` from the repo root, trigger the reload-required state in a session, and confirm the prompt spans the full width without modal/backdrop styling or a bottom shadow.

### DIFF
--- a/packages/app/src/app/pages/session.tsx
+++ b/packages/app/src/app/pages/session.tsx
@@ -4217,93 +4217,89 @@ export default function SessionView(props: SessionViewProps) {
           </Show>
 
           <Show when={props.showSkillReloadBanner}>
-            <div class="border-b border-dls-border/70 bg-dls-surface px-4 py-3 sm:px-6 lg:px-10">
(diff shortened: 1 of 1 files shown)


benjaminshafii:
Done - this is implemented in #978. The reload-required prompt now renders as a full-width, full-height inline surface without modal/backdrop styling or the lower drop shadow. Verified with: pnpm typecheck and pnpm --filter @different-ai/openwork-ui build.

PR #997, opened 2026-03-18
refactor(den): remove duplicate den-v2 service

OmarMcAdam:
## Summary
- remove the duplicate `services/den-v2` app and disabled `deploy-den-v2` workflow so `services/den` is the only Den service surface
- keep the v2-ready behavior in the canonical service by preserving the Daytona-first default and the newer desktop handoff implementation
- fix the Den and worker-proxy Docker entrypoints so the canonical service can be started and verified locally

## Testing
- `pnpm install`
- `pnpm --filter @openwork/den build`
- `bash packaging/docker/den-dev-up.sh`
- `curl -fsS http://localhost:61355/health`
- verified the signup screen in Chrome DevTools at `http://localhost:61356`


OmarMcAdam:
Superseded by #1002. I removed the wrong service in this PR; #1002 keeps  and removes legacy  instead.

PR #1025, opened 2026-03-18
Hide OpenWork to tray on close

jcllobet:
## Summary
- keep the desktop app alive when the main window is closed by preventing the close request and hiding the window instead
- add a tray icon with `Show OpenWork` and `Quit OpenWork` actions so the hidden app can be restored or exited explicitly
- keep the existing sidecar cleanup on real app exit, and enable Tauri's `tray-icon` feature to support the new tray behavior

## Testing
- `cargo check` in `packages/desktop/src-tauri`
- local dev run via `pnpm dev`
- temporary local-only instrumentation (removed before commit) verified the close path kept the process alive while the window count dropped to 0, and the show helper restored the window count to 1

## Notes
- I could not attach before/after screenshots because this environment does not have macOS Screen Recording permission (`CGPreflightScreenCaptureAccess()` was false and `screencapture` failed with `could not create image from display`)
- I attempted native capture before and after the change, but the OS permission blocker prevented artifact generation
- reviewer reproduction: run `pnpm dev`, close the window, then use the tray icon menu to choose `Show OpenWork` or `Quit OpenWork`

### DIFF
--- a/packages/desktop/src-tauri/Cargo.toml
+++ b/packages/desktop/src-tauri/Cargo.toml
@@ -21,7 +21,7 @@ json5 = "0.4"
 notify = "6.1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tauri = { version = "2", features = [] }
+tauri = { version = "2", features = ["tray-icon"] }
 tauri-plugin-deep-link = "2"
 tauri-plugin-dialog = "2"
 tauri-plugin-http = "2.5.6"
--- a/packages/desktop/src-tauri/src/lib.rs
+++ b/packages/desktop/src-tauri/src/lib.rs
@@ -236,6 +236,15 @@ pub fn run() {
             api.prevent_close();
             hide_main_window(&app_handle);
         }
+        #[cfg(target_os = "macos")]
+        RunEvent::Reopen {
+            has_visible_windows,
+            ..
+        } => {
+            if !has_visible_windows {
+                show_main_window(&app_handle);
+            }
+        }
         _ => {}
     });
 }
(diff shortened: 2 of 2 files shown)


jcllobet:
Follow-up on evidence: I attempted to capture before/after screenshots for this PR, but this machine currently lacks macOS Screen Recording permission for CLI tools. CGPreflightScreenCaptureAccess() returned false and screencapture failed with 'could not create image from display'.

What I verified instead:
- cargo check passed in packages/desktop/src-tauri
- a temporary local-only close-flow harness (removed before commit) confirmed the app process stayed alive after a close request while the visible OpenWork window count dropped to 0
- the same local harness confirmed the restore helper brought the visible window count back to 1

Happy to attach screenshots later once Screen Recording is available.

jcllobet:
Pushed a follow-up fix for tray restore.\n\nWhat changed:\n- left-clicking the tray icon now restores and focuses the main window\n- tray menus no longer open on left click, so the click consistently acts as restore\n\nVerification:\n- cargo check in packages/desktop/src-tauri

jcllobet:
Pushed a macOS dock-reopen follow-up.

Root cause:
- closing the last window hid it successfully, but macOS dock/app reopen events were not restoring the hidden main window

Fix:
- handle `RunEvent::Reopen` on macOS and call the existing main-window restore helper when the app has no visible windows

Validation before commit:
- reproduced the bug locally on the PR branch: after a close request, visible window count dropped to 0 and activating the running app still left the window hidden
- added a temporary local-only test hook (removed before commit) to auto-trigger close during startup
- launched the debug app bundle, confirmed the same app PID stayed alive while hidden, then reopened it via `open -a /.../OpenWork.app`
- observed the macOS reopen event fire and the visible window count return from 0 to 1 on the same process
- ran `cargo check` after removing the temporary test hook

PR #1043, opened 2026-03-19
fix(app): render session errors as typed transcript rows

jcllobet:
## Summary
- replace synthetic session-error message injection with typed transcript rows so chat errors render inline without pretending to be assistant messages
- keep session-scoped non-tool errors inline, while suppressing tool-specific timeout/invalid-tool chat errors so those stay inside the tool UI
- remove the red failed footer for inline chat errors and add before/after screenshots for the updated transcript behavior

## Testing
- `pnpm --filter @openwork/app test:sessions` *(fails in this environment: `Timed out waiting for /global/health: Unauthorized`)*
- `pnpm --filter @openwork/app test:events` *(fails in this environment: `Timed out waiting for /global/health: Unauthorized`)*
- manual Chrome verification on local Vite builds using the real `MessageList` component to capture before/after transcript screenshots

## Screenshots
Before

Before inline error handling

After

After inline error handling


jcllobet:
I switched session-scoped chat failures over to typed transcript rows, so the inline error surface is no longer a fake assistant message and tool timeout/invalid-tool errors stay suppressed at the chat level.

Before

Before inline error handling

After

After inline error handling

jcllobet:
Tightened the inline session error layout so the heading and detail now share a wrapped row instead of being forced onto separate lines. The detail can still drop to the next line and take the remaining width when needed.

PR #1087, opened 2026-03-21
Show settings scope and inheritance in the UI

benjaminshafii:
## Summary
- add a settings legend that explains global, workspace, and inherited values
- label key settings cards so users can tell which layer they are editing
- carry the same provenance language into plugin and MCP configuration surfaces

## Testing
- Not run: Docker + Chrome MCP end-to-end verification for this branch

### DIFF
--- a/apps/app/src/app/pages/mcp.tsx
+++ b/apps/app/src/app/pages/mcp.tsx
@@ -652,6 +652,21 @@ export default function McpView(props: McpViewProps) {
               </button>
             </div>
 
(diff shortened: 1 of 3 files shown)


benjaminshafii:
Chrome MCP verification pass completed on `localhost:5173`.

Verified:
- Settings shows the new global/workspace/inherited legend
- settings cards now explain which scope owns each value
- the Extensions surface shows matching workspace-scope provenance for MCP/plugins

Screenshot:
- https://files.catbox.moe/64kx19.png

PR #1088, opened 2026-03-21
Refine sidebar session hierarchy for enterprise histories

benjaminshafii:
## Summary
- collapse child task sessions behind disclosures so parent conversations stay easy to scan
- preserve child sessions in the local session store so the sidebar can keep nested history visible
- tighten sidebar density while keeping workspace actions pinned at the bottom

## Testing
- Not run: Docker + Chrome MCP end-to-end verification for this branch


benjaminshafii:
Chrome MCP verification pass completed on `localhost:5173` against the local preview + headless host.

Verified:
- the sidebar shell loads correctly with the updated hierarchy code in place
- new-session navigation still works from the workspace rail after the hierarchy changes

Screenshot:
- https://files.catbox.moe/simmp6.png

Note: I did not manage to force a child sub-session from the local preview, so the disclosure state for nested task sessions still needs reviewer confirmation on a branch/run that produces child sessions.

Issue #1103, opened 2026-03-22
[Bug]: Agent stuck in infinite session summary loop**


abyshajan:
### Summary

the agent automatically generates a "Goal / Instructions / Discoveries / Accomplished" summary block after every single message — including simple greetings like "hi". The summary then gets passed as context to the next response, which triggers another summary, creating an infinite loop.

### To Reproduce

1. Open OpenWork (Starter workspace)
2. Start a new session
3. Type "hi"
4. Agent responds with a full structured summary instead of a normal reply
5. Every subsequent message triggers another summary — the agent never responds normally

### Expected behavior

Agent should respond naturally to messages. Session summaries should only appear when explicitly requested or when context compression is genuinely needed.

### Actual behavior

Every response includes a structured block like:
- Goal
- Instructions
- Discoveries
- Accomplished
- Relevant files / directories
- Next steps for the next agent

### Additional context (optional)

The tool is completely unusable in this state. Every session immediately enters the loop regardless of input.
This appears to be a built-in session handoff / context compression feature that is misfiring on every message instead of only triggering when context is long or explicitly requested.

robinszeto:
Same here.

benjaminshafii:
Linking this with #1074 and #777 (and the broader agent-loop discussion in #932) since these reports look like the same recurring loop family.

surapuramakhil:
@benjaminshafii this issues still exist in latest version 1.84 - its a breaking issue, unable to use openwork

sacrtap:
This problem is so fatal that openwork cannot be used at all.

OmarMcAdam:
We have a fix for this
Will be in next release

OmarMcAdam:
Can you try in .185 + and report if ths issue is still happening?
We shipped a fix to detect if the convo is in a `doom_loop` and require input from the user
This mimics the underlying opencode implementation

will keep issue open for now till we get confirmation of a fix

sacrtap:
> Can you try in .185 + and report if ths issue is still happening? We shipped a fix to detect if the convo is in a `doom_loop` and require input from the user This mimics the underlying opencode implementation
> 
> will keep issue open for now till we get confirmation of a fix

Still not resolved @OmarMcAdam 

Version 0.11.186 (0.11.186)

<img width="1740" height="1614" alt="Image" src="https://github.com/user-attachments/assets/a527b403-098a-47c9-a6d4-3515469ba207" />

[...]

Issue #1074, opened 2026-03-20
[Bug]: Seems to be looping constantly when trying to execute tasks


sharan01x:
### Summary

I asked it to execute a task. The first task was executed well, no issues there. When I asked for a follow-up task, it summarised my request again under those headings "1. User Requests (As-Is), 2. Final Goal, 3. Work Completed...", fine. But it started doing that again and again after that in a loop. It wasted a lot of tokens doing this. It never went ahead to execute the task. 

### To Reproduce

1. Request a task to be performed
2. Ask for a follow-up task

### Expected behavior

It may have summarised my request and planned, but after that it should have executed the task. 

### Actual behavior

I've detailed this already

### OW version & Desktop info (optional)

Open Work Version 0.11.175 (0.11.175)
Mac OS Tahoe 26.3.1

### Additional context (optional)

It's been happening over the last two versions I think

youxufkhan:
This happens to me everytime I use openwork, I thought its how it used to work so I stopped using it because it was very frustrating.

benjaminshafii:
Linking this with #1103 and #777 since it looks like the same looping / repeated-summary family. Assigning @OmarMcAdam so the root-cause work stays consolidated.

PR #1109, opened 2026-03-22
fix(session): stop repeated synthetic continuation loops

src-opn:
## Summary
- add an OpenWork-side loop guard in `apps/app/src/app/context/session.ts` that watches for repeated synthetic continuation prompts and aborts the session before it can keep burning tokens
- cover both repeated auto-compaction continue prompts and the repeated synthetic task-summary control prompt that tends to precede the `Goal / Instructions / Discoveries / Accomplished` loop family
- surface the abort back into the transcript as a synthetic session error so the user sees why the run was paused instead of getting stuck in an endless responding state

## Validation
- `pnpm --filter @openwork/app typecheck` ✅
- `CI=true pnpm install --frozen-lockfile && pnpm --filter @openwork/app build` ✅
- `pnpm --filter @openwork/app test:sessions` ⚠️ fails in this environment with `Timed out waiting for /global/health: Unauthorized`
- `pnpm --filter @openwork/app test:permissions` ⚠️ fails in this environment with `Timed out waiting for /global/health: Unauthorized`
- Docker + Chrome MCP smoke check via `packaging/docker/dev-up.sh`: opened `http://localhost:60296/session`, sent `hi`, confirmed the session replied normally, and captured `apps/app/pr-issue-777-greeting-smoke.png`

## Notes
- I validated the user-written manual but did not apply it directly: the local `autoCompactContext` toggle is off by default, the SQL permission patch does not match the broad repro pattern in `#777`, and filtering assistant summaries in the UI would hide the symptom without stopping the loop
- this PR is an OpenWork-side mitigation for the current issue family (`#777`, `#1074`, `#1103`) that pauses likely synthetic continuation loops and leaves a visible explanation in-session


src-opn:
A bit more detail on the issue analysis and what this PR actually fixes:

The user-written manual was directionally useful, but after checking the current `apps/app` code I do not think the main problem is a simple local auto-compaction debounce or a bad SQLite permission row.

What seems to be happening instead is a loop in the synthetic control messages that OpenCode/OpenWork use to keep a run moving forward:
- one path is the synthetic compaction continuation prompt: `Continue if you have next steps, or stop and ask for clarification if you are unsure how to proceed.`
- another path is the synthetic task-summary continuation prompt: `Summarize the task tool output above and continue with your task.`

When those keep getting re-injected, the model can fall into the recurring summary family reported in `#777`, `#1074`, and `#1103` (`Goal / Instructions / Discoveries / Accomplished`, etc.). That is why the symptom is not just visual clutter: it can keep spending tokens and keep the session in a pseudo-running state.

Because of that, this PR intentionally does **not** do the following:
- it does **not** just hide summary-looking assistant messages in the UI, because that would mask the loop while the run keeps burning tokens
- it does **not** rely on the local `autoCompactContext` toggle alone, because that setting is off by default and does not explain the broader repro pattern
- it does **not** apply the database SQL patch, because that did not line up with the current session creation path or the cross-issue repro shape

What this PR **does** solve:
- it adds an OpenWork-side guard in `apps/app/src/app/context/session.ts` that watches the live session event stream for repeated synthetic continuation control prompts
- if OpenWork sees the session fall into one of those repeated synthetic control loops, it aborts the session proactively
- it then surfaces a synthetic session error in the transcript so the user sees *why* the run was stopped instead of being left in an endless responding/token-burning loop

So this is best described as a **defensive mitigation in OpenWork** for the current infinite-summary / repeated-continuation issue family. It makes the app fail safe when the upstream run gets stuck in these synthetic continuation patterns.

What it does **not** claim to solve:
- it is not a root-cause fix inside OpenCode's core loop/compaction behavior
- it does not prevent every possible future looping behavior
- it does not change compaction policy itself; it detects the specific repeated synthetic continuation patterns and cuts them off

In short: this PR is meant to stop the user-visible failure mode that burns tokens and makes the app unusable, while leaving a clear explanation in-session when it happens.

Issue #860, opened 2026-03-12
[Bug]: Session history cannot be saved and restored after app restart/exit on Windows 11 Desktop v0.11.142


baddieflex:
### Summary

On Windows 11, the OpenWork desktop application (v0.11.142) fails to persist chat session history. All created conversations and message records are completely lost after closing the app, restarting the program, or refreshing the workspace, and cannot be recovered when reopening the same workspace.

### Steps to reproduce

1. Install and launch OpenWork Desktop v0.11.142 on a Windows 11 device
2. Select and load a local workspace 
3. Create a new chat session, send multiple messages to generate complete conversation content
4. Fully close the OpenWork application (confirm no related background processes remain in Windows Task Manager)
5. Reopen the OpenWork desktop app, and navigate back to the same workspace used in step 2
6. (Additional reproduce scenario) Refresh the current workspace after creating a session, the session also disappears immediately

### Expected behavior

1. All previously created chat sessions and complete message records should be retained and displayed normally in the sidebar after the app is restarted, exited, or the workspace is refreshed
2. The "No sessions yet" prompt should only appear when the workspace is opened for the first time and no sessions have ever been created

### Actual behavior

1. All chat sessions are completely lost after restarting/exiting the app or refreshing the workspace, the sidebar shows empty or only the "No sessions yet" prompt
2. The session data cannot be restored even when opening the exact same workspace, and the issue persists after multiple retries

lrq3000:
Same issue for me on Windows 11 Pro. First time I install OpenWork. This issue persists with the latest version released today with binaries (v0.11.142).

It's not even necessary to restart the app, with Developer mode enabled, just click "Reload the engine" in the "Advanced" menu, and all the sessions are lost.

I asked my agent to see if the sessions were truly lost, and apparently not, they are still present, but just they are not shown in the UI for some reason. So it is an UI problem.

I can confirm the sessions were indeed present (I checked the content) in:

* Main session store: `~\.local\share\opencode\opencode.db`
* Auxiliary per-session diff files: `~\.local\share\opencode\storage\session_diff\ses_*.json`

For some reason, OpenWork cannot show them.

It is worth mentioning I also have OpenCode installed and I used it before. So maybe this is caused by some incompatibliity? Because OpenCode already added several sessions, so maybe the sessions format is different enough from what OpenWork expects that OpenWork silently fails to decode the session files?

sslcheese:
[...]

PR #1164, opened 2026-03-25
fix(app): bypass path normalization to fix sidebar session queries on…

shanliuling:
**Fixes:** #860 
### What
Bypass aggressive path normalization for the `directory` query when fetching sidebar workspace sessions from the OpenCode server on Windows.
### Why
Currently on Windows, the `normalizeDirectoryQueryPath` function converts raw Windows paths (e.g., `D:\Projects\MyWork`) into forward-slashed lowercase paths (e.g., `d/projects/mywork`). However, the OpenCode server's underlying SQLite database stores the raw paths. This mismatch causes the server-side directory filter to fail, resulting in an empty session history list in the sidebar on app restart.
### How
- Removed `normalizeDirectoryQueryPath` from the `queryDirectory` assignment in [refreshSidebarWorkspaceSessions](cci:1://file:///d:/z-work/openwork/apps/app/src/app/app.tsx:3369:2-3475:4) (located in [app.tsx](cci:7://file:///d:/z-work/openwork/apps/app/src/app/app.tsx:0:0-0:0)).
- By passing the raw `directory` string directly to the server, the query correctly matches the stored SQLite payload.
- Client-side filtering logic (`filtered = root ? list.filter(...) : list`) remains unchanged, safely using `normalizeDirectoryPath` on both sides to ensure exact matching after the data is received.


lrq3000:
Please merge this PR, OpenWork is unusable on Windows without this fix (bug still happening on v0.11.196).

shanliuling:
> Please merge this PR, OpenWork is unusable on Windows without this fix (bug still happening on v0.11.196)

 Thanks for the support!
  If you are building from source and need this fix immediately, you can manually apply the changes:
   1. Open the file: apps/app/src/app/context/sidebar-sessions.ts
   2. Find line 142: const queryDirectory = normalizeDirectoryQueryPath(directory) || undefined;
   3. Change it to: const queryDirectory = directory || undefined;
  This will bypass the path normalization that causes the query mismatch on Windows. Hope this helps while we wait for the

lrq3000:
Please devs merge this, currently Windows users cannot use this app at all due to this bug, that's a big part of computer users.

I tried to build but it is a nightmare to build and run on Windows.

shanliuling:
Has the latest version fixed it? If it has, I will close this PR.

lrq3000:
@shanliuling Unfortunately I cannot test anymore since the access to the Windows build was made to be paid.

Given how the app is freely accessible for other platforms and how unreliable it is on Windows, I don't think that was a good idea at all and is quite discriminative, but maybe the idea was to reduce the devs workload more than really making a commercial offering?

OmarMcAdam:
Trying to borrow a laptop to test this on

OmarMcAdam:
Thanks for digging into this. I did a deeper pass on the bug and the proposed fix is pointing at the right root cause: Windows session history breaks when the sidebar query path no longer matches the exact directory string stored by the server/SQLite layer. That said, this PR appears to be superseded by #1256, which is already merged into `dev` and updates the same code path in `apps/app/src/app/context/sidebar-sessions.ts` to use `toSessionTransportDirectory(directory)` instead of the local-only query normalizer. That landed as commit `13f909a1` and is a slightly safer version of the same fix because it preserves the transport format expected by the server while still handling Windows-specific path variants.

So my current read is:
- the issue in #860 was real and this PR identified the right failure mode
- this PR itself is no longer needed because `dev` already contains the fix
- if Windows users still hit this on builds that include `v0.11.197+`, then there is likely a second bug beyond this sidebar query mismatch

Unless we find a remaining repro on current `dev`, I think this PR should be closed as superseded by #1256.

PR #1165, opened 2026-03-25
fix(aur): handle /usr/bin/opencode conflict with opencode-bin package

### LINKED ISSUE #1163: [Bug]: openwork AUR PKGBUILD needs modification to be compatible with existed opencode file
### Summary

I'm the co-maintainer of Archlinux openwork AUR package. Many user(including me) encounter an error that the openwork installation would conflict with the pre-installed opencode file. I added the compatibility with this code snippet:
`if [ -f /usr/bin/opencode ]; then rm -f "${pkgdir}/usr/bin/opencode"; fi;`
But I found that the openwork auto update bot does not retain this, so the old error occurred again.
Please modify the auto update bot to fix this. 
Thanks a lot.

### To Reproduce

Install/update the openwork AUR package

### Expected behavior

completed without error 

### Actual behavior

File conflict occurred:

> (1/1) checking for file conflicts                                                                                              [############################################################################] 100%
> error: failed to commit transaction (conflicting files)
> openwork: /usr/bin/opencode exists in filesystem (owned by opencode-bin)
> Errors occurred, no packages were upgraded.

### Screenshots (optional)

_No response_

_No response_

_No response_

Fizza-Mukhtar:
## Summary

Fixes #1163

When `opencode-bin` is installed alongside `openwork`, pacman aborts
with a file conflict because both packages install `/usr/bin/opencode`.

## Root Cause

The `package()` function in the PKGBUILD extracts the full `data.tar`
from the `.deb` release artifact, which includes both `/usr/bin/openwork`
and `/usr/bin/opencode`. The standalone `opencode-bin` AUR package also
owns `/usr/bin/opencode`, causing pacman to refuse the transaction.

## Fix

Added a guard at the end of `package()` that checks whether `opencode-bin`
is installed on the system. If it is, the staged `/usr/bin/opencode` is
removed from `$pkgdir` before pacman takes ownership — allowing both
packages to coexist cleanly.

openwork does not require `/usr/bin/opencode` to be in PATH; it
invokes the binary from its bundled location under `/opt/openwork/`
at runtime.

## Why this fix persists across bot updates

The AUR update bot (`aur_validate.yml`) patches only `pkgver`,
`pkgrel`, `arch`, `source_*`, and `sha256sums_*` fields via regex.
It does not modify the `package()` function body, so this fix will
not be reverted by automated release updates.

## Testing

- [x] Verified conflict error occurs without the fix when `opencode-bin` owns `/usr/bin/opencode`
- [x] Verified clean install (no `opencode-bin`) still installs both `/usr/bin/openwork` and `/usr/bin/opencode`
- [x] Verified coexistence install removes only the staged copy, leaving `opencode-bin`'s ownership intact

### DIFF
--- a/packaging/aur/PKGBUILD
+++ b/packaging/aur/PKGBUILD
@@ -25,14 +25,11 @@ package() {
   # Extract the internal data archive directly to the cwd.
   bsdtar -O -xf "${pkgname}-${pkgver}.deb" 'data.tar*' | bsdtar -C "${pkgdir}" -xf -
 
-  # Remove the bundled opencode binary from the package staging dir if
-  # opencode-bin is already installed on the system. Both packages install
-  # /usr/bin/opencode and pacman will abort with a file conflict otherwise.
-  # openwork does not require this binary to be in PATH; it uses the copy
-  # inside /opt/openwork/ at runtime.
-  if pacman -Qq opencode-bin &>/dev/null 2>&1; then
-    rm -f "${pkgdir}/usr/bin/opencode"
-  fi
+  # The upstream .deb includes /usr/bin/opencode, but the AUR package should
+  # not claim the global CLI name. OpenWork uses its bundled copy under
+  # /opt/openwork/ at runtime, while opencode-bin should remain the owner of
+  # /usr/bin/opencode when users install the standalone CLI.
+  rm -f "${pkgdir}/usr/bin/opencode"
 }
 # .deb Internal Structure Reference:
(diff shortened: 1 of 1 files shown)


benjaminshafii:
Thanks for jumping on this. I pushed a small follow-up on top of your fix: we now remove the staged \ unconditionally instead of checking the build machine for \, because the AUR package should not claim the global CLI name at all. OpenWork still keeps and uses its bundled pinned copy under \, so this avoids the pacman conflict without depending on the builder's local package state.

PR #1177, opened 2026-03-26
fix(sessions): preserve sessions when server returns empty list on reconnect

### LINKED ISSUE #1138: [Bug]: SlackBot not properly stabilizing connection to workspace
### Summary

When the app is updated, the slackbot is unable to keep it's connection with the previous workspace. 

### To Reproduce

1. Send message via slackbot in the workspace
2. Update the app
3. Send message via slackbot in the same workspace

### Expected behavior

the message gets sent

### Actual behavior

no message / connection

### Screenshots (optional)

Image

- OW version 0.11.189

it might be connected to the fact that the port for the workspace is restarting and the slackbot is connected to the old port

Fizza-Mukhtar:
**Summary**

When the engine restarts (app update, worker switch, workspace change),
loadSessions() may receive an empty list from the server before it has
finished indexing the workspace. reconcile([]) was wiping all sessions
from the store, making session history appear lost to the user.

Add a guard in loadSessions() that skips the reconcile when:
- a specific scope root was requested (scoped load, not global)
- the server returned an empty filtered list
- the store already has sessions belonging to that root

This preserves existing sessions until the server returns a non-empty
result. Global loads (no scopeRoot) are unaffected and always reconcile.

Fixes session history disappearing after:
- worker switch (Variant A)
- dashboard status bar workspace open (Variant B)
- app update (Variant C)

Fixes #1138

### DIFF
--- a/apps/app/src/app/context/session.ts
+++ b/apps/app/src/app/context/session.ts
@@ -828,6 +828,30 @@ export function createSessionStore(options: {
       })),
     });
     sessionDebug("sessions:load:filtered", { root: root || null, count: filtered.length });
+
+    // Guard: do not wipe existing sessions with an empty list when a specific
+    // scope root was requested. This happens when the engine is restarting
+    // (app update, worker switch, workspace change) and session.list() returns
+    // [] before the server has finished indexing — reconcile([]) would destroy
+    // all session history from the store. We only skip the store update; the
+    // loadedScopeRoot is still updated so the next successful load can proceed
+    // normally. If no scope root was given (global load), we always reconcile
+    // since an empty global list is authoritative.
+    if (root && filtered.length === 0 && store.sessions.length > 0) {
+      const existingRootSessions = store.sessions.filter(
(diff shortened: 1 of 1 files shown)


src-opn:
Interesting discovery @Fizza-Mukhtar 

But i think it would be better to check if indexing complete !true, or if array is 0, then recheck after 60 seconds for a limit of 5 seconds



your current method could mean that conversations do not sync, so something new could exist on server but not client and client will never know about it.

Fizza-Mukhtar:
> Interesting discovery @Fizza-Mukhtar
> 
> But i think it would be better to check if indexing complete !true, or if array is 0, then recheck after 60 seconds for a limit of 5 seconds
> 
> your current method could mean that conversations do not sync, so something new could exist on server but not client and client will never know about it.

Thanks for the suggestion! Just to make sure I implement this correctly  by "recheck after 60 seconds for a limit of 5 seconds", do you mean retry every 5 seconds for up to 60 seconds total, or retry once after 60 seconds with a 5 second request timeout?

PR #1333, opened 2026-04-04
fix(app): stabilize empty-session draft updates

jcllobet:
## Summary
- keep the `SessionView` prop object stable in `apps/app/src/app/app.tsx` and expose reactive fields through getters so composer draft updates only invalidate the props that actually changed
- stop the empty-session starter cards from getting pulled into the same update path as `prompt`, which removes the visible flicker while typing the first message in a fresh session
- add before/after screenshots for the empty-session state in `apps/app/pr/screenshots/session-empty-state-before-fix.png` and `apps/app/pr/screenshots/session-empty-state-after-fix.png`

## Before / After
- Before: every draft change recreated the full `sessionProps()` object, so typing into the composer on a brand new session caused the empty-state quickstart panel (`What do you want to do?`, `Work on a CSV`, `Automate Chrome`) to re-render and visibly flicker.
- After: the `SessionView` props stay stable and only `props.prompt` updates during typing, so the empty-state panel stays visually steady while the composer continues to update normally.

## Verification
- `pnpm typecheck`
- `pnpm build:ui`
- browser verification against a local preview build at `http://localhost:4173/session`
- connected the app to a local OpenWork orchestrator instance at `http://localhost:60535`, opened a brand new session, typed into the composer, and confirmed the empty-state quickstart panel remained mounted while typing

## Notes
- I started the Docker dev stack with `packaging/docker/dev-up.sh`, but the `share` service failed to boot (`ENOTEMPTY: directory not empty, rmdir '/app/node_modules/.pnpm'`), so I completed the UI verification with a local preview build instead.


jcllobet:
@src-opn thoughts? It's very hacky but it should resolve the flicker when starting a new chat and typing in the composer

benjaminshafii:
this has been fixed already have you tested head of dev @jcllobet ?

jcllobet:
@benjaminshafii not on the latest version, no. It wasn't working for me when I got this issue but it was happening only on the app, not in dev (and bundling it didn't surface it either). Closing PR.

PR #1341, opened 2026-04-04
fix(i18n/zh): cleanups to Chinese translations

johnnyshields:
- Skills: titles use "Skills（技能）", inline uses "skills" following Anthropic official Chinese translations
- MCP: keep as literal name "MCP", do not translate to 应用
- Apps (where EN says "Apps"): 应用
- Commands: 命令, Plugins: 插件, Sessions: 会话
- Remove 571 incorrect CJK-Latin spaces
- Standardize 沙盒→沙箱, 您→你, 仪表盘→主页
- Update header comment with terminology glossary


johnnyshields:
This should be the last translation PR for awhile. I may do one last pass for anything we missed. Thanks for merging 🙏

jcllobet:
I think there's a slight regression here as it removes the spaces between Chinese words and english words when both translations are in the same line.

PR #1357, opened 2026-04-05
docs(changelog): render headings from tracker titles

jcllobet:
## Summary
- update the changelog generator to read all split `changelog/release-tracker-YYYY-MM-DD.md` files instead of the removed single tracker file
- use each release `Title` to render combined headings like `v0.11.202 - ...` when tracker entries include one
- keep the generated docs stable when titles are absent by falling back to version-only headings

## Testing
- `node scripts/generate-changelog.mjs`
- `git diff --check -- scripts/generate-changelog.mjs packages/docs/changelog.mdx`

## Notes
- current `origin/dev` tracker files do not yet include `#### Title`, so the regenerated docs output stays unchanged on this PR until the tracker-title branch lands

### DIFF
--- a/packages/docs/changelog.mdx
+++ b/packages/docs/changelog.mdx
@@ -1,548 +1,536 @@
 ---
 title: "Changelog"
 ---
-<Update label="April 4th" tags={["🚀 New Features", "🐛 Bug Fixes"]}>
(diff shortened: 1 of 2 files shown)


jcllobet:
Superseded by #1358, which carries the full tracker rewrite plus the generator update in one PR.

PR #1379, opened 2026-04-07
fix: clear stale model ID from config and update DEFAULT_MODEL

### LINKED ISSUE #1237: bug: Workspace templates reference removed model ID `opencode/big-pickle`
## Summary

OpenWork starter workspace templates pin the model ID `opencode/big-pickle` in their `opencode.jsonc` config files. This model no longer exists in newer OpenCode builds, causing model selection failures and broken chat flows in workspaces created from these templates.

## To Reproduce

1. Create a new workspace from a template (or open an existing template-based workspace)
2. Open `opencode.jsonc` inside the workspace directory — contains `"model": "opencode/big-pickle"`
3. Try to start a chat — model not found, chat fails

## Expected behavior

- Templates should use a model ID that exists in the current build
- Invalid/removed model IDs should fall back to the user's configured default
- Templates should reference a stable alias (e.g., `default`) rather than a hardcoded model ID

## Actual behavior

Workspaces created from templates are broken out of the box because they reference a model that no longer exists. There is no fallback — the chat simply fails.

## OW version & Desktop info

- **OS:** Arch Linux x86_64

## Additional context

Workaround: replace on every launch via script:

```python
from glob import glob
for config in glob("~/.local/share/com.differentai.openwork/workspaces/*/opencode.jsonc"):
    content = open(config).read()
    if '"opencode/big-pickle"' in content:
        open(config, 'w').write(content.replace('"opencode/big-pickle"', '"litellm/qwen-coder"'))
```

caoergou:
## Summary

Workspace configs (`opencode.jsonc`) written when `opencode/big-pickle` was a valid model now cause silent chat failures because that model no longer exists in any provider.

## Root cause

`DEFAULT_MODEL` in `constants.ts` still references `opencode/big-pickle`. When a workspace is first loaded with no model explicitly chosen, `applyDefault()` reads `"opencode/big-pickle"` from `opencode.jsonc` and sets it as `defaultModel()`. OpenCode then rejects every session with an unknown model error.

### `constants.ts`
- Replace removed `opencode/big-pickle` with `anthropic/claude-sonnet-4-5` as the stable fallback placeholder.

### `model-config.ts` — two-part stale-model reconciliation

**In `applyDefault()`** (providers already loaded when workspace config is read):
- Validate `configDefault` against the live provider list before applying it.
- If the model is not found, treat `configDefault` as `null` — the workspace has no valid configured model.

**New `createEffect`** (race condition: providers load *after* `applyDefault()` has already run):
- Once providers arrive, re-validate `defaultModel()`.
- If it no longer exists in any provider (e.g. a removed alias), clear the explicit override and reset to `DEFAULT_MODEL` so the user can select a valid model from their connected providers.

## Behavior after fix

| Scenario | Before | After |
|---|---|---|
| New workspace, no model set | Defaults to `opencode/big-pickle` → chat fails | Defaults to `anthropic/claude-sonnet-4-5` |
| Existing workspace with stale `opencode/big-pickle` in config | Chat fails silently | Stale model is cleared; user selects from connected providers |
| Workspace with valid model in config | Works | Works (unchanged) |

Closes #1237

### DIFF
--- a/apps/app/src/app/constants.ts
+++ b/apps/app/src/app/constants.ts
@@ -7,9 +7,12 @@ export const VARIANT_PREF_KEY = "openwork.modelVariant";
 export const LANGUAGE_PREF_KEY = "openwork.language";
 export const HIDE_TITLEBAR_PREF_KEY = "openwork.hideTitlebar";
 
+// UI placeholder shown before providers load. Must reference a real, stable model.
+// If this model is not available for the user, the stale-model reconciliation in
+// model-config.ts will clear it and let the user pick from their connected providers.
 export const DEFAULT_MODEL: ModelRef = {
-  providerID: "opencode",
-  modelID: "big-pickle",
+  providerID: "anthropic",
+  modelID: "claude-sonnet-4-5",
 };
 
 export const SUGGESTED_PLUGINS: SuggestedPlugin[] = [
--- a/apps/app/src/app/context/model-config.ts
+++ b/apps/app/src/app/context/model-config.ts
@@ -541,6 +541,14 @@ export function createModelConfigStore(options: {
     return provider?.models?.[ref.modelID] ?? null;
   };
 
+  // Returns true if the model exists in the loaded provider list.
+  // Returns true when providers are not yet loaded (defer validation until they arrive).
+  const isModelRefValid = (ref: ModelRef) => {
+    const providers = options.providers();
+    if (!providers.length) return true;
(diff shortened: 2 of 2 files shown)


src-opn:
defaulting to `anthropic/claude-sonnet-4-5` ?

src-opn:
`opencode/big-pickle` still works with the embedded opencode version

we have not tested on later versions of opencode

closing for now

PR #1384, opened 2026-04-07
fix(settings): correct ready-to-install update label

src-opn:
## Summary
- fix the settings update badge copy by removing a stray template fragment from the English locale string
- keep the existing version suffix behavior so the ready state renders as `Ready to install · v...`

## Verification
- confirmed the PR diff only updates `apps/app/src/i18n/locales/en.ts`
- attempted `pnpm --filter @openwork/app typecheck`, but this worktree is missing dependencies and failed during `pretypecheck` because `tsup` was not installed


johnnyshields:
FYI this was originally broken in commit https://github.com/different-ai/openwork/commit/52f282b94d0e1ba1852617f813704970b087244b

PR #1403, opened 2026-04-08
fix(i18n): make english source-first across the app

jcllobet:
## Summary
- add a backward-compatible `td(key, english, localeOrParams?, params?)` helper so visible English copy lives at the callsite while locale keys stay stable
- migrate static app i18n callsites to source-first defaults, including shared/common labels, so grep-by-English now points back to the real source files
- add `--source-first`, `--extract-source-first`, and `pnpm i18n:sync-en` so `en.ts` is verified against and synced from source-first defaults
- keep one explicit dynamic fallback in `apps/app/src/app/app.tsx` for unknown future busy-status keys; all static callsites now carry inline English defaults

## Testing
- `pnpm install`
- `pnpm test:i18n:find`
- `pnpm test:i18n:source-first`
- `node scripts/i18n-audit.mjs --source-first`
- `pnpm i18n:sync-en`
- `pnpm i18n:find -- "OpenWork server URL"`
- `pnpm i18n:find -- "Add Remote Workspace"`
- `node scripts/i18n-audit.mjs --duplicates`
- `pnpm build:ui`

## Result
- `scripts/i18n-audit.mjs --source-first` now reports `1296` source-first defaults in sync with `en.ts`
- static app i18n callsites now require inline English defaults, so the codebase stays searchable while preserving localization

## Notes
- common/shared strings are still translated; `common.*` remains a namespace for shared UI copy, not an escape hatch from localization
- I did not run Docker/Chrome MCP because this is an i18n/source-ownership refactor rather than a new end-user flow


jcllobet:
do not merge, @johnnyshields has a better take on it

PR #1430, opened 2026-04-12
fix(desktop): correct truncate_output for multibyte (CJK) characters

ahuachen:
## Summary

- `truncate_output` in `src/utils.rs` used `input.len()` (byte length) as the guard condition, but `chars().count()` (char count) for the skip offset
- For CJK characters (Chinese/Japanese/Korean), each character is 3 bytes — so `len > max_chars` could pass while `chars().count() < max_chars`, causing a `usize` underflow **panic** when expanding session history containing CJK content
- Fix: compute `char_count` once via `chars().count()` and use it for both the guard and the skip offset
- Also adds `dev:local` / `build:local` convenience scripts for running the app without Docker

## Test plan

- [ ] Open the app with session history containing Chinese/Japanese text
- [ ] Expand historical conversations — should no longer crash
- [ ] Verify `truncate_output` correctly truncates CJK strings

### DIFF
--- a/apps/desktop/src-tauri/src/engine/spawn.rs
+++ b/apps/desktop/src-tauri/src/engine/spawn.rs
@@ -98,7 +98,6 @@ pub fn spawn_engine(
         let dev_paths = resolve_dev_mode_paths(app)?;
         command = command.env("OPENWORK_DEV_MODE", "1");
         command = command.env("OPENCODE_TEST_HOME", &dev_paths.home_dir);
-        command = command.env("HOME", dev_paths.home_dir);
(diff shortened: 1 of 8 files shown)


jcllobet:
@ahuachen , if I understand correctly, the core commit is just related to apps/desktop/src-tauri/src/utils.rs, right? Why remove the home? And what's the value of the other added files and tests? They don't seem to be relevant to the pr and just add extra noise.

src-opn:
@johnnyshields Does this issue exist for JP translations? I remember something was merged to fix this

PR #1439, opened 2026-04-12
fix(session): restore transcript autoscroll growth detection

pascalandr:
## Summary
- remove the accidental `h-full` sizing from the Solid session transcript content wrapper
- restore the scroll controller's ability to observe real transcript height growth and keep follow-scroll pinned to the latest message
- avoid the recent regression where the message pane stops slightly above the bottom while new text is still arriving

## Likely Regression Source
- `9784c618` originally introduced the `h-full` wrapper change in `apps/app/src/app/pages/session.tsx`
- `5b9da8a8` re-applied that same change onto current `dev`

## Testing
- `pnpm --filter @openwork/app exec vite build`
- `pnpm --filter @openwork/desktop run prepare:sidecar`
- `pnpm --filter @openwork/desktop exec tauri build --ci --no-sign --config '{"build":{"beforeBuildCommand":""}}'`
- launched `apps/desktop/src-tauri/target/release/OpenWork-Dev.exe` for manual verification

### DIFF
--- a/apps/app/src/app/pages/session.tsx
+++ b/apps/app/src/app/pages/session.tsx
@@ -3266,7 +3266,7 @@ export default function SessionView(props: SessionViewProps) {
                 }}
               >
                 <div
-                  class="mx-auto h-full w-full max-w-[800px]"
+                  class="mx-auto w-full max-w-[800px]"
                   ref={(el) => {
                     chatContentEl = el;
                   }}


jcllobet:
Seems like this works if the following is our expected behaviour between the `auto-scroll` and `manual browse`. Merging:

```
 Switching to manual-browse (lines 153-159): When the user scrolls up (scroll   
  gesture detected + scrollTop decreasing + more than 96px from bottom), mode    
  flips to "manual-browse".                                                      
                                                                       
  Auto-scroll only fires in follow-latest (lines 201-203):                       
  if (grew && isAtBottom()) {
      scrollToBottom("auto");                                                    
  }                                                                    
   
  So the ResizeObserver detects content growth, but it only auto-scrolls if the 
  user hasn't scrolled away. If you're reading earlier messages, mode is         
  "manual-browse" and the growth is ignored.
  ```
  
 ** Edge case to revisit**: FOLLOW_LATEST_BOTTOM_GAP_PX = 96 in scroll-controller.ts
  ``` 
  means if you scroll up within ~96px of the streaming bottom of a long message  
  (e.g. to re-read a tool call while the LLM is still writing), auto-scroll yanks
   you back down. Not a regression from this PR — the h-full bug just masked it —
   but worth noting if we get UX complaints about fighting the scroll during long
   streaming messages.  
   
   Long streaming message — you scroll up to check a tool call                    
  ┌─────────────────────────┐                                          
  │  │ ... tool call output │ │ ◄─ you want to read this                         
  │  │ ... more output      │ │                                                  
  │  │ ... result           │ │                                                  
  │  └──────────────────────│ │                                                  
  │  │ ... new streamed text│ │ ◄─ LLM still writing here              
  └──────────────────────────┘                                                   
           ▲                                                                     
      bottom of viewport                                                         
                                                                                 
    scrollHeight - (scrollTop + clientHeight) = bottomGap                        
   
   ```
   
   Potential alternatives in case the issues come exactly from this behaviour:
    - Shrink `FOLLOW_LATEST_BOTTOM_GAP_PX` (e.g., 32px) — tighter snap zone, easier  
  to break free                                                                  
  - Only re-engage follow-latest on explicit `scroll-to-bottom` (button click), not
   on proximity                                                                  
  - Add a brief `debounce` after a user scroll gesture before re-engaging follow   
  mode

PR #1461, opened 2026-04-16
fix(automations): clear stale scheduler install gate

### LINKED ISSUE #1422: [Bug]: Install scheduler reports success but does not unlock automations
**Title:** [Bug]: `Install scheduler` reports success but does not unlock automations

**Summary:** *
The UI can show `Scheduler install requested.` while Automations stays blocked or the reload and install state behaves inconsistently.

**Steps to reproduce:** *
1. Open `Settings` -> `Automations` in a workspace where `opencode-scheduler` is not installed.
2. Click `Install scheduler`.
3. Observe the toast and the Automations gate state.
4. Stay on the page or navigate away and back without restarting OpenWork.

**Expected behavior:** *
The scheduler is added successfully and the page consistently prompts for reload, or the user sees a direct error explaining why installation failed.

**Actual behavior:** *
The UI shows `Scheduler install requested.` but Automations can remain locked, and the install or reload state can appear stuck, reset, or non-functional.

---
Label: bug

pascalandr:
## Summary
- clear the local `schedulerInstallRequested` gate once the scheduler plugin actually appears as installed
- also clear that gate when plugin installation fails, so the Automations UI does not stay stuck in a pseudo-installed state
- keep the existing reload/install flow intact while removing the stale local state that could block automations after a successful install

## Context
- fixes #1422

## Testing
- `pnpm --filter @openwork/app exec vite build`

### DIFF
--- a/apps/app/src/app/pages/automations.tsx
+++ b/apps/app/src/app/pages/automations.tsx
@@ -441,6 +441,12 @@ export default function AutomationsView(props: AutomationsViewProps) {
     onCleanup(() => window.clearInterval(interval));
   });
 
+  createEffect(() => {
+    if (props.schedulerInstalled) {
+      setSchedulerInstallRequested(false);
+    }
+  });
+
   const showToast = (title: string, tone: AppStatusToastTone = "info") => {
     statusToasts.showToast({ title, tone });
   };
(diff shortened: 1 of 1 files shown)


src-opn:
Thanks!

PR #1466, opened 2026-04-16
fix(i18n): restore share_skill_destination strings for en and fr

jcllobet:
## Summary

The skill-destination modal (shown when importing a shared skill via a share link or deep link) was rendering raw translation keys like `share_skill_destination.add_to_workspace` in the primary CTA and related labels for **English** and **French** users.

Five keys were stubbed in a previous i18n cleanup (#1345) with their own key names as the translation value, and `fr.ts` inherited the same stubs when it was added. All other locales (ca, es, ja, pt-BR, th, vi, zh) already have real translations and are unaffected.

### Keys restored

| Key | English | French |
|---|---|---|
| `share_skill_destination.add_to_workspace` | Add skill to workspace | Ajouter le Skill à l'espace de travail |
| `share_skill_destination.adding` | Adding skill... | Ajout du Skill... |
| `share_skill_destination.connect_remote_hint` | Attach an OpenWork host, then choose it from the list to import this skill. | Attachez un hôte OpenWork, puis choisissez-le dans la liste pour importer ce Skill. |
| `share_skill_destination.create_worker_hint` | Open the workspace setup flow, then add this skill after the new workspace is ready. | Ouvrez le flux de configuration de l'espace de travail, puis ajoutez ce Skill une fois que le nouvel espace de travail est prêt. |
| `share_skill_destination.more_options` | More options | Plus d'options |

The wording for the hint/CTA keys mirrors the existing `confirm_button`, `confirm_busy`, `create_worker_desc`, and `connect_remote_desc` entries that these newer keys supersede in `apps/app/src/app/bundles/skill-destination-modal.tsx`.

## Scope

- Files touched: `apps/app/src/i18n/locales/en.ts`, `apps/app/src/i18n/locales/fr.ts`
- No component or logic changes.
- Non-English locales fall back to English, so the worst case (before) for users of unshipped locales was already handled by fallback; this PR ensures no locale falls through to the raw key.

## Verification

- Static review: confirmed via `grep` that no locale still has any `share_skill_destination.*` value equal to its own key (pattern `"<key>": "<key>"`).
- Confirmed all 20 keys referenced by `apps/app/src/app/bundles/skill-destination-modal.tsx` are defined in `en.ts` and `fr.ts`.
- **End-to-end (Docker + Chrome MCP): not run in this environment** — Docker daemon was not running on the host. The exact user flow to verify locally:
  1. `packaging/docker/dev-up.sh`
  2. Open the app and navigate to Settings → Language; pick English or French.
  3. Open any `https://openwork.so/share/<id>` link for a shared skill and click "Open in OpenWork", or fire a bundle deep link.
  4. The modal's confirm button should read "Add skill to workspace" (en) / "Ajouter le Skill à l'espace de travail" (fr), the "More options" section heading should render, and both option tiles (Create new workspace / Connect remote workspace) should show their subtitle hints.

## Screenshot reference (pre-fix)

Reporter's screenshot showed the CTA rendered as `share_skill_destination.add_to_workspace` in English.


jcllobet (commented):
lgtm

PR #1481, opened 2026-04-17
docs: make get started app-first and add missing languages page

jcllobet:
## Summary
- make the Start Here entrypoint app-first so the main OpenWork CTA is clearer and team/cloud guidance no longer leads the page
- move self-host into its own docs navigation group and collect docs-improvement links together
- add a Missing Languages page with the video, current locale list, PR invitation, and i18n roadmap link

## Testing
- `git diff --check`
- `node -e \"JSON.parse(require('fs').readFileSync('packages/docs/docs.json','utf8')); console.log('docs.json ok')\"`

## Notes
- Docs-only change; no app/runtime flow changed.

### DIFF
--- a/packages/docs/docs.json
+++ b/packages/docs/docs.json
@@ -37,8 +37,19 @@
               "start-here/do-work-with-it/share-your-setup"
             ]
           },
-          "start-here/self-host",
-          "start-here/missing-docs"
+          {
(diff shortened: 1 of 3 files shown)


jcllobet (commented):
lgtm

PR #1482, opened 2026-04-17
feat(den): rename organization from Members page

benjaminshafii:
## Summary

- Adds a **PATCH `/v1/orgs/:orgId`** endpoint (owner-only) on `den-api` to update an organization's display name. Slug stays immutable so dashboard URLs do not break.
- Wires a new `updateOrganizationName` mutation through `OrgDashboardProvider` so the web app can call the endpoint.
- Adds an **Organization** settings card at the top of the Members page (`/o/:orgSlug/dashboard/members`). Owners see a Rename button that swaps in an inline edit form; non-owners see a read-only summary explaining who can rename.
- The sidebar org chip + dashboard header reflect the new name immediately because they all read from the refreshed org context.
- Includes a companion fix (`fix(den-api)`) that tolerates a missing `apps/desktop/package.json` when building the den-api container. PR #1476 added a build step that reads the Tauri package.json, but `packaging/docker/Dockerfile.den` does not ship those sources, which had broken `packaging/docker/den-dev-up.sh`. We fall back to \`0.0.0\` (matching the runtime default) and let deployers override via `DEN_API_LATEST_APP_VERSION`.

## UX notes

- **Scope**: name only, owners only. Slug is intentionally not editable to keep \`/o/:slug\` links stable.
- **Guardrails**: API validates \`name\` (2–120 chars, trimmed) and requires \`ensureOwner\`; UI uses the same min length plus a \`maxLength=120\`.
- **Feedback**: on success the card collapses and shows \"Organization renamed.\"; on failure we surface a red page-level error banner (same pattern as other Members mutations).

## Screenshots

Before (Members page with Organization card + Rename button, owner):

before

Editing (inline form with current name pre-filled):

editing

After (new name + success state, sidebar also updated):

after

Screenshots are also committed under \`ee/apps/den-web/docs/screenshots/\`.

## Testing

End-to-end via Chrome MCP against the Docker dev stack:

1. \`packaging/docker/den-dev-up.sh\` → healthy at web=64449, api=64448 (verified \`/\` = 200, \`/health\` = 200).
2. Signed up fresh (\`rename-test@openwork.dev\`), completed OTP verification (code pulled from \`docker logs\`).
3. Created org \"Acme Starter\", navigated to Members.
4. Clicked **Rename**, changed name to \"Acme Inc.\", clicked **Save name** → UI updated, \"Organization renamed.\" success shown, sidebar chip flipped to \"AI Acme Inc. Owner\".
5. Verified persistence in MySQL: \`SELECT id, name, slug FROM organization\` returned the new name and the unchanged slug.
6. Re-opened the form and tried \`x\` (1 char) → browser HTML5 validation blocked the submit (\"Please lengthen this text to 2 characters or more\").
7. Navigated back to Dashboard → org badge reads \"Acme Inc.\"; URL path still contains the original \`org_01kpev9w...\` slug.
8. No console errors during the flow.

Commands for reviewers:

\`\`\`bash
cd _repos/openwork
packaging/docker/den-dev-up.sh      # prints web/api URLs and env file path
# then Chrome MCP:
#   /  → signup (use OTP from \`docker logs <den>\`)
#   create org → /o/<slug>/dashboard/members
#   click Rename → edit → Save
\`\`\`

## Out of scope

- Renaming the slug / changing the URL path.
- Non-owner rename permissions (e.g. admins).
- Changing the logo/avatar.


benjaminshafii:
Added a follow-up commit: extracted a reusable `DenCard` primitive under `ee/apps/den-web/app/(den)/_components/ui/card.tsx` and removed the inline drop shadows that had been copy-pasted on every Members/API Keys card (Den visual language does not use drop shadows).

- Two explicit size presets (`comfortable` — current Members/API Keys style, `spacious` — editor-screen style) so we can keep standardising the rest of the dashboard in follow-up PRs.
- No shadow utilities are emitted by the primitive; consumers pass layout classes only.
- Swept 6 cards total (5 in `manage-members-screen.tsx`, 1 in `api-keys-screen.tsx`) which all rendered with the identical inline class string.
- Re-captured the members-rename screenshots against the refactored UI.

Other screens still using inline shadow-laden cards are out of scope for this PR; I left them for a follow-up so this diff stays reviewable.

Issue #652, opened 2026-02-23
[Bug]: Openwork does not terminate OpenCode process on close, leading to multiple instances and memory leak on macOS Apple Silicon


jack118:
### Summary

When opening and then closing the Openwork application on macOS (Apple Silicon), the associated OpenCode process is not terminated. Repeatedly opening and closing Openwork results in multiple OpenCode processes accumulating in the background, causing a steady increase in memory usage over time. This eventually leads to excessive memory consumption and may degrade system performance.

### Steps to reproduce

1. Launch Openwork on a Mac with Apple Silicon (M1/M2/M3).
2. Observe that an OpenCode process is started (e.g., via Activity Monitor or ps command).
3. Close Openwork normally (e.g., by clicking the red close button or using Quit).
4. Check if the OpenCode process is still running.
5. Repeat steps 1–4 several times.
6. Monitor the number of OpenCode processes and memory usage.

### Expected behavior

Closing Openwork should terminate all associated processes, including OpenCode, so that no orphaned processes remain. Memory should be freed accordingly.

### Actual behavior

The OpenCode process persists after Openwork is closed. Each subsequent launch of Openwork spawns a new OpenCode instance, leaving previous instances running. Over multiple open/close cycles, numerous OpenCode processes accumulate, consuming increasing amounts of memory.

benjaminshafii:
@jack118 Thanks for noticing. I'll make sure this is fixed in next update.

PR #1497, opened 2026-04-19
docs(desktop): explain why we match both ExitRequested and Exit on quit

jcllobet:
## What this PR actually ships (final)

A **6-line inline comment** in `apps/desktop/src-tauri/src/lib.rs` documenting why the cleanup arm matches both `RunEvent::ExitRequested` and `RunEvent::Exit`, pinned to upstream [tauri-apps/tauri#9198](https://github.com/tauri-apps/tauri/issues/9198). No functional code changes.

## Why not the original fix

The PR initially tried two changes to silence a duplicate `[orchestrator] Failed to request shutdown ... Connection refused` log line on macOS red-X close. Both were reverted in this branch after deeper review:

1. **`71670825` — dropping `RunEvent::Exit` from the cleanup arm** (reverted in `15d1f3f4`).
   On macOS, the default menu's Quit, Cmd+Q, and dock → Quit fire only `Exit`, not `ExitRequested` ([tauri#9198](https://github.com/tauri-apps/tauri/issues/9198)). Dropping `Exit` would silently skip cleanup for those quit paths, orphaning orchestrator / opencode / openwork-server sidecars — exactly the failure mode fixed by [#655](https://github.com/different-ai/openwork/pull/655).

2. **`f68ee858` — early-returning from `OrchestratorManager::stop_locked` when `state.child` is `None`** (reverted in `b57642ec`).
   `stop_locked` is intentionally called with `state.child == None` in two places:
   - `engine_start` (`commands/engine.rs:386-389`) calls it before spawning a new daemon, to reap a surviving orchestrator from a crashed previous session.
   - `engine.rs:152` comment: *"The orchestrator can keep running across app relaunches."*
   
   Early-returning there would leave a surviving daemon running while the desktop spawns a second one — the exact orphan-process / memory-leak regression from [#652](https://github.com/different-ai/openwork/issues/652) that [#655](https://github.com/different-ai/openwork/pull/655) fixed.

## Outstanding

The original symptom (a spurious "Connection refused" log line on red-X close when a stale `base_url` is in the state file but no orchestrator is live) is **not fixed by this PR**. It remains cosmetic — no functional impact. A proper follow-up would probe `GET /health` before POSTing `/shutdown` inside `request_orchestrator_shutdown`, so a dead URL returns `Ok(false)` silently instead of logging. That's out of scope here.

## Test plan

- [x] `cargo check --no-default-features` on `apps/desktop/src-tauri` — clean
- [x] `pnpm --filter @openwork/app test:e2e` on macOS — all steps `ok`
- [x] Net diff vs `dev` is comment-only: `apps/desktop/src-tauri/src/lib.rs | 6 ++++++`

## Rollback

```bash
gh pr revert 1497
```

Comment-only change, nothing to migrate.

### DIFF
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -223,8 +223,11 @@ pub fn run() {
     // Best-effort cleanup on app exit. Without this, background sidecars can keep
     // running after the UI quits (especially during dev), leading to multiple
     // orchestrator/opencode/openwork-server processes and stale ports.
+    // Only run on ExitRequested — Exit fires immediately after and would re-run
+    // the same teardown, causing duplicate shutdown HTTP requests against an
+    // already-stopped orchestrator.
     app.run(|app_handle, event| match event {
-        RunEvent::ExitRequested { .. } | RunEvent::Exit => stop_managed_services(&app_handle),
+        RunEvent::ExitRequested { .. } => stop_managed_services(&app_handle),
         // On macOS the default behavior is to keep the process alive after the
         // last window closes. We want parity with Windows/Linux: closing the
         // main window quits the app.
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -223,6 +223,12 @@ pub fn run() {
     // Best-effort cleanup on app exit. Without this, background sidecars can keep
     // running after the UI quits (especially during dev), leading to multiple
     // orchestrator/opencode/openwork-server processes and stale ports.
+    //
+    // We intentionally match both ExitRequested AND Exit: on macOS the built-in
+    // Quit menu, Cmd+Q, and dock -> Quit only fire RunEvent::Exit — they don't
+    // fire ExitRequested (tauri-apps/tauri#9198). The red-X path fires both
+    // because we translate WindowEvent::CloseRequested into app_handle.exit(0)
+    // below. stop_managed_services is idempotent, so firing on both is safe.
     app.run(|app_handle, event| match event {
         RunEvent::ExitRequested { .. } | RunEvent::Exit => stop_managed_services(&app_handle),
         // On macOS the default behavior is to keep the process alive after the
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -223,11 +223,8 @@ pub fn run() {
     // Best-effort cleanup on app exit. Without this, background sidecars can keep
     // running after the UI quits (especially during dev), leading to multiple
     // orchestrator/opencode/openwork-server processes and stale ports.
-    // Only run on ExitRequested — Exit fires immediately after and would re-run
-    // the same teardown, causing duplicate shutdown HTTP requests against an
-    // already-stopped orchestrator.
     app.run(|app_handle, event| match event {
-        RunEvent::ExitRequested { .. } => stop_managed_services(&app_handle),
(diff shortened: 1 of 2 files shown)


jcllobet:
### What this fixes (and how to roll back)

**Symptom:** every macOS quit logged this twice:
```
[orchestrator] Failed to request shutdown: Failed to request orchestrator shutdown at http://127.0.0.1:NNNNN/shutdown: Connection Failed: Connect error: Connection refused (os error 61)
```

**Root causes:**
1. `RunEvent::ExitRequested | RunEvent::Exit` both ran the teardown — Exit fires immediately after ExitRequested, so cleanup ran twice per close.
2. `OrchestratorManager::stop_locked` always tried to POST `/shutdown`, even in sessions where we never spawned the orchestrator. In that case it would read a stale `daemon.base_url` from the on-disk state file (left over from a previous run) and POST to a port nothing was listening on.

**Fixes (one commit each):**
- 71670825 — drop `RunEvent::Exit` from the cleanup match arm.
- f68ee858 — early-return from `stop_locked` when `state.child.is_none()` so we only attempt graceful shutdown for daemons we actually own this session.

**Rollback:**
- Revert just the early-return: `git revert f68ee858` (restores the always-POST behavior; brings back the spurious log line but otherwise harmless).
- Revert just the RunEvent change: `git revert 71670825` (restores the double-teardown).
- Revert the whole PR after merge: `gh pr revert 1497` or `git revert -m 1 <merge-sha>`.

Neither change touches persisted state, schemas, IPC commands, or the graceful-shutdown HTTP contract — both are safe to revert at any time without data migration.

jcllobet:
### Correction — reverted my own first commit

After asking what we could learn from other Tauri apps, I dug into [tauri-apps/tauri#9198](https://github.com/tauri-apps/tauri/issues/9198) and found that `RunEvent::ExitRequested` does not fire on macOS for Cmd+Q, the App menu → Quit, or the dock → Quit when using the default menu (which we do). My original commit `71670825` removed `RunEvent::Exit` from the cleanup arm, which would have silently broken cleanup for all three of those common quit paths — leaving orchestrator / opencode / openwork-server sidecars orphaned.

Reverted in `15d1f3f4`. The idempotent `OrchestratorManager::stop_locked` from `f68ee858` is doing all the real work here — it's what kills the duplicate log line. Added an inline comment in `143e75a7` so the next reviewer doesn't repeat the mistake.

#### Updated rollback

- Revert the orchestrator manager fix: `git revert f68ee858` (restores the spurious log line but otherwise harmless)
- Revert the whole PR: `gh pr revert 1497` — safe at any time, no persisted-state migrations involved

jcllobet:
### Reverted `f68ee858` as well — net diff is now comment-only

Git archaeology on [#655](https://github.com/different-ai/openwork/pull/655) (the commit that wrote the current `stop_locked`) revealed my early-return would regress the reap-before-spawn path in `engine_start` (`commands/engine.rs:386-389`) and the cross-relaunch adoption case (`engine.rs:152` comment: "The orchestrator can keep running across app relaunches"). Both paths deliberately call `stop_locked` with `state.child == None` and rely on the HTTP `/shutdown` POST to kill a surviving daemon from a crashed previous session — exactly the scenario [#652](https://github.com/different-ai/openwork/issues/652) was fixing.

Reverted in `b57642ec`. Verified locally: `cargo check` clean, `pnpm --filter @openwork/app test:e2e` all `ok`.

**Net effect of this PR is now a 6-line inline comment** explaining why we match both `ExitRequested` and `Exit` on macOS quit. The original symptom (cosmetic duplicate log line on red-X close when the state file has a stale `base_url`) is left for a follow-up that probes `GET /health` before POSTing `/shutdown`.

PR #1514, opened 2026-04-22
fix(config): respect .opencode opencode.jsonc paths

### LINKED ISSUE #824: [Bug]: opencode.jsonc in .opencode/ subdirectory is ignored, causing duplicate config files
### Summary

When a project initialized with OpenCode it creates `.opencode/opencode.jsonc` as its local configuration file (which is officially supported by OpenCode), OpenWork fails to detect it.

Instead, upon worker initialization or when modifying settings (like adding an MCP server), OpenWork automatically creates a new `opencode.jsonc` file directly in the workspace root. This results in two competing configuration files in the same workspace (`.opencode/opencode.jsonc` and `./opencode.jsonc`), which breaks automatic settings synchronization between OpenCode and OpenWork unless a symlink is manually created.

### Steps to reproduce

The discrepancy comes from how config paths are resolved. In OpenCode (e.g., resolveConfigPath), the fallback chain explicitly checks the .opencode directory:

```
const candidates = [
  path.join(baseDir, "opencode.json"), 
  path.join(baseDir, "opencode.jsonc"),
  path.join(baseDir, ".opencode", "opencode.json"), 
  path.join(baseDir, ".opencode", "opencode.jsonc")
]
```

However, OpenWork strictly checks only the root directory across multiple core files (both in Tauri Rust backend and TypeScript server).

### Expected behavior

Еxisting config (.opencode/opencode.jsonc) is used, no duplicates created

### Actual behavior

OpenWork automatically creates a new opencode.jsonc file directly in the workspace root even if there is `.opencode/opencode.jsonc`

pascalandr:
## Summary
- resolve project-scoped OpenCode config from `.opencode/opencode.jsonc` and `.opencode/opencode.json` in both desktop and server paths
- keep the lookup order aligned with OpenCode's config resolution behavior

Closes #824.

## Testing
- `cargo test --manifest-path "apps/desktop/src-tauri/Cargo.toml" config`

## Notes
- I did not run a full app/Docker flow for this config-path change.
- Manual follow-up: `bash packaging/docker/dev-up.sh`, open a workspace that only has `.opencode/opencode.jsonc`, and confirm the config is discovered without fallback renames.

### DIFF
--- a/apps/desktop/src-tauri/src/config.rs
+++ b/apps/desktop/src-tauri/src/config.rs
@@ -7,14 +7,19 @@ use crate::types::{ExecResult, OpencodeConfigFile};
 fn opencode_config_candidates(
     scope: &str,
     project_dir: &str,
-) -> Result<(PathBuf, PathBuf), String> {
+) -> Result<Vec<PathBuf>, String> {
     match scope {
         "project" => {
             if project_dir.trim().is_empty() {
                 return Err("projectDir is required".to_string());
             }
             let root = PathBuf::from(project_dir);
-            Ok((root.join("opencode.jsonc"), root.join("opencode.json")))
+            Ok(vec![
+                root.join("opencode.jsonc"),
+                root.join("opencode.json"),
+                root.join(".opencode").join("opencode.jsonc"),
+                root.join(".opencode").join("opencode.json"),
+            ])
         }
         "global" => {
             let base = if let Ok(dir) = env::var("XDG_CONFIG_HOME") {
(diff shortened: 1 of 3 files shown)


src-opn (changes requested):
Security audit: I scanned the raw diff and changed files for suspicious Unicode/non-ASCII, bidi controls, zero-width characters, C0/C1 controls, and whitespace-only anomalies. `git diff --check` is clean and I did not find hidden/malformed text, prompt-like injected content, secret exfiltration logic, or unrelated/suspicious functions in the changed lines.

Requesting changes for an incomplete fix.

Blocking issue:
- `apps/desktop/src-tauri/src/workspace/files.rs:424-432` still resolves only `opencode.jsonc` / `opencode.json` at the workspace root during `ensure_workspace_files`. The linked bug is specifically that initialization can create a duplicate root config when `.opencode/opencode.jsonc` already exists. This PR fixes the newer config helper and the TS server path, but the desktop workspace initialization path still has the old behavior, so desktop-hosted workspace setup can still reproduce the bug.

Verification run:
- `pnpm --filter openwork-server typecheck` passed.
- `pnpm --filter openwork-server build` passed.
- `pnpm --filter openwork-server test` passed: 83 tests.
- `cargo test --manifest-path "apps/desktop/src-tauri/Cargo.toml" config` passed but matched 0 tests, so it does not validate this change.
- `cargo test --manifest-path "apps/desktop/src-tauri/Cargo.toml"` passed: 29 tests.

Please update the desktop `ensure_workspace_files` config resolution to include `.opencode/opencode.jsonc` and `.opencode/opencode.json` consistently, and ideally add focused coverage for the duplicate-config case.

pascalandr:
desktop workspace init now respects .opencode/opencode.jsonc/.json too, with focused Rust tests added.

PR #1515, opened 2026-04-22
fix(workspace): reload opencode.jsonc on initial activate

### LINKED ISSUE #870: [Bug]: Permissions in opencode.jsonc are ignored on every launch

pascalandr:
## Summary
- emit a `config` reload event when `POST /workspaces/:id/activate` completes
- trigger `activateWorkspace` during initial hydration and workspace switching so `.opencode/opencode.jsonc` changes are reloaded consistently

Closes #870.

## Testing
- `pnpm --filter @openwork/app typecheck` currently fails on `upstream/dev` with pre-existing React cutover resolution/type errors; I reproduced the same failure on detached `upstream/dev`

## Notes
- I did not run the Docker + Chrome MCP flow for this UI/runtime interaction.
- Manual follow-up: `bash packaging/docker/dev-up.sh`, switch into a workspace with a changed `.opencode/opencode.jsonc`, and confirm the config reload event is applied without restarting the app.


src-opn (changes requested):
Security verification: I did not find malicious code, hidden prompt-like text, bidi/zero-width/control characters, or non-ASCII bytes in the added lines. I also compared the raw diff against an ASCII-visible diff and ran git diff --check; no added-line character or whitespace issues were found. The only non-ASCII bytes observed were pre-existing context comments, not PR additions.

Requesting changes for a functional/code-verification issue:

- apps/server/src/server.ts:1514 records a debounced reload event from POST /workspaces/:id/activate, but that does not actually reload the OpenCode engine. In this codebase the actual reload path is POST /workspace/:id/engine/reload, which calls reloadOpencodeEngine at apps/server/src/server.ts:2589-2594. The event endpoint at apps/server/src/server.ts:2581-2586 only lists queued events, and I could not find a current app-side consumer of listReloadEvents beyond the client wrapper at apps/app/src/app/lib/openwork-server.ts:1340. As written, this PR appears to enqueue a config event and then swallow the activate call result in the app, but it does not force the engine to re-read opencode.jsonc. That means issue #870 can still reproduce on launch unless some other unshown process consumes the event and calls the reload endpoint.

Please either wire the event into the reload flow with a tested consumer, or call the server-side engine reload path from activation when that is the intended behavior. Add/adjust tests that prove activating a workspace with changed opencode.jsonc causes the runtime instance to reload/re-read config.

Verification performed locally in a dedicated worktree:

- pnpm install --frozen-lockfile: passed
- pnpm --filter openwork-server typecheck: passed
- pnpm --filter @openwork/app typecheck: passed
- pnpm --filter openwork-server test: passed, 83 tests
- pnpm --filter openwork-server build: passed
- pnpm --filter @openwork/app build: passed, with existing Vite chunk-size/dynamic-import warnings

pascalandr:
workspace activation now calls the real engine reload path, with a focused e2e test added, and the session route now avoids re-triggering that reload on every refresh.

PR #1517, opened 2026-04-22
fix(engine): bypass HTTP proxy for loopback engine traffic

### LINKED ISSUE #1276: [Bug]:OpenCode engine fails to start on Windows when using HTTP proxy
### Summary

On Windows, the OpenWork desktop application is unable to successfully start or connect to the OpenCode engine. The system environment uses an HTTP proxy for internet access, which seems to interfere with the engine's initialization or the internal communication between OpenWork and OpenCode.

### To Reproduce

Install OpenWork on a Windows machine.

Configure a system-wide or environment-level HTTP proxy (e.g., set HTTP_PROXY=http://...).

Launch the OpenWork application.

Observe the status: The "OpenCode engine" remains in a "Failed" state and never becomes ready.

### Expected behavior

The OpenCode engine should be able to start and bypass the proxy for local communication (127.0.0.1), or respect the proxy settings for external model API calls without crashing the local process.

### Actual behavior

The engine fails to initialize. No clear error message is provided in the UI, but the application cannot perform any AI tasks because the underlying OpenCode service is unreachable.

### Screenshots (optional)

_No response_

_No response_

_No response_

pascalandr:
## Summary
- merge `NO_PROXY` and `no_proxy` values case-insensitively and always include loopback hosts
- prevent local engine traffic from being routed through user-configured HTTP proxies on Windows

Closes #1276.

## Testing
- `cargo test --manifest-path "apps/desktop/src-tauri/Cargo.toml" engine::spawn::tests`

## Notes
- This includes 3 unit tests covering empty, dedupe, and preserve-order cases.

### DIFF
--- a/apps/desktop/src-tauri/src/engine/spawn.rs
+++ b/apps/desktop/src-tauri/src/engine/spawn.rs
@@ -57,6 +57,65 @@ pub fn find_free_port() -> Result<u16, String> {
     Ok(port)
 }
 
+/// Merge an existing NO_PROXY/no_proxy list with the loopback hosts required
+/// for local OpenCode engine communication. Preserves the user's existing
+/// entries and order; only appends loopback hosts that are not already listed.
+pub(crate) fn merge_no_proxy(
+    no_proxy: Option<&str>,
+    no_proxy_lower: Option<&str>,
+    loopback_hosts: &[&str],
+) -> String {
+    let mut entries: Vec<String> = Vec::new();
+    for raw in [no_proxy, no_proxy_lower].into_iter().flatten() {
+        for part in raw.split(',') {
+            let trimmed = part.trim();
+            if !trimmed.is_empty()
+                && !entries
+                    .iter()
+                    .any(|existing| existing.eq_ignore_ascii_case(trimmed))
+            {
+                entries.push(trimmed.to_string());
+            }
+        }
+    }
+    for host in loopback_hosts {
+        if !entries
+            .iter()
+            .any(|existing| existing.eq_ignore_ascii_case(host))
+        {
+            entries.push((*host).to_string());
+        }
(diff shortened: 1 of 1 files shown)


jcllobet:
@OmarMcAdam surfacing this as it looks like relevant fix:

My doubts on wether it's valid or not: 
How will this work for enterprise windows machines with a pre-existing proxy? 
Is this going to raise a flag on their end or are they okay with the way we solve for this?

pascalandr:
> @OmarMcAdam surfacing this as it looks like relevant fix:
> 
> My doubts on wether it's valid or not: How will this work for enterprise windows machines with a pre-existing proxy? Is this going to raise a flag on their end or are they okay with the way we solve for this?

Hi @jcllobet,
The current patch only bypasses the proxy for loopback destinations (127.0.0.1, localhost, ::1) in the spawned OpenCode child process. The idea is to keep local traffic to the local engine off the proxy, while external API traffic still goes through the enterprise proxy as usual. So no problem because localhost never leaves the machine.

pascalandr:
The current patch only bypasses the proxy for loopback destinations (`127.0.0.1`, `localhost`, `::1`) in the spawned OpenCode child process. The idea is to keep local traffic to the local engine off the proxy, while external API traffic still goes through the enterprise proxy as usual. So this should not be a problem, since `localhost` never leaves the machine.

OmarMcAdam:
Hey @pascalandr 
We're going to be moving to Electron this week.
Maybe we can review once the move is done and ensure same thing is in place 🫡
Thanks for this

src-opn:
We are moving our desktop builds to electron and thus these changes will not apply.
Closing this PR out, but if the issue still exists with electron, please update PR or open a new one

PR #1535, opened 2026-04-23
fix(extensions): show directory-based project plugins

### LINKED ISSUE #1105: [Bug]: Extensions UI does not display plugins that come from local plugin directories
### Summary

OpenWork discovers OpenCode plugins on the backend, but the Extensions UI does not display plugins that come from local plugin directories.

### To Reproduce

1. Place an OpenCode plugin in the expected local plugin directory.
2. Open the same workspace in OpenWork.
3. Go to the Extensions / Plugins area.

### Expected behavior

Plugins discovered from local directories should appear in OpenWork, together with plugins declared in `opencode.json`.

### Actual behavior

Only plugins coming from `opencode.json` are shown. Directory-based plugins are not displayed.

### Screenshots (optional)

_No response_

-OpenWork version [0.11.177]
-OS [Fedora Linux 43 (KDE Plasma Desktop Edition) x86_64]

The current behavior seems to come from a UI filter that keeps only plugin items with `source === "config"` and `scope === "project"`, even though the backend plugin loader returns more than that. This makes it look like directory plugins are discovered internally but hidden in the UI.

pascalandr:
## Summary

Fixes #1105 by updating the Extensions settings store to include all project-scoped plugins returned by `openworkClient.listPlugins(...)`, deduped by spec, instead of only showing plugins whose source is `config`.

## Testing

- `git diff --check upstream/dev...fix/1105-local-plugin-visibility` OK

## Notes

- This is the React port of the local pre-cutover fix.


src-opn (changes requested):
Requesting changes.

Finding:
- `apps/app/src/react-app/domains/settings/state/extensions-store.ts:1306` and `apps/app/src/react-app/domains/settings/pages/plugins-view.tsx:250-267`: directory-discovered project plugins are now added to `pluginList` as bare strings, but `pluginList` does not preserve each item's `source`. The Plugins UI renders every entry with the same `Enabled` state and `Remove` action, and `removePlugin()` only removes config-backed plugins from `opencode.json` / `DELETE /workspace/:id/plugins/:name`. For a `dir.project` item, clicking Remove will either report `Plugin not found` or refresh back to the same item, which makes the new visible plugin look removable when it is not. Please preserve plugin metadata in state or otherwise render directory plugins read-only / source-aware, and only enable removal for config-backed plugins.

Verification performed:
- `git diff --check origin/dev...HEAD`
- byte-level scan of the changed patch for non-ASCII/control/Bidi/BOM characters

pascalandr:
Made directory plugins visible but read-only, and kept Remove only for config plugins.

PR #1536, opened 2026-04-23
fix(settings): honor remote skills write capability

### LINKED ISSUE #1405: [Bug]: "Read-only for skills" in web/remote deployments despite write capabilities
### Summary

The app displays a "OpenWork server is read-only for skills" hint when connected to a running instance (openwork-orchestrator in web mode or remote host mode), even when the server reports [skills.write: true].

This most probably occurs because openworkServerCanWriteSkills() is gated on openworkServerWorkspaceReady(), which requires a defined runtimeWorkspaceId. In web/remote-connected environments, the workspace readiness flow differs from local hosted setup, causing a false negative.

### To Reproduce

Run openwork-orchestrator as a remote backend (e.g., in a container behind Nginx).
Connect to it via a web-hosted client or remote Tauri client using the owner token.
Observe the "Read-only for skills" hint in the Skills tab in settings despite having proper write permissions.

### Expected behavior

The write-access hint should only appear if the server capabilities explicitly return [skills.write: false]

### Actual behavior

The hint appears because the workspace is not considered "ready" by the local logic, even though the server is connected and capable.

### Screenshots (optional)

_No response_

OpenWork version: 0.11.205
OS: macOS (client) / Linux (server)
Runtime: Web / Remote Connect

The following patch is a workaround. It decouples the UI hint from the strict workspace-readiness check and instead relies on raw connection status and capabilities:

```// apps/app/src/app/app.tsx

// Fix for canInstallSkillCreator
const canInstallSkillCreator = isRemoteWorkspace
-  ? openworkServerCanWriteSkills()
+  ? openworkStatus() === "connected" && (!resolvedOpenworkCapabilities() || Boolean(resolvedOpenworkCapabilities()?.skills?.write))
   : isTauriRuntime();

// Fix for skillsAccessHint / pluginsAccessHint
- : openworkServerCanWriteSkills() ? null : t("app.skills_hint_readonly")
+ : !resolvedOpenworkCapabilities() || resolvedOpenworkCapabilities()?.skills?.write ? null : t("app.skills_hint_readonly")
```

pascalandr:
## Summary

- Fixes #1405 by updating `apps/app/src/react-app/domains/connections/openwork-server-store.ts` so `openworkServerCanWriteSkills` and `openworkServerCanWritePlugins` follow connected server capabilities.
- Removes the `runtimeWorkspaceId` / `openworkServerWorkspaceReady` gating that incorrectly made remote/web deployments appear read-only for skills and plugins.

## Testing

- `git diff --check upstream/dev...fix/1405-remote-skills-write-capability` OK

## Notes

- This is the React port of the local pre-cutover fix.


src-opn (changes requested):
Requesting changes: this PR does not appear to fix the linked issue in the React settings route.

`apps/app/src/react-app/shell/settings-route.tsx:953-955` still passes `canInstallSkillCreator={!isRemoteWorkspace}` and `accessHint={isRemoteWorkspace ? t("app.skills_hint_readonly") : null}`. That means every remote workspace still shows the read-only skills hint regardless of `openworkServerCapabilities.skills.write`.

`apps/app/src/react-app/shell/settings-route.tsx:967-969` has the same problem for plugins: `canEditPlugins={!isRemoteWorkspace}` and the read-only plugin hint are still hard-coded from `isRemoteWorkspace`, not from the connected server capabilities.

The only code changed by this PR is `openworkServerCanWriteSkills` / `openworkServerCanWritePlugins` in `openwork-server-store.ts`, but those snapshot fields have no consumers in this branch. Because the UI paths above do not read those fields, the PR is effectively a no-op for the reported behavior.

Please wire the settings route to the connected server capability state for skills/plugins access hints and actions, and add a regression test or targeted coverage for remote workspace + `skills.write: true` / `plugins.write: true`.

Verification performed:
- `pnpm --filter @openwork/app test:health` passed.

PR #1538, opened 2026-04-23
fix(workspace): surface actionable remote worker diagnostics

### LINKED ISSUE #1419: [Bug]: Remote worker Test connection and Recover fail without actionable diagnostics
**Summary:** *
When a remote worker enters an error state, clicking `Test connection` or `Recover` does not restore connectivity and does not explain the blocking condition. The failure is reduced to generic or missing diagnostics, so users cannot tell whether the issue is the worker itself, the token, or the OpenWork server endpoint.

**Steps to reproduce:** *
1. Connect a remote worker in OpenWork.
2. Put that worker into a broken or unreachable state so it shows an error or offline state.
3. Open the worker actions menu and click `Test connection`.
4. Open the same menu and click `Recover`.
5. Check the worker status and any diagnostics shown in the session list or Settings.

**Expected behavior:** *
`Test connection` should validate the selected remote worker endpoint and return a specific result for that worker. `Recover` should restart or reconnect the worker and either bring it back online or show the exact blocking error. If diagnostics cannot be loaded, the UI should say why and from which endpoint the check failed.

**Actual behavior:** *
`Test connection` and `Recover` do not fix the remote worker state and do not surface a precise failure reason. The user is left with generic or missing diagnostics instead of an actionable message that identifies the real problem, such as a wrong endpoint, token failure, or worker availability issue.

---
Label: bug

pascalandr:
## Summary

- Fixes #1419 by restoring actionable remote worker diagnostics in the React app.
- Ports the old remote connection test/recover behavior into the current shell via `apps/app/src/react-app/domains/workspace/remote-connection.ts`.
- Wires connection state and test/recover handlers into `session-route.tsx` and `settings-route.tsx`, and shows the real error message in the workspace sidebar instead of a generic status.

## Testing

- `git diff --check upstream/dev...fix/1419-remote-worker-diagnostics` OK

## Notes

- This ended up being a larger React re-port than the original one-file Solid patch because the old workspace connection flow was no longer wired into the shell.


src-opn (approved):
it looks good

src-opn:
@pascalandr please fix conflicts

src-opn:
@pascalandr sorry didnt get to this in time, can you clear conflicts again plz

Issue #863, opened 2026-03-12
[Bug]: 输入prompt之后没有任何输出，但是在opencode中输入同样的内容就有输出，模型都一样


dwshmilyss:
### Summary

输入prompt之后没有任何输出，但是在opencode中输入同样的内容就有输出，模型都一样

### Steps to reproduce

<img width="2364" height="1626" alt="Image" src="https://github.com/user-attachments/assets/1aaa85fc-6a8d-4921-99d8-1b8de2e79eb8" />

<img width="2150" height="1028" alt="Image" src="https://github.com/user-attachments/assets/78840c7f-1433-4c1d-864f-150d000ae7ca" />

### Expected behavior

openwork有输出

### Actual behavior

没有任何输出，也不报错

liunux4odoo:
same problem on Windows with the latest release

liunux4odoo:
I think I found the problem: it doesn’t respond properly when run as a regular user, but it works fine when run with administrator privileges.

jcllobet:
@liunux4odoo @dwshmilyss can you let us know if this is still happening in the latest windows release? https://github.com/different-ai/openwork/releases/tag/v0.11.207

liunux4odoo:
> [@liunux4odoo](https://github.com/liunux4odoo) [@dwshmilyss](https://github.com/dwshmilyss) can you let us know if this is still happening in the latest windows release? https://github.com/different-ai/openwork/releases/tag/v0.11.207

still same problem, openwork must run as administrator, otherwise there is no response after sending the query

liunux4odoo:
the problem disappeared on version v0.11.212

PR #1541, opened 2026-04-23
fix(desktop): fall back to a writable sidecar cwd on Windows

pascalandr:
## Summary

- Mitigates #863 by adding a Windows-only fallback when the preferred working directory for desktop sidecars is not writable.
- Applies the fallback to the three native spawns that currently inherit the workspace/current directory directly: `opencode`, `openwork-server`, and `opencode-router`.
- Uses a writable directory under `app_local_data_dir()/runtime-cwd/<process>` when the preferred cwd probe fails.

## Testing

- `git diff --check upstream/dev...fix/863-sidecar-working-dir` OK
- `cargo check --manifest-path apps/desktop/src-tauri/Cargo.toml` OK (only preexisting warnings)

## Notes

- This is intentionally framed as a Windows retest candidate, not a guaranteed root-cause proof for every no-response report.
- The goal is to remove one plausible privilege-dependent failure mode: bundled sidecars inheriting a cwd they cannot write to unless the app is elevated.


src-opn (changes requested):
Review completed. Requesting changes.

Findings:
- `apps/desktop/src-tauri/src/engine/spawn.rs:95`: when the Windows fallback is used for `opencode`, `resolve_process_working_dir` returns `app_local_data_dir()/runtime-cwd/opencode`, and `opencode serve` is then started from that directory. `build_engine_args` does not pass `project_dir` as an argument, so this appears to start the OpenCode engine against the fallback runtime directory instead of the selected workspace. That breaks the core behavior exactly in the fallback path this PR is intended to fix. Please either preserve the selected workspace explicitly when spawning `opencode`, or avoid applying this cwd fallback to the engine unless the engine has another guaranteed way to target `project_dir`.
- `apps/desktop/src-tauri/src/paths.rs:2`, `apps/desktop/src-tauri/src/paths.rs:4`, `apps/desktop/src-tauri/src/paths.rs:6`, `apps/desktop/src-tauri/src/paths.rs:132`, `apps/desktop/src-tauri/src/paths.rs:140`: the new Windows-only helper code is compiled on non-Windows and produces new warnings on macOS (`unused import: tauri::Manager`, `writable_probe_name is never used`, `ensure_dir_writable is never used`). Please gate the imports/helpers with `#[cfg(windows)]` or otherwise structure this so non-Windows builds stay warning-clean.

Verification performed:
- Pulled PR into a dedicated local worktree.
- Checked the changed diff for whitespace errors: `git diff --check origin/dev...HEAD` passed.
- Scanned added diff lines for non-ASCII bytes, control bytes, zero-width/BOM bytes, and bidi isolate controls: passed.
- Installed dependencies with `pnpm install --frozen-lockfile`: passed, with pnpm warning that `sharp@0.34.5` build scripts were ignored.
- Ran `cargo check --manifest-path apps/desktop/src-tauri/Cargo.toml`: completed, but emitted the new warnings listed above.
- Ran `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml`: failed in `commands::orchestrator::tests::docker_command_falls_back_after_timeout` (`stdout.contains("Docker version 0.0.0")`). This test is outside the touched files, so I am not treating it as the primary PR blocker, but the failing verification should still be noted.

src-opn (approved):
looks good to me

PR #1543, opened 2026-04-23
server: harden MCP config validation

dharmateja03:
## Summary
- tighten validateMcpConfig for local MCP entries by requiring a non-empty command array of non-empty strings
- tighten remote MCP validation by rejecting blank URLs, malformed URLs, and non-http(s) schemes
- extend validator tests to cover localhost URLs and malformed/non-http URL and command cases

## Why
The server currently accepts malformed MCP configs that can later break MCP runtime behavior (for example, empty command argv entries or invalid remote URLs). This change fails fast at write-time and keeps workspace config clean.

## Testing
- pnpm install --filter openwork-server...
- pnpm --filter openwork-server test


src-opn (approved):
it looks good

PR #1548, opened 2026-04-23
feat(settings): add user env vars UI + shell injection

### LINKED ISSUE #1436: [Feature] Add a UI for managing environment variables
### Summary

Provide a settings screen where users can define and persist environment variables (API keys, endpoints, etc.) directly in the app, removing the need to edit shell config files or use gtk-launch workarounds.

### Problem / goal

Environment variables must currently be set in .bashrc or similar, and the app must be launched via gtk-launch from a terminal for those vars to be inherited. This is invisible friction that causes silent failures and is inaccessible to non-technical users.

### Primary user(s)

- [x] Bob (IT / power user)
- [x] Susan (non-technical)
- [ ] Other team roles

session.* (env vars feed into model provider config and tool auth), mcp (MCP server tokens/keys), skills/plugins (skill-level env overrides)

Making the app self-contained and configurable through its own UI, without terminal knowledge, directly supports the product's goal of being accessible to non-technical team members (Susan persona). It also prevents the common silent-failure scenario where the app starts without required keys.

### Testability

pnpm dev + open Settings > Environment Variables + add a key/value pair + restart session + verify the variable is available in a tool call or model request

Yes

### Additional context

Suggested behaviour:
- Key/value editor in Settings UI
- Values persisted to a local config file (e.g. ~/.config/openwork/env)
- Injected into the app process at startup
- Sensitive values masked (shown as ••••••) with a reveal toggle

This removes the need for gtk-launch or any shell-level workarounds for GUI users.

shizhigu:
## Summary
- Add Settings -> Environment for local API keys and tokens.
- Inject saved values into OpenWork-launched local processes: OpenCode, server, router, skills/plugins, MCPs, and sandbox runtimes.
- Send only configured key names to agent context so agents know what exists without seeing secret values.

## Why
GUI-launched desktop apps do not reliably inherit shell env. This gives users a predictable place to save service credentials for local agents, skills, MCP servers, and sidecars.

## Issue
Closes #1436

## Scope
- Local env store with restrictive writes, serialized mutations, and reserved-key filtering.
- Host-token-only env API routes.
- Tauri, Electron, and orchestrator env loading.
- Environment settings UI with masked values, reveal, delete confirmation, pending/apply state, and remote-workspace secret hiding.
- Key-name-only agent context. This is not written to `AGENTS.md`.
- Corrupt env stores return 409 instead of being overwritten.
- Applying env changes is blocked while local tasks are running.
- Pending apply state is scoped to the current local runtime, so a manual restart clears stale prompts.

## Out of scope
- OS keychain storage.
- Per-workspace env overrides.
- Cloud-managed secret distribution.

## Test plan
- `bun test apps/app/tests/env-context.test.ts apps/app/tests/openwork-env-runtime.test.ts`
- `pnpm --filter openwork-server exec bun test src/env-file.test.ts src/env-routes.e2e.test.ts`
- `pnpm --filter openwork-server build:bin`
- `pnpm --filter @openwork/desktop typecheck:electron`
- `pnpm --filter @openwork/desktop check:electron`
- `pnpm -C apps/desktop prepare:sidecar`
- `cargo test env_file`
- `git diff --check`

All passed locally on a branch containing current `origin/dev`.

## How I tested the app
1. Opened Settings -> Environment from the desktop app settings surface.
2. Added a local env key and confirmed it renders masked by default.
3. Revealed, edited, and deleted the key from the UI.
4. Applied changes and confirmed the local runtime reconnect flow.
5. Verified remote workspaces do not render local secrets.
6. Verified agents see configured key names only, not values.
7. Verified OpenCode appends per-message `system` context to the existing agent/provider prompt instead of replacing it.

Vercel fork previews require team authorization, so preview deploys may stay blocked until a maintainer approves them.

## Evidence
Settings -> Environment screenshot:

Environment variables dark mode

Demo recording of the local desktop flow: add key -> mask/reveal -> apply/reconnect -> agent sees the key name without seeing the value.

https://raw.githubusercontent.com/shizhigu/openwork/shizhigu/environment-variables-ui/apps/app/pr/environment-variables-demo.mp4

Design note: `apps/app/pr/environment-variables.md`

## Risk
Medium surface area because this touches runtime launch paths. The change is additive, host-token gated, and strips reserved OpenWork/OpenCode keys before injection.

## Rollback
Revert this PR. If `env.json` is absent, existing behavior is unchanged.


shizhigu:
Two follow-up commits on the same branch:

- `49a47c28` — fix: `/settings/environment` was redirecting to `/settings/general`. `parseSettingsPath` had a hardcoded whitelist that didn't include the new tab. One-line fix, case added.
- `b53a4fdf` — docs: reframed UI copy + PRD to lead with user value (credentials for the services skills and MCPs call into — ElevenLabs, Gemini image APIs, GitHub, LangSmith, proxies, etc.) instead of the original "env vars so the app reads them without a terminal" plumbing framing. en/zh/ja updated; other locales fall back to en.

Screenshots to follow in one more commit.

src-opn (approved):
it looks good

PR #1551, opened 2026-04-24
fix(session): poll sidebar session list to surface externally created sessions

### LINKED ISSUE #1262: [Bug]: Automated Runs not visible until Session renamed
### Summary

Session with automated runs generates a new session for each automated run, but these don't appear until you rename the original session. 

### To Reproduce

1. In a session generate a script to do a basic task on a recurring basis.
2. Ask openwork to automate this with an automation/cron
3. The automation is saved
4. After it has ran next, rename the original session and you will see a newly generated session below it that was not previously viewable.

### Expected behavior

The automated run should just generate the new session in the sidebar regardless of renaming.

I should probably also put this in feature request but I am going to do so here, tell me if you would prefer it separate in the future:
- Automated Run sessions should not just populate under the same workspace. If anything they should populate under the original session thread. But realistically there should be a proper place for these, maybe in the automations area, but I think just having it below the original session in a toggleable drop down would be the best UX.

### Actual behavior

Automated run sessions are not visible until original session renamed.

### Screenshots (optional)

_No response_

Version 0.11.195 (0.11.195)
Mac

Lmk if need any more clarification on this, also apologies if this is patched in a new version I am doing a security sweep before updating any apps atm.

Formatted:
## Summary
- Add a 30s polling interval that refreshes the sidebar session list when the tab is visible
- The existing `refreshInFlightRef` guard prevents overlapping calls, so this does not interfere with user-triggered refreshes
- Only fires when `document.visibilityState === "visible"` to avoid unnecessary network requests

## Problem
Automated runs, messaging bot sessions (Slack/Telegram), and other externally created sessions were invisible in the sidebar until the user triggered an indirect refresh (e.g. renaming a session). This is because `refreshRouteState()` was only called on explicit user actions and visibility change events, with no periodic refresh mechanism.

## Solution
Added a lightweight `useEffect` in `session-route.tsx` that calls `refreshRouteState()` every 30 seconds when the tab is visible. The existing dedup guard in `refreshRouteState` ensures overlapping calls are skipped, so this integrates safely with all existing refresh paths.

## Files Changed
- `apps/app/src/react-app/shell/session-route.tsx` — added periodic polling `useEffect`

## Issue
Fixes #1262

### DIFF
--- a/apps/app/src/react-app/shell/session-route.tsx
+++ b/apps/app/src/react-app/shell/session-route.tsx
@@ -529,6 +529,22 @@ export function SessionRoute() {
     };
   }, [refreshRouteState]);
 
+  // Periodic session list refresh so that sessions created externally (e.g. by
+  // automations, messaging bots, or other clients) appear in the sidebar without
+  // requiring the user to rename a session or manually refresh the page. Only
+  // fires when the tab is visible to avoid unnecessary network requests. The
+  // existing refreshInFlightRef guard prevents overlapping calls.
+  // Fixes #1262
+  useEffect(() => {
+    if (typeof window === "undefined") return;
+    const interval = setInterval(() => {
+      if (typeof document !== "undefined" && document.visibilityState !== "visible") return;
+      refreshInFlightRef.current = false;
+      void refreshRouteState();
+    }, 30000);
+    return () => clearInterval(interval);
+  }, [refreshRouteState]);
+
   useEffect(() => {
     if (!isDesktopRuntime()) return;
     let cancelled = false;


src-opn:
we're moving more to query caching, and having an interval in a useEffect is generally not great practice
closing for now, to be implemented with query caching system

Formatted:
Hi @src-opn — thanks for the feedback. I've opened a replacement PR #1584 that addresses the query caching concern:

- Session lists are now written to the TanStack Query cache
- Polling is extracted to a reusable hook with a query key pattern
- Uses a lightweight session-only refresh instead of the heavy refreshRouteState
- Still respects the existing refreshInFlightRef dedup guard

Let me know if this approach works better!

PR #1552, opened 2026-04-24
fix(desktop): disable WebKit compositing on Linux to prevent NVIDIA/X…

### LINKED ISSUE #1238: bug: Crash on startup with NVIDIA GPU on X11: `Failed to create GBM buffer`
## Summary

OpenWork crashes immediately on startup on systems with an NVIDIA GPU under X11. The crash is caused by WebKitGTK's hardware compositing, which fails to create a GBM buffer. The Tauri app does not set `WEBKIT_DISABLE_COMPOSITING_MODE=1` by default.

## To Reproduce

1. System with NVIDIA GPU + proprietary driver + X11 (not Wayland)
2. Launch OpenWork
3. Immediate crash with error:
   ```
   Failed to create GBM buffer of size 2297x1368: Invalid argument
   ```

## Expected behavior

OpenWork should either:
- Detect NVIDIA + X11 and disable compositing automatically
- Set `WEBKIT_DISABLE_COMPOSITING_MODE=1` as a safe default for Linux builds
- Gracefully fall back to software rendering instead of crashing

## Actual behavior

The application crashes immediately with a GBM buffer allocation error. No fallback rendering is attempted.

## OW version & Desktop info

- **OS:** Arch Linux x86_64
- **GPU:** NVIDIA (proprietary driver)
- **Display server:** X11 / KDE KWin

## Additional context

Workaround:

```bash
export WEBKIT_DISABLE_COMPOSITING_MODE=1
openwork
```

This is a known issue with WebKitGTK + NVIDIA + X11 that affects many Tauri/WebKitGTK apps. Setting the env var has no negative side effects — it simply uses software compositing.

notHim0:
## Summary
- Set `WEBKIT_DISABLE_COMPOSITING_MODE=1` automatically on Linux 
  builds before the Tauri window initializes

## Why
- WebKitGTK attempts to use GPU compositing via GBM on startup
- On systems with NVIDIA proprietary drivers + X11, this fails with
  "Failed to create GBM buffer: Invalid argument" and crashes the app
- Setting this env var tells WebKitGTK to use software compositing
  instead, which works reliably across all Linux GPU/display configs

## Issue
- Closes #1238 

## Scope
- Single line change in `apps/desktop/src-tauri/src/main.rs`
- Linux builds only (guarded by `#[cfg(target_os = "linux")]`)
- No effect on Windows or macOS builds

## Out of scope
- Detecting specific GPU vendors or display servers at runtime
- Any changes to the frontend or WebKit configuration beyond this var

### Ran
- `cargo check --manifest-path apps/desktop/src-tauri/Cargo.toml`
- `pnpm dev`

### Result
- pass/fail: pass

## CI status
- pass: yes (compilation succeeds, no new warnings introduced)
- code-related failures: none

## Manual verification
1. On a Linux system with NVIDIA proprietary driver + X11, launch 
   OpenWork without the fix → app crashes with GBM buffer error
2. Apply this fix and relaunch → app starts successfully
3. On Wayland or non-NVIDIA systems, app continues to work normally

Note: I was unable to personally reproduce the crash due to my laptop
running on integrated AMD graphics. The fix is consistent with the
workaround documented in the project README and is the standard
approach used across the Tauri ecosystem for this known WebKitGTK issue.

## Evidence
- N/A (could not reproduce crash on my hardware setup, see note above)

## Risk
- Very low. The env var only affects Linux builds and disables GPU
  compositing in WebKitGTK — there is no user-visible difference in
  rendering quality for a desktop app of this type

## Rollback
- Revert the single line addition in `apps/desktop/src-tauri/src/main.rs`

### DIFF
--- a/apps/desktop/src-tauri/src/main.rs
+++ b/apps/desktop/src-tauri/src/main.rs
@@ -1,5 +1,8 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
 fn main() {
+    #[cfg(target_os = "linux")]
+    std::env::set_var("WEBKIT_DISABLE_COMPOSITING_MODE", "1");
+
     openwork::run();
 }


src-opn:
We're moving away from tauri to electron
not sure if this will still be an issue
leaving it open for reference

src-opn:
We are moving our desktop builds to electron and thus these changes will not apply.
Closing this PR out, but if the issue still exists with electron, please update PR or open a new one

AasheeshLikePanner:
@src-opn why did you guys move from tauri to electron? i keep seeing this happen. a lot of products start with tauri and then later switch to electron. what was the reason?

PR #1554, opened 2026-04-24
fix(workspace): diagnose remote worker failures

shizhigu:
## Why
Remote worker failures could look like a generic loading state. The sidebar also exposed Edit connection, but that action was not wired, so users could not fix a stale URL or token from the failing workspace row.

## What changed
- Run OpenWork worker diagnostics for Test, Recover, and task-load failures.
- Redact token-like values before showing diagnostic errors.
- Stop OpenWork probes for non-OpenWork remotes, even if stale OpenWork fields exist.
- Wire Edit connection through the existing remote workspace modal and ignore stale async diagnostic results after connection changes.

## Test plan
- `pnpm --filter @openwork/app test:remote-diagnostics`
- `git diff --check`

## Evidence
Before, Test connection could leave the workspace in a generic loading state.

Before

After, the sidebar shows the failing check and recovery actions.

After

## Notes
This does not restart or repair remote infrastructure. It only diagnoses the failing connection and lets the user edit the saved connection details.


src-opn (approved):
it looks good

PR #1565, opened 2026-04-26
feat: add workspace import previews

shizhigu:
## Why
Workspace/template imports can change `.opencode` config, skills, commands, and portable files. Without a preview, users are approving a write they cannot inspect, and replace mode can remove existing setup with no server-side proof that the reviewed state still matches the workspace.

This adds the server primitive for safe imports: preview first, then apply only if the workspace still matches that preview.

## What changed
- Add `POST /workspace/:id/import/preview` with a public diff and preview fingerprint.
- Require changed `POST /workspace/:id/import` calls to include a matching `previewFingerprint`; stale previews return 409 with the latest preview.
- Make replace mode honor empty `skills`, `commands`, and `files` arrays, so "replace with none" removes existing items.
- Share skill/command content builders so preview and write use the same formatting rules.

## Test plan
- `pnpm --filter openwork-server exec bun test src/workspace-import-preview.test.ts`
- `pnpm --filter openwork-server build:bin`
- `git diff --check`

## Notes
No app UI in this revision. The older team-template UI this branch originally targeted was removed from `dev` in #1593, so this keeps the server/API import safety primitive without bringing that surface back.

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -202,6 +202,25 @@ export type OpenworkWorkspaceExport = {
   files?: Array<{ path: string; content: string }>;
 };
 
+export type OpenworkWorkspaceImportChange = {
+  kind: "opencode" | "openwork" | "skill" | "command" | "file";
+  action: "create" | "update" | "replace" | "delete" | "unchanged";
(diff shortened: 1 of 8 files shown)


src-opn (approved):
it looks good

PR #1567, opened 2026-04-26
feat(orchestrator): forward --opencode-log-level to managed opencode serve

### LINKED ISSUE #1495: Feature request: add --opencode-log-level passthrough for managed opencode serve
## Feature Request

OpenWork orchestrator spawns `opencode serve` without any log-level argument:

```
/Applications/OpenWork.app/Contents/MacOS/opencode serve --hostname 127.0.0.1 --port 65491 --cors *
```

This causes opencode to log at INFO level by default, generating **hundreds of megabytes** of log files in `~/.local/share/opencode/log/`. A simple "hello" task creates 34 MB of logs; a running `opencode serve` can generate 300+ MB in minutes.

## Requested Change

Add an `--opencode-log-level` option to the orchestrator CLI (alongside existing `--opencode-host`, `--opencode-port`, etc.) that gets passed through to the managed `opencode serve` process as `--log-level`.

For example:
```
openwork start --opencode-log-level ERROR
```

Would result in spawning:
```
opencode serve --hostname 127.0.0.1 --port <port> --cors * --log-level ERROR
```

## Context

- OpenWork already has many `--opencode-*` passthrough options (host, port, auth, hot-reload, etc.), so this fits the existing pattern.
- The opencode config file has a `logLevel` key in its schema, but it is currently not honored at runtime (reported separately as [anomalyco/opencode#23448](https://github.com/anomalyco/opencode/issues/23448)). Even if that gets fixed, having a CLI passthrough is a good safety net.
- Users who set shell aliases for `opencode --log-level ERROR` are unaffected in their terminal, but the OpenWork-managed process bypasses aliases entirely.

## Environment

- OpenWork orchestrator version: 0.11.207
- opencode version (bundled): 1.2.27
- OS: macOS (darwin arm64)

yashranaway:
## Summary

- Adds `--opencode-log-level <DEBUG|INFO|WARN|ERROR>` (and `OPENWORK_OPENCODE_LOG_LEVEL` env equivalent) to `openwork start`, `openwork serve`, and `openwork daemon start`.
- When set, the orchestrator forwards `--log-level <value>` to the managed `opencode serve` invocation. When unset, opencode keeps its default behavior — no change for existing users.
- Updates CLI help and README so the flag is discoverable.

## Why

`opencode` defaults to `INFO`, which produces large log files in `~/.local/share/opencode/log/` (the issue reporter measured ~34 MB for a single "hello" task and 300+ MB after a few minutes of `opencode serve`). The orchestrator already passes `--hostname`, `--port`, and `--cors` through to `opencode serve`, but never `--log-level`, so users have no orchestrator-level knob to lower verbosity.

## Issue

- Closes #1495

## Scope

- `apps/orchestrator/src/cli.ts`
  - New `resolveOpencodeLogLevel()` validator next to `resolveManagedOpencodeHost()`. Same throw-on-invalid pattern; normalizes case; rejects values outside `DEBUG | INFO | WARN | ERROR`.
  - `startOpencode()` gains an optional `logLevel?: string` and pushes `--log-level <value>` only when set.
  - Flag is read in all three relevant entry points (`spawnRouterDaemon`, `runRouterDaemon`, `runStart`) using the existing `readFlag(...) ?? process.env.OPENWORK_OPENCODE_LOG_LEVEL` pattern, and threaded into all three `startOpencode()` call sites.
  - `spawnRouterDaemon` also forwards `--opencode-log-level` to the child orchestrator subprocess via `commandArgs.push(...)` so daemon-spawned children inherit the value.
  - Help text gains one line under `--opencode-workdir`.
- `apps/orchestrator/README.md`
  - Short paragraph in the **Logging** section explaining the flag, env var, and the reason (large default log files).

## Out of scope

- The sandbox-mode `opencode` invocation built as a bash-string template at `cli.ts:4266` is intentionally not changed in this PR. It runs inside the container and has different shell-quoting concerns; the issue reporter's pain is host-side. Happy to do that as a follow-up.

### Ran

- `bun src/cli.ts --help` (verify help text placement)
- `bun src/cli.ts start --workspace /tmp --opencode-log-level BANANA` (invalid value)
- `bun src/cli.ts start --workspace /tmp --opencode-log-level warn ...` (lowercase normalization)
- `bun src/cli.ts daemon start --workspace /tmp --opencode-log-level FOO` (daemon path validation)

### Result

- pass: typecheck clean
- pass: `--help` lists the new flag in the expected position alongside the other `--opencode-*` flags
- pass: invalid value rejected with `Unsupported --opencode-log-level value: BANANA. Expected one of: DEBUG, INFO, WARN, ERROR.`
- pass: lowercase `warn` accepted and normalized to `WARN`
- pass: same validator path enforced on `daemon start`

## CI status

- pass: pending CI on the PR

## Manual verification

1. `openwork start --workspace /path/to/workspace --opencode-log-level WARN` — confirm the spawned `opencode serve` command line includes `--log-level WARN`.
2. `OPENWORK_OPENCODE_LOG_LEVEL=ERROR openwork start --workspace /path/to/workspace` — confirm env var fallback works when no flag is passed.
3. Run a small session; observe that log volume in `~/.local/share/opencode/log/` is materially lower than the default `INFO`.
4. Omit the flag entirely; confirm no behavior change vs. `dev` baseline (no `--log-level` is forwarded, opencode keeps its default).

## Evidence

- N/A (CLI text-only change). Help-output and rejection messages are reproducible with the commands above.

## Risk

- Low. Optional flag, default behavior unchanged when unset. Validator throws early on bad input rather than passing it to `opencode serve`.

## Rollback

- Revert this commit. No persisted state, no schema changes, no config migrations.

### DIFF
--- a/apps/orchestrator/README.md
+++ b/apps/orchestrator/README.md
@@ -124,6 +124,11 @@ OPENWORK_LOG_FORMAT=json openwork start --workspace /path/to/workspace
 
 Use `--run-id` or `OPENWORK_RUN_ID` to supply your own correlation id.
 
+OpenCode runs at `INFO` by default, which produces large log files in
+`~/.local/share/opencode/log/`. Pass `--opencode-log-level <DEBUG|INFO|WARN|ERROR>` (or set
+`OPENWORK_OPENCODE_LOG_LEVEL`) to forward `--log-level` to managed `opencode serve` and reduce log
+volume.
+
 OpenWork server logs every request with method, path, status, and duration. Disable this when running
 `openwork-server` directly by setting `OPENWORK_LOG_REQUESTS=0` or passing `--no-log-requests`.
 
--- a/apps/orchestrator/src/cli.ts
+++ b/apps/orchestrator/src/cli.ts
@@ -4185,6 +4185,7 @@ async function writeSandboxEntrypoint(options: {
     username?: string;
     password?: string;
     hotReload: OpencodeHotReload;
+    logLevel?: string;
   };
   openwork: {
     token: string;
@@ -4213,6 +4214,10 @@ async function writeSandboxEntrypoint(options: {
   const opencodeCors = options.opencode.corsOrigins
     .map((origin) => `--cors ${shQuote(origin)}`)
     .join(" ");
+
+  const opencodeLogLevelArg = options.opencode.logLevel
+    ? `--log-level ${shQuote(options.opencode.logLevel)}`
+    : "";
 
   const openworkCors = options.openwork.corsOrigins.length
     ? `--cors ${shQuote(options.openwork.corsOrigins.join(","))}`
@@ -4282,7 +4287,7 @@ async function writeSandboxEntrypoint(options: {
     '  if [ -n "$opencode_pid" ]; then kill "$opencode_pid" 2>/dev/null || true; fi',
     "}",
     "trap cleanup INT TERM",
-    `${shQuote(opencodeBin)} serve --hostname 127.0.0.1 --port ${shQuote(String(SANDBOX_INTERNAL_OPENCODE_PORT))} ${opencodeCors} &`,
+    `${shQuote(opencodeBin)} serve --hostname 127.0.0.1 --port ${shQuote(String(SANDBOX_INTERNAL_OPENCODE_PORT))}${opencodeLogLevelArg ? ` ${opencodeLogLevelArg}` : ""} ${opencodeCors} &`,
     "opencode_pid=$!",
     options.openwork.opencodeRouterEnabled
       ? `${shQuote(opencodeRouterBin)} serve ${shQuote(workspaceDir)} &`
@@ -4326,6 +4331,7 @@ async function startDockerSandbox(options: {
     username?: string;
     password?: string;
     hotReload: OpencodeHotReload;
+    logLevel?: string;
   };
   openwork: {
     token: string;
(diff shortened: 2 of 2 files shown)


src-opn (changes requested):
This needs changes before merge.

Issue:
- `apps/orchestrator/src/cli.ts:7938-7997` wires sandbox `openwork start` without passing the resolved `opencodeLogLevel` into the sandbox options, and `apps/orchestrator/src/cli.ts:4246-4285` builds the sandbox entrypoint without appending `--log-level <level>` to the sandboxed `opencode serve` command. As a result, `openwork start --sandbox ... --opencode-log-level WARN` accepts and validates the new flag but silently ignores it, while the PR advertises the flag for `openwork start` generally. Please either thread the resolved log level into the sandbox launch path and shell-quote it into the entrypoint, or reject the flag when sandbox mode is enabled instead of silently ignoring the user's explicit setting.

Verification run:
- `pnpm --filter openwork-orchestrator build`
- `bun src/cli.ts --help`
- `bun src/cli.ts start --workspace /tmp --opencode-log-level BANANA --check`
- `bun src/cli.ts start --workspace /tmp --opencode-log-level warn --check`

yashranaway:
@src-opn  Threaded the resolved `opencodeLogLevel` through both sandbox launchers and shell-quoted `--log-level `into the `opencode serve` line in the entrypoint. Verified with the same `--check` runs you listed; valid value passes, invalid still rejected upfront.

src-opn (approved):
looks good to me

PR #1576, opened 2026-04-27
feat: add MiniMax-M2.7 provider support

octo-patch:
## Summary

- Pin minimax in the provider sort order (PINNED_PROVIDER_ORDER) so it appears featured in the Connections UI alongside opencode, openai, and anthropic
- Add MM abbreviation in ProviderIcon for the minimax provider ID (the generic substring fallback produced MI instead of MM)
- Update eval examples from opencode/minimax-m2.5-free to opencode/minimax-m2.7

## Test plan

- [ ] Open Connections tab: MiniMax appears before unlisted providers
- [ ] Connected MiniMax provider shows MM badge in provider icon
- [ ] Flows 2 and 6 in evals/react-session-flows.md work with opencode/minimax-m2.7

### DIFF
--- a/apps/app/src/app/utils/providers.ts
+++ b/apps/app/src/app/utils/providers.ts
@@ -1,6 +1,6 @@
 import type { Provider as ConfigProvider, ProviderListResponse } from "@opencode-ai/sdk/v2/client";
 
-const PINNED_PROVIDER_ORDER = ["opencode", "openai", "anthropic"] as const;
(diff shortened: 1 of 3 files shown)


src-opn:
Hey, thanks for the PR. Unfortunately we dont prioritize uncommon providers. OpenRouter would come way ahead of minimax.
Thanks for the effort though

PR #1582, opened 2026-04-27
Fix release sidecar workflow job

src-opn:
## Summary
- Restores the missing `release-orchestrator-sidecars` job key in the macOS release workflow.
- Fixes GitHub Actions duplicate-key parsing errors for `name`, `needs`, `if`, `runs-on`, `env`, and `steps`.

## Verification
- `ruby -e 'require \"yaml\"; YAML.load_file(\".github/workflows/release-macos-aarch64.yml\"); puts \"yaml ok\"'`

## Notes
- `actionlint` was not installed locally, so only YAML parsing was run.


src-opn:
Closing as redundant: origin/dev already contains the release workflow job fix.

PR #1585, opened 2026-04-28
fix release latest updater feed

shizhigu:
## Why
Users see this as `Settings > Updates` failing with `Could not fetch a valid release JSON from the remote`. The app is not the broken part in that case. GitHub Latest is pointing at a release that is not a complete desktop updater release yet.

There are two ways this can happen today: a sidecar-only release can become Latest with no `latest.json`, and an Electron macOS release can publish a `latest-mac.yml` from the last matrix job only. The current live `v0.13.0/latest-mac.yml` only lists the x64 zip, so arm64 Macs can be offered the x64 app.

## What changed
- Keep sidecar-only and unfinished desktop releases out of GitHub Latest
- Check Tauri `latest.json` by tag before the release becomes Latest
- Merge Electron `latest*.yml` files from matrix builds before upload
- Reject macOS updater metadata unless it contains both arm64 and x64 zip assets

## Test plan
- `node --check scripts/release/verify-updater-endpoint.mjs`
- `node --check scripts/release/consolidate-electron-updater-yml.mjs`
- `node --test scripts/release/verify-updater-endpoint.test.mjs`
- `ruby -e 'require "yaml"; %w[.github/workflows/release-macos-aarch64.yml .github/workflows/alpha-macos-aarch64.yml .github/workflows/ci-tests.yml].each { |f| YAML.load_file(f); puts "yaml ok: #{f}" }'`
- `git diff --check`
- Expected-failure check against the current live feed: `latest-mac.yml` is missing `openwork-mac-arm64-0.13.0.zip`

## Notes
This does not mutate existing GitHub releases. The already-published `v0.13.0/latest-mac.yml` needs a release rerun or manual manifest repair. This PR stops the same class of bad feed from becoming Latest again.


src-opn:
this should be solved now

PR #1587, opened 2026-04-28
fix: the user-message copy button hover behavior

### LINKED ISSUE #1586: Copy button overlaps message text in session chat bubbles
In the session view, the copy-message action is positioned too close to the message bubble content and visually overlaps/interferes with the message text. This makes the message harder to read and makes the UI feel cramped.

Image

Environment : 
OpenWork desktop app on Windows

Steps to Reproduce : 
1. Open a session in the OpenWork desktop app.
2. Send a short user message.
3. Hover or focus the message so the copy button appears.
4. Look at the right edge of the message bubble.

veeradyani222:
https://github.com/user-attachments/assets/7681df81-74f5-4e7d-9aa9-0729bca78df6

Fixed the user-message copy button hover behavior.

The copy button used to sit inside the message bubble, so on longer user messages it could overlap the text and make the bubble look messy. I moved the copy action outside the bubble, positioned at the bottom-right on hover, so the message content keeps its full readable space.

I also added an invisible hover area across the existing gap between the user message and the assistant reply. That keeps the copy button reachable without adding extra vertical spacing or shifting the transcript layout.

/closes #1586

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/message-list.tsx
+++ b/apps/app/src/react-app/domains/session/surface/message-list.tsx
@@ -949,10 +949,10 @@ function SessionTranscriptInner(props: SessionTranscriptProps) {
     return (
       <div
         key={`message-${block.messageId}`}
-        className={`flex group ${block.isUser ? "justify-end" : "justify-start"}`.trim()}
+        className={`relative flex group ${block.isUser ? "justify-end" : "justify-start"}`.trim()}
         data-message-role={block.isUser ? "user" : "assistant"}
         data-message-id={block.messageId}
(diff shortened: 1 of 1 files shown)


veeradyani222:
Hi Ben! This is my first contribution to Openwork. Hope you approve it. Lmk for any changes!
@benjaminshafii

veeradyani222:
@src-opn have a look!

src-opn:
No need to tag people
the implementation is in the right direction, but now causes UI confusion since theres no separation between the copy button and the message below
have another go

PR #1594, opened 2026-04-28
fix session resume render loop

shizhigu:
## Why
I hit this in the desktop app after leaving a session idle: the transcript could go blank and the renderer would spin, while the OpenWork server and OpenCode stayed healthy. The failure path is in the React session surface, not the runtime.

## What changed
- Keep session cache entries available across surface unmounts instead of clearing transcript/status/todo state immediately.
- Fall back to the server snapshot when the live transcript cache is empty.
- Use stable empty/idle fallbacks for `useSyncExternalStore`, so the session surface does not see a new snapshot every render.
- Reconnect the OpenCode event stream after it ends, and defer message copy serialization until the user clicks copy.

## Test plan
- `bun test apps/app/scripts/session-render-state.test.ts`
- `pnpm --filter @openwork/app test:session-error-recovery`
- `git diff --check`

I also ran a 10-minute idle smoke against the same release server/session that reproduced the issue. Before this patch, the renderer hit ~110% CPU and close to 1GB RSS around the 5-minute mark. With this branch, it stayed around 0-0.3% CPU and ~103-121MB RSS, with the title, transcript text, and composer still present.

## Notes
No screenshot attached here. A static screenshot does not really prove an idle render loop is fixed; the smoke metrics above are the useful evidence. This is scoped to the session resume/cache path and does not claim to fix every possible app freeze.

### DIFF
--- a/apps/app/scripts/session-error-recovery.ts
+++ b/apps/app/scripts/session-error-recovery.ts
@@ -1,6 +1,9 @@
 import assert from "node:assert/strict";
 
-import { latestSessionErrorTurnTime, shouldResetRunState } from "../src/app/session/run-state";
+import {
+  latestSessionErrorTurnTime,
+  shouldResetRunState,
+} from "../src/react-app/domains/session/sync/run-state";
 
 assert.equal(latestSessionErrorTurnTime([]), null);
 assert.equal(
--- a/apps/app/scripts/session-render-state.test.ts
+++ b/apps/app/scripts/session-render-state.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "bun:test";
+import type { UIMessage } from "ai";
+
+import type { OpenworkSessionSnapshot } from "../src/app/lib/openwork-server";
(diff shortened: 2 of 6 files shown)


benjaminshafii:
Thank you for jumping on this perf issue.

I am going to add the missing safeguards directly on top of this PR: gate the stale snapshot fallback so session A cannot render under session B during uncached switches, add a regression test for that switch case, sanity-check retained session cache behavior, and make sure the branch is updated/rebased before it is ready to merge.

shizhigu:
> Thank you for jumping on this perf issue.
> 
> I am going to add the missing safeguards directly on top of this PR: gate the stale snapshot fallback so session A cannot render under session B during uncached switches, add a regression test for that switch case, sanity-check retained session cache behavior, and make sure the branch is updated/rebased before it is ready to merge.

@benjaminshafii  Thanks for adding those safeguards and getting this merged. I read through the follow-up changes. The sessionId guard and regression test make sense. I’ll keep that in mind for future PRs.

PR #1613, opened 2026-04-29
fix(settings): gate updater controls by capability

shizhigu:
## Why

Settings -> Updates can run in three different contexts: web, Electron dev, and packaged desktop. Only the packaged desktop app can actually check and install updates, but the settings page still showed active updater controls in unsupported contexts. That made the desktop dev app look broken with errors like `Electron update checks are available only in the Electron desktop app.`

## What changed

- Have the Electron bridge report whether update checks are supported.
- Gate the Updates controls until support is known.
- Hide release channel, background checks, auto-update, and manual check when the bridge or packaged updater support is unavailable.
- Keep packaged desktop behavior and release-channel persistence unchanged.

## Evidence

Before, unsupported desktop/runtime states still showed updater controls and failed after clicking Check:

Before

After, unsupported Electron dev builds show the passive state directly and hide controls that cannot work there:

After

## Test plan

- `bun test apps/app/tests/updater-capability.test.ts`
- `node --check apps/desktop/electron/updater.mjs && node --check apps/desktop/electron/preload.mjs`
- `pnpm --dir apps/desktop check:electron`
- `pnpm --dir apps/desktop typecheck:electron`
- `pnpm --filter @openwork/desktop package:electron:dir`
- `git diff --check origin/dev...HEAD`

## Notes

This does not change release manifests, signing, notarization, or the GitHub update feed. Recent release-workflow commits on `dev` handle the feed side; this PR only fixes the Settings capability model.

Known limits:

- Electron dev remains unsupported for update checks because `electron-updater` only loads when `app.isPackaged` is true.
- A packaged app can still fail later if a future release publishes a bad `latest.json`. That belongs in the release workflow verifier, not this settings view.
- The local package-dir check verifies the app can package, but not a signed public update install. That requires a tagged release with published assets.


src-opn:
@benjaminshafii can you check if this is solved in your recent commits

PR #1632, opened 2026-05-01
Feat/published workflows mcp

PetaHarshith:
Adds a way to expose an OpenWork skill as an MCP-callable tool.

You can publish a skill from the UI, get a URL, and call it from Claude/Cursor via MCP. Under the hood it creates an OpenCode session and sends a prompt that asks the agent to use that skill.

What changed:
- server routes for publish / list / revoke
- MCP endpoint at `/published/:token/mcp`
- UI for publish, copy URL, and revoke
- token is shown once and stored hashed
- tests for storage + MCP flow

Tested end-to-end against real OpenCode. Curl transcript is in `apps/app/pr/published-workflows-mcp.curl.txt`.

Didn’t include multi-tool or streaming. Wanted to keep v1 small.

Let me know what you think


src-opn:
1. why

PetaHarshith:
@src-opn 

Right now workflows are only usable inside OpenWork. This feature makes them callable from outside, e.g. Claude, Cursor, or any MCP client can trigger a workflow via a URL.

So instead of just running workflows locally, you can integrate them into other tools and flows.
It felt like a natural step towards making OpenWork workflows composable.

PR #1650, opened 2026-05-05
I18n: Prune unused translations

johnnyshields:
Run with `script/i18n-audit.mjs --prune`. 593 unused keys removed.


src-opn (approved):
looks good to me

Issue #1095, opened 2026-03-21
Make updates active by default with gentle upgrade flow


benjaminshafii:
## Context

\nWe should make updates active by default so users upgrade smoothly without feeling forced. The experience should be visible enough to drive upgrades, but not aggressive.\n\n## Goals\n- Updates are enabled by default for new installs (and preferably existing users where safe).\n- Upgrade nudges are noticeable and clear, without being pushy.\n- Provide a path to force update when a release is required (implementation choice left to owner).\n\n## Requirements\n- Define the default update setting and migration behavior for existing users.\n- UX surfaces should be visible (banner, callout, or similar) with clear intent.\n- Include a safe escalation path for required updates (e.g., backend-gated or app-enforced) without blanket aggressiveness.\n\n## Notes\nImplementation details are up to the implementer; prioritize a smooth, non-aggressive path that still drives upgrades.

benjaminshafii:
Updating the scope here: render actual new lines instead of literal `\n` text, and explore a pill-like updates presentation using the Codex app as inspiration.

PR #1656, opened 2026-05-05
fix(app): normalize escaped newlines in release notes

yashranaway:
Refs #1095 (does not close it; only addresses the rescoped sub-comment about escaped newlines).

## Summary
- Convert literal `\n` and `\r\n` escape sequences to real newlines in release notes before they reach the Settings updates view.
- Apply the same normalization to both the string and array shapes returned by `electron-updater`, so notes built from `releaseNotes` arrays render with intended line breaks.
- Keep the existing `whitespace-pre-wrap` rendering path unchanged; this only fixes the source string when the upstream payload is double-escaped.

## Verification
- `git diff --check` clean.
- `pnpm --filter @openwork/app typecheck` currently fails on unrelated `origin/dev` TypeScript errors in `messaging-view-state.ts`, `command-palette.tsx`, and `settings-route.tsx`; this branch does not touch those files.

## Notes
- The broader "updates active by default" piece from the original issue body is split out into #1663 (default flip plus explicit-opt-out-respecting migration).
- Still out of scope: the pill-like updates presentation and the required-update / backend-gated escalation path described on the issue.

### DIFF
--- a/apps/app/src/react-app/domains/settings/state/electron-updater-state.ts
+++ b/apps/app/src/react-app/domains/settings/state/electron-updater-state.ts
@@ -65,14 +65,18 @@ function describeError(error: unknown) {
   return serialized && serialized !== "{}" ? serialized : String(error);
 }
 
+function normalizeReleaseNotes(value: string): string {
+  return value.replace(/\\r\\n|\\n/g, "\n");
+}
+
 function releaseNotesToText(value: unknown): string | undefined {
-  if (typeof value === "string") return value;
+  if (typeof value === "string") return normalizeReleaseNotes(value);
   if (Array.isArray(value)) {
     return value
       .flatMap((entry) => {
-        if (typeof entry === "string") return entry;
+        if (typeof entry === "string") return normalizeReleaseNotes(entry);
         if (entry && typeof entry === "object" && "note" in entry) {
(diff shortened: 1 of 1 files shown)


src-opn (changes requested):
Code verification issue:

The PR description says `Closes #1095`, but #1095 asks to make updates active by default with a gentle upgrade flow, including default/migration behavior and escalation paths. This patch only normalizes escaped release-note newlines in `apps/app/src/react-app/domains/settings/state/electron-updater-state.ts`.

Please remove or change the closing reference before merge, otherwise GitHub will close #1095 even though the requested update-default flow remains unimplemented.

Verification run:
- `pnpm --filter @openwork/app typecheck` failed on existing unrelated TypeScript errors outside this patch.

yashranaway:
@src-opn updated the body to `Refs #1095`. The broader "updates active by default" piece from the issue body is up in #1663 (one-line default flip that respects explicit opt-out via the existing `readStoredBoolean` semantics). The pill-style presentation and required-update path stay open for a follow-up.

yashranaway:
@src-opn rebased onto current `dev` so the diff is clean against the recent updater-state refactor. The body has been `Refs #1095` (not `Closes`) since the original review — happy to dismiss the `CHANGES_REQUESTED` whenever you have a minute.

PR #1682, opened 2026-05-06
fix(composer): allow all attachment file types

src-opn:
## Summary

- Removes the composer attachment MIME allowlist so users can attach any file type.
- Keeps existing image compression for image attachments while passing all other files through unchanged.
- Updates attachment helper copy through i18n in every locale and removes the obsolete unsupported-type translation key.

### Build verification
- `pnpm --filter @openwork/app typecheck` -- failed on pre-existing unrelated errors in MCP/provider-auth/messaging/settings code before this PR's touched files.

### API/request-path verification
- Verified with a local mock OpenCode server that arbitrary file parts are sent via `POST /session/:id/prompt_async` with `type: "file"`, data URL, filename, and arbitrary MIME preserved.
- Verified both normal `promptAsync` and the OpenWork reasoning-effort override path preserve the same file payload.

### i18n verification
- Bun locale import check passed: every locale exports `composer.any_file_type_supported` and no locale keeps `composer.unsupported_attachment_type`.

### UI verification
- Attempted Chrome MCP verification, but the Chrome MCP bridge was unavailable (`127.0.0.1:9823` refused / not connected).
- Attempted local browser verification with Playwright against the Vite app and server-v2. The route loaded, but the local web flow could not fully reach the session surface because the dev OpenCode health route returned local connection/CORS issues (`/opencode/global/health`).
- No screenshots were committed.

## Test instructions

1. `pnpm install`
2. `pnpm --filter @openwork/app build`
3. Start the desktop app/dev stack.
4. Open a session composer.
5. Attach non-image/non-PDF files, for example `.zip`, `.csv`, extensionless files, or custom binary files.
6. Send the message and verify the OpenCode request includes file parts for each attachment.


src-opn:
## Verification

- Local mock OpenCode request-path check passed for arbitrary attachment MIME/file payloads through `POST /session/:id/prompt_async`.
- Locale import check passed for all locales with `composer.any_file_type_supported`; obsolete `composer.unsupported_attachment_type` removed.

UI verification note: Chrome MCP was unavailable locally (`127.0.0.1:9823` refused / not connected). Playwright could load the Vite route, but the local web session flow was blocked by local OpenCode health route connection/CORS issues, so no screenshot evidence was attached.

PR #1715, opened 2026-05-08
fix(updater): prevent architecture mismatch updates

src-opn:
## Summary

- Prevent Electron release matrix jobs from clobbering updater manifests by publishing packaged artifacts through a single manifest merge job.
- Add Windows arm64 Electron release coverage and validate the merged macOS feed includes both arm64 and x64 artifacts.
- Block app startup when the running Electron app architecture does not match the host architecture, with a direct download CTA for the correct build.

### Screenshots
- Not captured. The new UI only appears on architecture mismatch, which requires running a packaged wrong-architecture build under Rosetta or an equivalent host mismatch.

### Build verification
- `node --check apps/desktop/electron/main.mjs && node --check apps/desktop/electron/preload.mjs && node --check scripts/release/publish-electron-assets.mjs` -- passed
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-macos-aarch64.yml")'` -- passed
- Local mocked manifest merge test -- passed; generated `latest-mac.yml` included both `openwork-mac-arm64-...zip` and `openwork-mac-x64-...zip`
- `pnpm --filter @openwork/app typecheck` -- failed on existing repo-wide TypeScript errors unrelated to this change after dependencies were installed

### API verification
- No API changes.

### UI verification
- Not run through Chrome MCP because the gating condition depends on packaged app/process architecture mismatch rather than a normal browser route.

## Test instructions

1. Trigger `Release App` for a test tag with Electron publishing enabled.
2. Confirm the release upload includes one merged `latest-mac.yml` containing both `openwork-mac-arm64-<version>.zip` and `openwork-mac-x64-<version>.zip`.
3. On Apple Silicon, launch an x64 packaged build under Rosetta and verify the startup gate appears before runtime boot.
4. Click `Download correct version` and verify it opens the latest release asset URL for the system architecture.

### DIFF
--- a/.github/workflows/release-macos-aarch64.yml
+++ b/.github/workflows/release-macos-aarch64.yml
@@ -281,6 +281,11 @@ jobs:
             artifact: electron-windows-x64
             electron_args: "--win --x64"
             target_triple: x86_64-pc-windows-msvc
+          - platform: windows-2022
+            os_type: windows
+            artifact: electron-windows-arm64
+            electron_args: "--win --arm64"
+            target_triple: aarch64-pc-windows-msvc
 
     steps:
       - name: Checkout
@@ -374,7 +379,7 @@ jobs:
           TARGET: ${{ matrix.target_triple }}
         run: pnpm --filter @openwork/desktop build:electron
 
-      - name: Package + publish Electron (macOS, signed + notarized)
+      - name: Package Electron (macOS, signed + notarized)
         if: matrix.os_type == 'macos' && env.MACOS_NOTARIZE == 'true'
         env:
           CSC_LINK: ${{ secrets.APPLE_CODESIGN_CERT_P12_BASE64 }}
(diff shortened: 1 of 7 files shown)


src-opn:
## Verification

- `node --check apps/desktop/electron/main.mjs && node --check apps/desktop/electron/preload.mjs && node --check scripts/release/publish-electron-assets.mjs` passed.
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-macos-aarch64.yml")'` passed.
- Local mocked manifest merge test passed and verified the generated `latest-mac.yml` contains both arm64 and x64 macOS zip entries.
- `pnpm --filter @openwork/app typecheck` was attempted after `pnpm install --frozen-lockfile`; it failed on existing repo-wide TypeScript errors unrelated to this PR.

End-to-end packaged mismatch verification still needs a wrong-architecture packaged build on Apple Silicon/Rosetta or equivalent host mismatch.

src-opn:
## Follow-up

Updated the architecture mismatch download CTA to resolve the latest Electron updater manifest first, then select the artifact matching the system architecture. This avoids stale direct URLs when the installed app version is older than the current latest release.

Important behavior: this recovery path intentionally opens the installer download externally instead of using `electron-updater`, because a wrong-architecture app can cause `electron-updater` to select the running app architecture again.

PR #1719, opened 2026-05-08
fix(den): stabilize redirect loading layout

benjaminshafii:
## Summary

- Keep the Den auth screen in one stable two-column shell while session state resolves instead of switching to a standalone loading card.
- Replace the oversized progress-card loading state with a quieter inline account status panel.
- Keep checkout redirect/loading states inside the checkout page shell so returning from billing does not jump layouts.

### Build verification

- `pnpm install --frozen-lockfile` passed
- `pnpm --filter @openwork-ee/den-web build` passed

### UI verification

- Code-level verification: the `AuthScreen` no longer has an early `!sessionHydrated` return, so the same auth shell renders before and after session hydration.
- Code-level verification: checkout auth/redirect states now render through `CheckoutStatusPanel`, which uses the same `den-page` and hero shell as the checkout page.
- Browser screenshot evidence was not captured in this turn because the reported state is transient during session/redirect resolution.

### API verification

- No API changes.

## Test instructions

1. `pnpm install --frozen-lockfile`
2. `pnpm --filter @openwork-ee/den-web build`
3. Run Den web locally.
4. Open `/` while signed out and confirm session checking does not jump from a standalone centered card to the auth layout.
5. Sign in from `/` and confirm the redirect state stays in the right-side auth panel slot.
6. Visit `/checkout` while unauthenticated or after checkout return and confirm the status appears within the checkout shell instead of a separate loading layout.

### DIFF
--- a/ee/apps/den-web/app/(den)/_components/auth-screen.tsx
+++ b/ee/apps/den-web/app/(den)/_components/auth-screen.tsx
@@ -18,17 +18,37 @@ function FeatureCard({ title, body }: { title: string; body: string }) {
   );
 }
 
-function LoadingPanel({ title, body }: { title: string; body: string }) {
+function SessionStatusPanel({ mode }: { mode: "checking" | "redirecting" }) {
+  const status = mode === "checking"
+    ? {
+        title: "Checking account",
+        body: "If you are already signed in, we will open your workspace. Otherwise you can continue here.",
+      }
+    : {
+        title: "Opening workspace",
+        body: "You are signed in. We are taking you to the right Cloud destination.",
+      };
+
   return (
-    <div className="den-frame grid gap-3 p-6 md:p-7">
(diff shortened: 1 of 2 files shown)


benjaminshafii:
## Verification

Ran on branch `task/fix-den-redirect-loading` based on `origin/dev` at `2ae389ae`.

- `pnpm install --frozen-lockfile` passed
- `pnpm --filter @openwork-ee/den-web build` passed
- `git diff --check` passed before commit

Implementation checks:

- `AuthScreen` now renders the same two-column auth shell while `sessionHydrated` is false; the previous standalone `Checking your session` card path was removed.
- Signed-in redirect state now occupies the auth panel column using `SessionStatusPanel`, avoiding a layout switch.
- Checkout loading/redirect states now use `CheckoutStatusPanel` inside the checkout shell instead of returning a small standalone card.

Screenshot evidence was not captured in this turn because the changed state is transient during session/redirect resolution.

PR #1753, opened 2026-05-11
fix(server): guard response writes after end

### LINKED ISSUE #1743: Bug: ERR_STREAM_WRITE_AFTER_END crash in Electron main process (in-process server #1675)
### Description
OpenWork v0.13.5 repeatedly crashes with `ERR_STREAM_WRITE_AFTER_END` in the Electron main process. The error appears as an `Uncaught Exception` dialog and occurs during normal usage of the desktop app.
### Error Stack
Error ERR_STREAM_WRITE_AFTER_END: write after end
    at ServerResponse.end (node:_http_outgoing:1097:15)
    at Server.<anonymous> (file:///C:/Users/yj/AppData/Local/Programs/@openworkdesktop/resources/app...:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
### Platform
- **OS**: Windows (x64)
- **OpenWork Version**: v0.13.4 (first seen), v0.13.5 (still present)
- **Installation type**: Electron EXE installer (from GitHub Releases)
### Root Cause
PR #1675 ("In-process server") moved the OpenWork HTTP server from a separate sidecar process into the Electron main process. The HTTP response handler has a race condition where `res.end()` is called twice on the same `ServerResponse` object. 
Previously, this would only crash the sidecar process (which would auto-restart). Now it crashes the entire Electron main process, making the app unusable.
Relevant PRs:
- **#1675** — In-process server (introduced in v0.13.4)
- **#1674** — Tauri fully removed, Electron is sole desktop runtime
### Steps to Reproduce
1. Install OpenWork v0.13.5 on Windows (Electron build)
2. Launch the desktop application
3. Use the app normally — the error appears repeatedly at runtime
### Workaround
✅ **Downgrading to v0.13.3 resolves the issue completely.** v0.13.3 does not include #1675 (server runs as a separate sidecar process), and the app runs without any errors.
Download: https://github.com/different-ai/openwork/releases/tag/v0.13.3
`openwork-win-x64-0.13.3.exe`
### Suggested Fix
Fix the `res.end()` double-call race condition in the in-process HTTP server's response handling (in `app.asar`). The handler at the reported call site should check if the response stream has already been closed before calling `.end()` or `.write()`.
```js
// Suggested guard pattern
if (!res.writableEnded) {
  res.end(...)
}
Additional Context
- First reported: May 9, 2026
- This bug likely affects all v0.13.4+ users, especially on Windows
- The error is intermittent (race condition), not deterministic on every action
---

src-opn:
## Summary

- Guard the Node HTTP adapter from writing or ending a `ServerResponse` after it has already closed.
- Handle `ERR_STREAM_WRITE_AFTER_END` as a response-stream lifecycle error instead of letting it crash Electron main process.
- Add regression coverage for a streaming response that fails after response output has started.

Fixes #1743.

### Screenshots
No UI changes.

### Build verification
- `bun test apps/server/src/serve-node.test.ts` -- passed
- `pnpm --filter openwork-server typecheck` -- passed
- `pnpm --filter openwork-server build` -- passed
- `pnpm --filter openwork-server build:bin` -- passed

### API verification
Server adapter lifecycle change only; no endpoint contract changes. Regression test verifies the server remains alive after the failing streaming response and can still serve `/health`.

### UI verification
No UI changes.

## Test instructions

1. `pnpm install --frozen-lockfile`
2. `bun test apps/server/src/serve-node.test.ts`
3. `pnpm --filter openwork-server typecheck`
4. `pnpm --filter openwork-server build:bin`

### DIFF
--- a/apps/server/src/serve-node.test.ts
+++ b/apps/server/src/serve-node.test.ts
@@ -1,0 +1,53 @@
+import { describe, expect, test } from "bun:test";
+import { setTimeout as delay } from "node:timers/promises";
+import { serve } from "./serve-node.js";
+
+describe("serve", () => {
+  test("does not write an error response after a streaming response has ended", async () => {
+    const uncaught: unknown[] = [];
+    const onUncaughtException = (error: unknown) => {
+      uncaught.push(error);
+    };
+    process.on("uncaughtException", onUncaughtException);
+
+    const encoder = new TextEncoder();
+    const server = await serve({
+      hostname: "127.0.0.1",
+      port: 0,
+      fetch: (request) => {
+        if (new URL(request.url).pathname === "/health") {
+          return Response.json({ ok: true });
+        }
+
+        let wroteChunk = false;
+        return new Response(
+          new ReadableStream<Uint8Array>({
+            pull(controller) {
+              if (!wroteChunk) {
+                wroteChunk = true;
+                controller.enqueue(encoder.encode("partial"));
+                return;
+              }
+              controller.error(new Error("stream failed after response started"));
+            },
+          }),
+        );
+      },
+    });
+
+    try {
+      const response = await fetch(`http://127.0.0.1:${server.port}/stream`);
+      await response.text().catch(() => undefined);
+      await delay(25);
+
+      expect(uncaught).toEqual([]);
+
+      const health = await fetch(`http://127.0.0.1:${server.port}/health`);
+      expect(health.status).toBe(200);
+      expect(await health.json()).toEqual({ ok: true });
+    } finally {
+      process.off("uncaughtException", onUncaughtException);
+      server.stop();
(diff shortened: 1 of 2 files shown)


src-opn:
## Verification

Ran in `/Users/omar/code/factory/_repos/openwork/_worktrees/crash-on-exit`:

- `bun test apps/server/src/serve-node.test.ts` -- passed
- `pnpm --filter openwork-server typecheck` -- passed
- `pnpm --filter openwork-server build` -- passed
- `pnpm --filter openwork-server build:bin` -- passed

The regression test simulates a streaming response that errors after sending a chunk, confirms no uncaught process exception is raised, and verifies the server can still respond to a follow-up `/health` request.

PR #1769, opened 2026-05-12
fix(app): sync desktop active org before switching

src-opn:
## Summary

- Await the Better Auth organization `set-active` call before updating the desktop app's local active organization.
- Keep the previous organization selected and surface an error if the server-side active org switch fails.
- Align desktop Settings > Cloud > Account behavior with den-web so org-scoped cloud pages refetch against the newly selected org.

### Build verification

### API verification
- No API changes. Desktop now calls the existing `/api/auth/organization/set-active` flow through `client.setActiveOrganization(...)`.

### UI verification
- Not run via Chrome MCP in this pass; the local desktop/Den stack was not booted for this PR.
- Code path verified against den-web's existing active-org switch flow in `ee/apps/den-web/app/(den)/o/[orgSlug]/dashboard/_providers/org-dashboard-provider.tsx`.

## Test instructions

1. Start the desktop app against a Den account with at least two orgs.
2. Go to Settings > Cloud > Account.
3. Switch the active organization.
4. Open Marketplace and Plugins, Cloud Workers, and Cloud Providers.
5. Verify each page returns data for the newly selected organization, not the previous default org.

### DIFF
--- a/apps/app/src/app/lib/den.ts
+++ b/apps/app/src/app/lib/den.ts
@@ -1245,7 +1245,7 @@ async function ensureActiveOrganization(
     return;
   }
 
-  await requestJson<unknown>(baseUrls, "/api/auth/organization/set-active", {
+  await requestJson<unknown>(baseUrls, "/v1/me/active-organization", {
     method: "POST",
     token,
     body: {
--- a/apps/app/src/react-app/domains/settings/cloud/cloud-account-section.tsx
+++ b/apps/app/src/react-app/domains/settings/cloud/cloud-account-section.tsx
@@ -26,7 +26,7 @@ export interface CloudAccountSectionProps {
   orgsBusy: boolean;
   orgsError: string | null;
   sessionBusy: boolean;
-  onActiveOrgChange: (orgId: string) => void;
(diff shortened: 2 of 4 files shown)


src-opn:
## Verification

- Confirmed den-web switches orgs by posting to `/api/auth/organization/set-active` before loading org-scoped data.
- Desktop now awaits the same Better Auth active-org update before writing local active org settings and notifying cloud settings consumers.

UI e2e was not run in this pass because the desktop/Den stack was not booted. Manual verification steps are included in the PR body.

src-opn:
## Update

Added a hidden Den API route for desktop bearer-token sessions:

- `POST /v1/me/active-organization` with `hide: true` in OpenAPI metadata.
- Uses existing Den API `sessionMiddleware`, so desktop bearer tokens resolve to the same DB-backed Better Auth session row.
- Verifies the target org belongs to the signed-in user, then updates `session.activeOrganizationId` in `AuthSessionTable`.
- Desktop now calls this route instead of Better Auth's cookie-oriented `/api/auth/organization/set-active` endpoint.

## Verification

- `pnpm --filter @openwork-ee/den-api build` passed.

PR #1793, opened 2026-05-14
fix(cloud): apply org provider config after sign-in

src-opn:
## Summary

- Move multi-org selection into the post-sign-in onboarding flow so users choose the org before org resources are imported.
- Run org-managed provider sync from the Session route and on Den settings changes so `opencode.json` updates trigger reload without visiting Settings.
- Queue the existing new-provider/default-model prompt until onboarding closes and refresh the composer model picker when providers change.

### Build verification
- `pnpm --filter @openwork/app typecheck` -- passed

### API verification
- No API changes.

### UI verification
- Not run via Chrome MCP in this session; changes were verified by typecheck/build and manual reasoning against the reported desktop sign-in flow.
- Expected flow: sign in, choose org in onboarding if multiple orgs exist, continue through org summary, OpenCode reload applies config, then the new-provider prompt can open the Models picker.

## Test instructions

1. Start the desktop app against Den.
2. Sign in with a user that belongs to two orgs.
3. Confirm onboarding shows the org picker before the org summary.
4. Select an org with managed providers and continue.
5. Confirm OpenCode reloads/applies the imported provider config.
6. Confirm the bottom new-provider prompt appears after onboarding and opens the Models picker.
7. Confirm the composer model dropdown includes the newly imported provider after reload.

### DIFF
--- a/apps/app/src/components/model-select.tsx
+++ b/apps/app/src/components/model-select.tsx
@@ -31,6 +31,7 @@ import { isDesktopProviderBlocked } from "@/app/cloud/desktop-app-restrictions";
 import { readHiddenModels } from "@/react-app/domains/session/modals/model-picker-modal";
 import { Settings2 } from "lucide-react";
 import { openModelPickerEvent } from "@/react-app/shell/new-providers-toast";
+import { newProvidersEvent } from "@/app/lib/provider-events";
 
 function getProviderDisplayName(providerId: string) {
   return providerId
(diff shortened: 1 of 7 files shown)


src-opn:
## Verification

Ran in `/Users/omar/code/factory/_repos/openwork/_worktrees/fix-desktop-app-config-apply`:

- `pnpm --filter @openwork/app typecheck` -- passed

UI automation was not run in this session because the Den/desktop dev stack was not already running. Manual verification target is the desktop sign-in flow: multi-org sign-in should route through onboarding org selection, import org providers, reload OpenCode after onboarding, and refresh both the bottom Models prompt and composer model dropdown.

PR #1792, opened 2026-05-14
fix(updater): scope electron checks to release channel

src-opn:
## Summary

- Pass the selected release channel through Electron updater checks so Settings probes the active feed instead of relying on previously persisted state.
- Stop the packaged startup path from launching a silent `checkForUpdates()` that can race Alpha channel sync and cache a Stable-feed result.
- Keep Electron updater initialization in place while making renderer-owned checks the source of truth for channel-scoped update status.

### Screenshots
No screenshot captured. This change targets packaged Electron updater IPC/feed behavior, which is not fully exercisable in the dev renderer without a signed packaged app.

### Build verification
- `pnpm install --frozen-lockfile --prefer-offline` -- passed
- `pnpm --filter @openwork/app build` -- passed
- `node --check apps/desktop/electron/updater.mjs && node --check apps/desktop/electron/preload.mjs && node --check apps/desktop/electron/main.mjs` -- passed
- `git diff --check` -- passed
- `pnpm --filter @openwork/app typecheck` -- failed on existing unrelated React type mismatch in `src/react-app/domains/session/surface/markdown.tsx` between `react-markdown` and `streamdown` component types

### Updater verification
- Fetched live alpha feed: `https://github.com/different-ai/openwork/releases/download/alpha-macos-latest/latest-mac.yml`
- Confirmed it advertises `0.13.9-alpha.823`
- Confirmed comparator treats `0.13.9-alpha.823` as newer than installed `0.13.9-alpha.819+23a856e`

### API verification
No API changes.

### UI verification
Not run through Chrome MCP. The affected path depends on packaged Electron `electron-updater` behavior and signed release feeds, not browser/dev renderer behavior.

## Test instructions

1. Install an older packaged Electron alpha build, for example `0.13.9-alpha.819+23a856e`.
2. Open Settings -> Updates.
3. Select `Alpha` as the release channel.
4. Click `Check for updates`.
5. Verify the app reports `0.13.9-alpha.823` or newer as available from the alpha feed.

PR #1794, opened 2026-05-14
fix(updater): avoid cloning click events

src-opn:
## Summary

- Wrap the Updates page updater button handlers so React click events are not passed into updater actions.
- Prevent `checkForUpdates(channelOverride?)` from receiving a synthetic click event and forwarding it through Electron IPC.
- Fix the packaged app error: `An object could not be cloned.`

### Screenshots
No screenshot captured. This is a one-line UI event wiring fix; the packaged Electron error depends on clicking the Updates button in a signed app build.

### Build verification
- `pnpm install --frozen-lockfile --prefer-offline` -- passed
- `pnpm --filter @openwork/app typecheck` -- passed
- `git diff --check` -- passed

### API verification
No API changes.

### UI verification
Local packaged Electron UI was not exercised. Static verification confirms no direct `onClick={props.checkForUpdates}` / `downloadUpdate` / `installUpdateAndRestart` handlers remain in `updates-view.tsx`, so click events are no longer forwarded as function arguments.

## Test instructions

1. Install an alpha build containing PR #1792, for example `0.13.9-alpha.825+adf54d1`.
2. Open Settings -> Updates.
3. Click `Check for updates`.
4. Verify the page no longer shows `An object could not be cloned.`
5. If on Alpha, verify the check completes against the selected Alpha channel.

### DIFF
--- a/apps/app/src/react-app/domains/settings/pages/updates-view.tsx
+++ b/apps/app/src/react-app/domains/settings/pages/updates-view.tsx
@@ -140,7 +140,7 @@ export function UpdatesView(props: UpdatesViewProps) {
                   <div className="flex flex-wrap items-center gap-2">
                     <Button
                       variant="outline"
-                      onClick={props.checkForUpdates}
+                      onClick={() => void props.checkForUpdates()}
                       disabled={props.busy || updateState === "checking" || updateState === "downloading"}
                     >
                       {updateState === "checking" ? <Spinner className="size-4" /> : null}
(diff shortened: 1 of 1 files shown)


src-opn:
## Verification

- `pnpm install --frozen-lockfile --prefer-offline`: passed.
- `pnpm --filter @openwork/app build`: passed.
- `pnpm --filter @openwork/app typecheck`: passed.
- `git diff --check`: passed.
- Confirmed the direct updater callbacks were removed from `updates-view.tsx`, so React click events are no longer forwarded into Electron IPC.

The packaged Electron app flow was not rerun locally; the failure mode was traced to React passing the click event into `checkForUpdates(channelOverride?)`, which then attempted to IPC-clone the event object.

PR #1797, opened 2026-05-15
fix(app): refresh connected provider models

src-opn:
## Summary

- Route model/provider lists through the connected-aware OpenCode provider endpoint with a shared 5-minute React Query cache.
- Detect connected provider/model changes after refresh or reload and surface the existing bottom toast with new provider/model counts.
- Limit model pickers to connected providers, preserve OpenCode provider visibility, and block task submission when the selected model is no longer available.

### Screenshots
Not captured. This PR changes provider/model picker state and requires a live OpenWork runtime with model/provider changes to demonstrate visually.

### Build verification
- `pnpm typecheck` from `apps/app` -- passed

### API verification
No API changes.

### UI verification
Manual code-path verification: unified model picker reads now use `provider.list()` via the shared provider-list query; OpenCode reload paths invalidate/refetch provider-list queries; unavailable models disable composer submission and show inline warning.

## Test instructions

1. Start the OpenWork app against a workspace with at least one connected provider.
2. Open the session model picker and Settings > Preferences > Model; verify only connected provider models are shown, including OpenCode when connected.
3. Simulate a provider/model refresh or reload and verify the bottom toast shows new provider/model counts.
4. Select or retain a model that disappears from connected providers; verify Run task is disabled and `Model no longer available` appears by the model picker.

### DIFF
--- a/apps/app/src/app/lib/provider-events.ts
+++ b/apps/app/src/app/lib/provider-events.ts
@@ -18,9 +18,11 @@ export type NewProviderInfo = {
 
 export type NewProvidersEventDetail = {
   providers: NewProviderInfo[];
+  newProviderCount?: number;
+  newModelCount?: number;
   /** Where the change originated. "sign_in" is suppressed by the toast
    *  because the onboarding page handles first-time notification. */
-  source: "cloud_sync" | "local_config" | "sign_in";
+  source: "cloud_sync" | "local_config" | "models_refresh" | "sign_in";
 };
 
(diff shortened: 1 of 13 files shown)


src-opn:
## Verification

- `pnpm typecheck` from `apps/app` passed after building `packages/ui`.
- `pnpm --filter @openwork/app build` passed after building `packages/ui`.
- No API changes.
- UI flow was not screenshotted because this needs a live provider/model refresh scenario; test instructions are included in the PR body.

PR #1803, opened 2026-05-15
fix(inference): return OpenAI-compatible rate limit errors

src-opn:
## Summary

- Return an OpenAI-compatible 429 error envelope for exceeded inference usage limits.
- Add token rate-limit headers and retry timing for exhausted usage buckets.
- Preserve OpenWork bucket metadata in vendor-specific response headers.

### Screenshots
No UI changes.

### Build verification
- `pnpm i` -- passed
- `pnpm --filter @openwork-ee/inference build` -- passed

### API verification
Not run against a live inference database; behavior is covered by build/type verification for the changed proxy and limit metadata path.

### UI verification
No UI changes.

## Test instructions

1. Configure an org whose active inference usage bucket is exhausted.
2. Send an authenticated request through `ee/apps/inference` under `/api/v1/*`.
3. Verify the response is HTTP 429 with `error.code: rate_limit_exceeded`, `error.type: tokens`, `error.param: null`, and token rate-limit headers.

### DIFF
--- a/ee/apps/inference/src/limits.ts
+++ b/ee/apps/inference/src/limits.ts
@@ -107,7 +107,18 @@ export async function ensureUsableBuckets(organizationId: string, now = new Date
     }
     const remaining = effectiveLimit - bucket.used_amount
     if (remaining <= 0) {
-      return { ok: false as const, bucketIds, bucketLimits, limitedBy: bucket.id, windowType: policy.window_type }
+      return {
+        ok: false as const,
+        bucketIds,
+        bucketLimits,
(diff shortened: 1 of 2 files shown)


src-opn:
## Verification

- Installed dependencies with `pnpm i`: passed.
- Built the changed package with `pnpm --filter @openwork-ee/inference build`: passed.
- No UI verification was run because this change only affects the inference proxy 429 API response.
- Live API curl was not run because this requires an exhausted inference usage bucket in a configured inference database.

PR #1840, opened 2026-05-18
fix(app): contain native browser pane

benjaminshafii:
## Summary\n- measure the native browser WebContentsView against the right-pane content placeholder instead of the full panel\n- keep native bounds synchronized via layout/RAF/resize/scroll updates, including Fast Refresh stale-bound recovery\n- document screenshot evidence for browser containment in the session eval\n\n## Verification\n- pnpm --filter @openwork/app typecheck: passed\n- pnpm --filter @openwork/app build: passed\n- Local Electron CDP screenshot inspected after hot reload: /var/folders/d9/xqhkvsp94rg0n0n523snqztm0000gn/T/browser-screenshot-1779146278675.png showed browser content contained inside the right pane without transcript/composer overlap\n


benjaminshafii:
Superseded by the combined restore PR #1842, which includes the browser containment fix plus artifact/openable-items and reload-toast fixes.

PR #1866, opened 2026-05-21
fix(app): resolve extension icon asset paths

benjaminshafii:
## Summary
- resolve extension icon URLs through Vite's base URL before rendering
- keep CDN Simple Icons fallback behavior in one shared helper
- fixes Electron packaged prod builds where root-relative public assets resolve to file system root

## Validation
- pnpm install (needed because the isolated worktree had no node_modules)
- OPENWORK_ELECTRON_BUILD=1 pnpm --filter @openwork/app build
- confirmed built Electron bundle contains resolver output that maps root-relative icons to ./ paths and dist includes ext-*.svg public assets

## Evidence
- No screenshot/video captured; validated via production Electron asset build output and compiled bundle inspection.


src-opn:
already solved

PR #1876, opened 2026-05-21
fix(app): limit artifacts to written files

src-opn:
## Summary
- Limit file artifact candidates to write/edit/apply_patch-style tool activity instead of read/search output or prose-only mentions.
- Keep URL detection unchanged and allow inline message mentions to display only when they match globally verified written artifact targets.
- Add coverage for write metadata, read exclusion, apply_patch paths, and discovery/search exclusions.

## Verification
- `pnpm --filter @openwork/app test:open-target` passed.

## E2E Evidence
No video or screenshots captured; this is a logic-only artifact extraction change. Reviewer can reproduce with the commands above.


JeanMeijer:
@cubic review

PR #1898, opened 2026-05-22
chore(server): use OpenCode SDK for API calls

JeanMeijer:
## Summary
- Replaces server-side OpenCode HTTP calls with the OpenCode SDK.
- Simplifies server API/session handling around SDK methods.
- Updates the session read model to match SDK-backed responses.
- Adds the SDK dependency to the server package and lockfile.


JeanMeijer:
@cubic review

Issue #1777, opened 2026-05-13
Feature request: native path-injection convention for non-multimodal attachments in the web composer


MariusWilsch:
# Feature request: native path-injection convention for non-multimodal attachments in the web composer

## Summary

The web composer (`apps/app`) currently has a single attachment-to-prompt shape: convert the uploaded file to a base64 `data:` URL and emit a `type: "file"` content part. This works cleanly for media types that downstream providers accept as multimodal inputs — images, PDFs. It does not provide a path for the increasingly common case where the agent should **reason about the file via tools** (Read, MCP tool calls, Bash) rather than ingest its bytes as multimodal content.

The result, today, is that workspaces wiring up MCP servers for spreadsheet / CSV / document analysis (a use case OpenWork's docs front-and-center via the MCP server feature) have no upstream-supported way to surface "the user attached file X; it's at path Y; decide what to do with it" to the agent.

This issue proposes opening a design conversation about a native path-injection convention in the web composer.

## Why the current shape doesn't fit this case

The `attachmentToFilePart` path in [`apps/app/src/react-app/domains/session/sync/actions-store.ts`](https://github.com/different-ai/openwork/blob/dev/apps/app/src/react-app/domains/session/sync/actions-store.ts) emits:

```ts
{
  type: "file",
  url: await fileToDataUrl(attachment.file), // data:<mime>;base64,...
  filename: attachment.name,
  mime: attachment.mimeType,
}
```

For attachments outside the multimodal-supported set:

- Anthropic's Messages API rejects non-PDF MIME types in `document.source.base64` (only `application/pdf` accepted; non-PDF files are explicitly redirected to the [Files API](https://platform.claude.com/docs/en/build-with-claude/files), a separate beta endpoint).
- OpenRouter passes the same constraint through unchanged (its `file` content type is documented exclusively for PDFs in [its multimodal/pdfs guide](https://openrouter.ai/docs/guides/overview/multimodal/pdfs)).
- Other providers vary, but the multimodal-bytes-in-payload shape is fundamentally about model-ingestible content, not file-as-path.

So in the web composer, attachments outside the model-ingestible set effectively can't reach the agent at all today — they're encoded into a payload the downstream provider drops.

## The deeper design gap

Even if every provider accepted every MIME, the multimodal-bytes-in-payload shape is the **wrong primitive** for many real workflows. Consider:

[...]

PR #1903, opened 2026-05-22
feat(app): fall back to text part for non-multimodal attachments

yashranaway:
Refs #1777

## Summary
- Branch composer attachments by MIME instead of unconditionally encoding every file as a base64 `type: "file"` data URL. Images and PDFs keep the existing multimodal path. Text-like files (`text/*` plus a small allowlist of `application/{json,xml,yaml,csv,javascript,typescript,sql,x-sh,x-yaml}`) are emitted as `type: "text"` parts with the content inlined between `[attached file: name]` / `[end of name]` markers so the agent can reason about them without provider-side rejection.
- Cap inline text at 256 KB and emit a clear "too large to inline, save it to the workspace" notice above that size so very large CSVs do not balloon the prompt.
- For binary attachments outside the multimodal set (xlsx, zip, etc.), emit a placeholder text part naming the file and asking the agent to read it from the workspace via Read/Bash, instead of silently sending a payload the provider will drop.
- Slash commands still accept file parts only (SDK constraint), so the new text-fallback parts are filtered out of `buildCommandFileParts` rather than passed through as an invalid shape.

## Verification
- `pnpm --filter @openwork/app typecheck` reports the same pre-existing `origin/dev` errors as before (missing `@heroicons/react/24/solid`, `xlsx`, `marked-emoji`, `marked-shiki`, `emojilib`, `@shikijs/transformers`, `shiki` type decls); no new errors land in `actions-store.ts`.
- `node scripts/i18n-audit.mjs --ci` passes — no i18n surface changed.

## Notes
- The reporter on #1777 raised the broader design question of writing attachments to a real workspace path so MCP/Bash can reach them via tools. That is the right long-term shape but needs the Electron bridge plumbing and per-platform tempdir lifecycle. This PR is the narrower fix that unblocks "agent reasons about a CSV/JSON/code attachment" today without provider-side drops, and leaves the path-on-disk story for a follow-up that pairs with the desktop write surface.
- The MIME allowlist is intentionally small; happy to grow it once we see common attachment types in user reports.

### DIFF
--- a/apps/app/src/react-app/domains/session/sync/actions-store.ts
+++ b/apps/app/src/react-app/domains/session/sync/actions-store.ts
@@ -61,6 +61,51 @@ const fileToDataUrl = (file: File) =>
     reader.readAsDataURL(file);
   });
 
+const fileToText = (file: File) =>
+  new Promise<string>((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onerror = () => reject(new Error(`Failed to read attachment: ${file.name}`));
+    reader.onload = () => {
+      resolve(typeof reader.result === "string" ? reader.result : "");
+    };
+    reader.readAsText(file);
+  });
+
+// MIME types that downstream providers accept as multimodal bytes in the prompt.
+// Anthropic accepts image/* + application/pdf in messages; OpenAI accepts image/* + application/pdf.
+// Anything else is rejected or silently dropped, so we route those through a text fallback instead.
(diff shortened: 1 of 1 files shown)


yashranaway:
Superseded by upstream PR #1993 (`fix(app): resolve text attachments broken`), which solves the same crash by rewriting non-multimodal MIMEs to `text/plain` inside `attachmentMime()` rather than emitting a separate text part. The underlying poisons-history bug is no longer reproducible. Closing.

PR #1944, opened 2026-05-25
fix(app): resolve spaced file mention round-tripping in editor

JeanMeijer:
## Summary
- Fixes editor round-tripping for file mentions with spaces in their paths.
- Adds mention encoding/decoding helpers used by composer/session surface.
- Updates editor mention serialization to preserve spaced file paths.
- Adds tests for mention encoding behavior.


benjaminshafii (approved):
Approved via automated Daytona Electron smoke test (agent). Tested: Electron boot, app render, workspace creation, session view, settings navigation, back-to-app. All passed, no regressions from the spaced file mention fix.

PR #1945, opened 2026-05-25
Persist session activity status

src-opn:
## Summary
- Add a global session activity store keyed by workspace/session with idle, thinking, responding, waiting, compacting, and error states.
- Feed the store from the existing workspace SSE sync path so session status survives view remounts without adding another subscription.
- Use the shared status in the session surface and sidebar active indicators.

## Tests

## Evidence
- No video or screenshot captured; this is a state plumbing change verified with typecheck.


src-opn:
Addressed cubic review comments:

- Restored explicit streaming status checks instead of treating every non-idle state as streaming.
- Added a translated idle activity label.
- Reused the shared session activity label helper in the session surface.
- Added clearError and clear activity error state when dismissing the error card.

Verified with pnpm --filter @openwork/app typecheck.

PR #1960, opened 2026-05-26
chore(app): enable React Compiler in annotation mode

JeanMeijer:
## Summary
- Enables React Compiler in annotation mode for the app Vite config.
- Adds the React Compiler runtime dependency.


benjaminshafii (approved):
Approved via automated Daytona Electron smoke test (agent). Tested: Electron boot, app render (React Compiler annotation mode), workspace creation, settings navigation, back-to-app. All passed, no regressions detected.

PR #1964, opened 2026-05-26
fix(server): resolve artifacts not shown due to lack of absolute path handling

JeanMeijer:
## Summary
- Allows artifact target resolution to accept absolute file paths when they are inside the workspace, converting them back to workspace-relative paths.
- Skips absolute paths that point outside the workspace.
- Updates the artifact files E2E test to cover absolute in-workspace paths and preserve the higher confidence score when duplicate targets resolve to the same file.


benjaminshafii (approved):
Approved via automated Daytona Electron smoke test (agent). Tested: Electron boot, app render, settings navigation, back-to-app. All passed, no regressions detected.

PR #1997, opened 2026-05-29
refactor(config): centralize workspace config helpers in openwork-server (1/3)

benjaminshafii:
## What

First of a 3-PR cleanup of how OpenWork stores/injects config, plugins, and extensions. This one is a **pure, behavior-preserving refactor** — the foundation the next two build on.

Makes `apps/server` the single source of truth for:
- the per-workspace `openwork.json` schema + defaults
- the read/write helpers for it
- the default `opencode.json` writer

…and has the Electron main process consume them instead of keeping its own drifting copies.

## Why

The same path-resolution, `openwork.json` schema, and `opencode.json` default were defined **twice** — once in `apps/server/src` (the in-process server that actually boots on every launch) and once hand-rolled in `apps/desktop/electron/main.mjs` (the IPC handlers). They had already drifted. This collapses them to one definition.

## Changes

- `apps/server/src/workspace-files.ts` — now exports `WorkspaceOpenworkConfig` + `defaultWorkspaceOpenworkConfig` / `readWorkspaceOpenworkConfig` / `writeWorkspaceOpenworkConfig` (pure; the write helper returns the path string, no Electron-shaped return value).
- `apps/server/src/workspace-init.ts` — exports `ensureOpencodeConfig`; its `ensureWorkspaceOpenworkConfig` now reuses the shared default + writer (removes the duplicate inline schema).
- `apps/desktop/electron/main.mjs` — imports those helpers **lazily from the compiled server bundle** by absolute path, mirroring the existing `runtime.mjs:1052` dev-vs-packaged resolution. Drops the local `defaultWorkspaceOpenworkConfig`, `read/writeWorkspaceOpenworkConfig`, `ensureDefaultWorkspaceOpencodeConfig`, `workspaceOpencodeConfigPath`, and the now-unused `BROWSER_PLUGIN` const. The IPC handlers keep their exact return contracts (`execResult` wrapping for writes, raw config object for reads).

## Behavior change

None. Same file paths, same schema, same config content. `check-electron-bridge.mjs` only matches `case` labels (not handler bodies), so the swapped internals don't affect the bridge contract.

## Verification

- `pnpm --filter openwork-server typecheck` ✅
- `pnpm --filter openwork-server build` ✅ (exports present in `dist`)
- `bun test` workspace-init / jsonc / portable-files / portable-opencode — 15 pass ✅
- `pnpm --filter @openwork/desktop typecheck:electron` ✅
- `node apps/desktop/scripts/check-electron-bridge.mjs` — 50 methods covered ✅
- `node --test` runtime + remote-workspace — 13 pass ✅
- Adversarial multi-agent review (behavior-equivalence / runtime-packaging / regression-scope) — no confirmed findings ✅

## Follow-ups (stacked on this branch)

- **2/3** — move `openwork.json` out of `.opencode/` → `.openwork/` (with migration)
- **3/3** — single source of truth for the plugin injection channel (env `OPENCODE_CONFIG_CONTENT`)

### DIFF
--- a/apps/desktop/electron/main.mjs
+++ b/apps/desktop/electron/main.mjs
@@ -17,7 +17,7 @@ import {
 import { createRequire } from "node:module";
 import os from "node:os";
 import path from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 import { app, BrowserWindow, Menu, WebContentsView, clipboard, dialog, ipcMain, nativeImage, nativeTheme, session, shell, systemPreferences } from "electron";
 import { configureFakeMediaForTests, installMediaPermissionHandlers } from "./media-permissions.mjs";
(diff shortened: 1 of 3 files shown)


src-opn:
please wait for https://github.com/different-ai/openwork/pull/1999

PR #2004, opened 2026-05-29
feat(desktop): add steer/queue follow-up messages while agent is busy

src-opn:
## Summary
While the agent is running, the composer now offers **Steer**, **Queue**, and an outline **Stop** instead of a single Stop button:

- **Steer** — sends the message immediately. opencode accepts a mid-run user turn and injects it into the running loop (as a `<system-reminder>`), so the agent adjusts mid-task.
- **Queue** — opencode has no server-side queue, so messages are held client-side and **auto-sent once the session goes idle**. Multiple queued messages are merged and delivered as a single message.
- **Stop** — aborts the run (`session.abort`), styled as an outline button.

Queued messages appear in a panel above the composer (mirroring the question panel). Each entry has an X to remove it, and the panel hides when the queue is empty.

### Keyboard behavior (only while busy)
- **Enter** does not submit; it shakes the Steer/Queue buttons to prompt an explicit choice.
- **Escape** arms a "Hit Escape again to stop the agent" message for 3s; a second Escape stops the agent.

## Implementation notes
- Steer/Queue are disabled until there's text; Stop is always active.
- `mergeDrafts` concatenates queued drafts (text + parts + attachments) into one send; on failure the queue is restored.
- Steered/queued messages render as normal user-message bubbles via the existing snapshot+SSE transcript path (no special styling). Note: there is no optimistic echo, so the bubble appears once the server echoes it back over SSE (sub-second).

## Files
- `composer.tsx` — Steer/Queue/Stop UI, shake + Escape-to-stop logic.
- `session-surface.tsx` — `handleSteer`/`handleQueue`, local queue state, idle-drain effect, `mergeDrafts`.
- `queued-messages-panel.tsx` — new panel listing queued messages with per-item remove.
- `index.css` — `animate-shake` keyframe/utility.
- `en.ts` — new i18n keys (English is the fallback for all locales).

## Testing
Commands run from `apps/app`:
- `pnpm typecheck` — passes.
- `pnpm build` — succeeds.

I have **not** captured an end-to-end video/screenshots yet. To reproduce manually: start a session, send a prompt, and while the agent is streaming, type a follow-up and try Steer / Queue / Stop, plus Enter (shake) and Escape-Escape (stop).


src-opn:
## CI test failures: pre-existing / infra, not from this PR

The `OpenWork Tests` job fails with `Timed out waiting for /global/health: fetch failed` — the embedded opencode server didn't finish booting in the e2e harness. This is **not related to this PR**:

- This PR is React/UI-only (composer + transcript); the API e2e suite (`scripts/e2e.mjs`) actually **passed** on macOS (`"ok": true`, all steps ok). The timeout is in a later script's own server boot.
- `OpenWork Tests` is **already red on the base `dev` branch** for the last several commits (#2003, #2001, #1998 all show `OpenWork Tests = failure`).

Local verification on this branch:
- `pnpm typecheck` — passes
- `pnpm build` — succeeds
- `node scripts/i18n-audit.mjs` — no dangling/dynamic key errors

## Fixed cubic's finding (index alignment)

cubic correctly flagged that `queuedMessages` filtered out empty (attachment-only) drafts before rendering, breaking index alignment with `queuedDrafts` so the remove action could target the wrong entry. Fixed in 20164db6 by mapping one label per draft (no filtering), with an attachment-count fallback when there's no text. Removed an unused i18n key too.

PR #2011, opened 2026-05-30
feat(artifacts): show scrollable artifact list in side panel

monrheo:
## Summary
- Replaces the "no artifacts" empty state with a scrollable list of all artifacts
- Clicking an artifact opens it in the panel; closing returns to the list
- Adds `text` to `ARTIFACT_FILE_PREVIEWS` so plain-text files (`.ts`, `.js`, `.json`, etc.) appear

## Before
Clicking Artifacts (N) opened only the first artifact as a tab. The other N-1 were invisible.

## After
Clicking Artifacts (N) opens a scrollable list. Each artifact shows icon, name, and size.


apps/app/src/react-app/domains/session/panel/side-panel.tsx
```diff
@@ -455,6 +458,52 @@ export function SidePanel({
   );
 }
 
+function ArtifactList({ sessionId }: { sessionId: string }) {
+  const openTab = usePanelTabStore((state) => state.openTab);
+  const targets = usePanelTabStore((state) =>
```
Dhanus3133, line 463: Current filtering code causes an infinite render loop which crashes the UI. We need to select the raw state and wrap the filter in a  useMemo  in the component to avoid returning a new array reference on every selector run. You can use this:
```diff
diff --git a/apps/app/src/react-app/domains/session/panel/side-panel.tsx b/apps/app/src/react-app/domains/session/panel/side-panel.tsx
index d436efff..01e72ee5 100644
--- a/apps/app/src/react-app/domains/session/panel/side-panel.tsx
+++ b/apps/app/src/react-app/domains/session/panel/side-panel.tsx
@@ -460,9 +460,10 @@ export function SidePanel({
 
 function ArtifactList({ sessionId }: { sessionId: string }) {
   const openTab = usePanelTabStore((state) => state.openTab);
-  const targets = usePanelTabStore((state) =>
-    (state.transcriptArtifactTargets[sessionId] ?? []).filter(isCollectibleArtifactTarget)
-  );
+  const rawTargets = usePanelTabStore((state) => state.transcriptArtifactTargets[sessionId]);
+  const targets = React.useMemo(() => {
+    return (rawTargets ?? []).filter(isCollectibleArtifactTarget);
+  }, [rawTargets]);
 
   if (targets.length === 0) {
     return (
```

Dhanus3133 (commented):
I see this
Image

PR #2069, opened 2026-06-02
fix(app): attach first-line local paths

src-opn:
## Summary
- attach local file paths found on the first composer line as OpenCode file parts
- keep original prompt text while avoiding provider/model errors from raw path-only context
- restrict slash command parsing so absolute paths like /Users/... are not treated as commands

## Tests
- pnpm --dir "apps/app" exec bun test tests/prompt-file-parts.test.ts — passed
- pnpm --dir "apps/app" typecheck — passed

## Evidence
- No video captured for this small parser/send-path fix. Focused automated coverage verifies the exact `check ~/code/research/openwork-users/list.csv` style input and slash-command regression.


src-opn:
Addressed the issue identified by cubic: first-line auto-attachment now detects Windows absolute paths in both `C:\...` and `C:/...` forms.

Verification:
- `pnpm --dir "apps/app" exec bun test tests/prompt-file-parts.test.ts`
- `pnpm --dir "apps/app" typecheck`

PR #2106, opened 2026-06-09
fix: render artifacts inline

src-opn:
## Summary
- render artifact chips inline with their source messages instead of as an aggregated bottom row
- compact artifact chips into one horizontally scrollable row with smaller icons and truncated titles
- hide open affordances for unsupported/unverified artifacts and sort artifacts newest-first

## Tests
- pnpm --dir apps/app test:open-target
- bun test apps/app/scripts/artifacts.test.ts
- pnpm --dir apps/app typecheck

## Evidence
- No video/screenshot captured; validated with focused tests and typecheck.

### DIFF
--- a/apps/app/scripts/artifacts.test.ts
+++ b/apps/app/scripts/artifacts.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from "bun:test";
 import type { UIMessage } from "ai";
 
 import type { OpenTarget } from "../src/react-app/domains/session/artifacts/open-target";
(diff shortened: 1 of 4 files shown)


src-opn:
Addressed the cubic review comments:
- latest-message artifacts now render inline as message artifacts instead of being suppressed
- assistant text parsing now accepts any known artifact type, including slides like .pptx

Re-ran:
- pnpm --dir apps/app test:open-target
- bun test apps/app/scripts/artifacts.test.ts
- pnpm --dir apps/app typecheck

PR #2109, opened 2026-06-09
fix(workspace): stop swallowing server errors when connecting workspaces

benjaminshafii:
## Problem

All six workspace-creation handlers (local + remote, across session/settings/welcome routes) swallowed every error from the OpenWork server:

```ts
list = await client.createRemoteWorkspace(payload).catch(() => null);
if (!list) throw new Error("OpenWork server is unavailable. Start or reconnect the server ...");
```

The server returns precise, actionable 400s (e.g. `openwork_workspace_not_found`: "OpenWork server has no workspace matching <url>", invalid baseUrl, bad token), and the client already throws `OpenworkServerError` carrying that message — but the UI converted **every** failure, including a typo'd URL or wrong token, into "server is unavailable", pointing users at the wrong fix.

## Fix

New helper `toOpenworkConnectionError(error, unavailableMessage)` in `apps/app/src/app/lib/openwork-server.ts`:
- `OpenworkServerError` (server replied with a real error) → preserve the server's message.
- Anything else (network down, timeout) → keep the existing "server unavailable" guidance.

Applied to all six call sites:
- `session-route.tsx` — `handleCreateWorkspace`, `handleCreateRemoteWorkspace`
- `settings-route.tsx` — `handleCreateWorkspace`, `handleCreateRemoteWorkspace`
- `welcome-route.tsx` — `handleCreateWorkspaceFolder`, `handleCreateRemote`

No behavior change for genuine transport failures.

## Testing

- Reviewer repro: in the create-workspace modal choose "Remote custom", enter a reachable server URL with a wrong token or a URL the server rejects. Before: "OpenWork server is unavailable...". After: the server's actual message (e.g. "OpenWork server has no workspace matching ...").

Part of the provisioning-flow fix series (2/6). Builds on #2108.

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -794,6 +794,15 @@ export class OpenworkServerError extends Error {
   }
 }
 
+/**
+ * Preserve actionable server-side error messages (e.g. "no workspace matching
+ * <url>", invalid token) while mapping transport failures (network down,
+ * timeout) to a generic "server unavailable" message.
+ */
+export function toOpenworkConnectionError(error: unknown, unavailableMessage: string): Error {
+  return error instanceof OpenworkServerError ? error : new Error(unavailableMessage);
+}
+
 function buildHeaders(
   token?: string,
   hostToken?: string,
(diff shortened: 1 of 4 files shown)


benjaminshafii:

PR #2111, opened 2026-06-09
feat(workspace): pre-flight connection check before saving a remote worker

benjaminshafii:
## Problem

The "Connect remote" form persisted a workspace entry on any non-empty URL string. No URL validation, no reachability check, no token check — even though a full diagnostics utility (`testRemoteWorkspaceConnection`) already existed for *post-hoc* troubleshooting. A typo'd URL or wrong token created a broken sidebar entry the user only discovered later through connection errors. When the URL embeds a workspace mount, the server skips discovery entirely and persists a dead workspace without any error.

## Fix

New `preflightRemoteWorkspaceConnection({ hostUrl, token })` in `remote-workspace-diagnostics.ts`, reusing the module's existing normalization/health/capabilities helpers. Unlike the full diagnostics it doesn't require a token (the form allows token-less servers) or an existing workspace entry. Checks:

1. URL is a valid http(s) endpoint → "Enter a valid worker URL ..."
2. `GET /health` reachable + healthy → "Cannot reach <host> ..."
3. If a token was entered, `GET /capabilities` doesn't 401/403 → "Token was rejected by <host> ..."

`CreateWorkspaceModal` (shared by session, settings, and welcome routes) runs the pre-flight on submit and shows the failure in the existing error slot; nothing is persisted until it passes. The submit button/cancel reuse the existing "Connecting..." busy state during the check.

## Testing

- Reviewer repro: create workspace -> "Remote custom":
  - Enter `not-a-url` → inline "Enter a valid worker URL..." (previously: submitted and failed downstream).
  - Enter an unreachable https URL → "Cannot reach <host>..." with no workspace entry created.
  - Enter a valid worker URL + wrong token → "Token was rejected by <host>...".
  - Valid URL + token → connects as before.

Part of the provisioning-flow fix series (4/6). Related: #2108, #2109, #2110.

### DIFF
--- a/apps/app/src/react-app/domains/workspace/create-workspace-modal.tsx
+++ b/apps/app/src/react-app/domains/workspace/create-workspace-modal.tsx
@@ -4,6 +4,7 @@ import {
   useMemo,
   useReducer,
   useRef,
+  useState,
   type SetStateAction,
 } from "react";
 import { ArrowLeft, FolderPlus, Globe, Loader2 } from "lucide-react";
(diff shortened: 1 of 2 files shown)


benjaminshafii:

PR #2112, opened 2026-06-09
fix(den-api): persist provisioning failure reason and sweep stuck provisioning workers

benjaminshafii:
## Problem

Two dead-end states in cloud worker provisioning:

1. **Permanent `provisioning`**: `POST /v1/workers` runs provisioning as a fire-and-forget promise (`void continueCloudProvisioning(...)`). If the API process restarts or crashes mid-provision, the worker row stays `provisioning` forever — no reconciliation, no timeout. The desktop renders it as "Starting" with Open disabled, permanently.
2. **Discarded failure reason**: when provisioning fails, the worker is set to `failed` and the actual error only goes to `console.error`. Users see an "Attention" badge with zero context.

## Fix

- **Schema**: new nullable `worker.status_reason` varchar(1024) column (`den-db` migration `0021_fancy_gravity.sql`, additive only).
- **Persist failure reason**: `continueCloudProvisioning` now stores the provisioning error message in `status_reason` on failure and clears it on success. Returned as `statusReason` in all worker API responses (additive OpenAPI field).
- **Read-time sweep**: `failStuckProvisioningWorkers(orgId)` marks org workers stuck in `provisioning` for >30 min as `failed` with an actionable reason. Called from `GET /v1/workers` and `GET /v1/workers/:id` — deployment-agnostic (no background timers, works in serverless), and the desktop polls the list while provisioning so stuck rows resolve automatically. 30 min comfortably exceeds the longest legit provisioning path (Render: 15 min provision + 3 min healthcheck).
- Exported `lt` from den-db's drizzle re-exports.

## Testing

- `tsc --noEmit` in `ee/apps/den-api` — passes (exit 0).
- `pnpm --filter @openwork-ee/den-db build` — passes; `db:generate` produced a clean additive migration.
- Could not run a live MySQL + Daytona provisioning loop here. Reviewer repro: run local Den stack (`pnpm dev:den`), create a cloud worker with `PROVISIONER_MODE=stub` killed mid-flight or backdate a `provisioning` row's `created_at` by >30 min, then `GET /v1/workers` → row flips to `failed` with `statusReason`.

Part of the provisioning-flow fix series (5/6). Surfacing `statusReason` in the desktop UI lands in the next PR. Related: #2108, #2109, #2110, #2111.

### DIFF
--- a/ee/apps/den-api/src/routes/workers/core.ts
+++ b/ee/apps/den-api/src/routes/workers/core.ts
@@ -14,6 +14,7 @@ import {
   continueCloudProvisioning,
   createWorkerSchema,
   deleteWorkerCascade,
+  failStuckProvisioningWorkers,
   getLatestWorkerInstance,
   getWorkerByIdForOrg,
   getWorkerTokensAndConnect,
@@ -45,6 +46,7 @@ const workerSchema = z.object({
   description: z.string().nullable(),
   destination: z.string(),
   status: z.string(),
+  statusReason: z.string().nullable(),
   imageVersion: z.string().nullable(),
   workspacePath: z.string().nullable(),
   sandboxBackend: z.string().nullable(),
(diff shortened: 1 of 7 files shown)


benjaminshafii:

PR #2113, opened 2026-06-09
fix(cloud): show worker provisioning failure reason and fix status mapping

benjaminshafii:
## Problem

1. When cloud worker provisioning fails, the desktop shows only an "Attention" badge — no explanation, no way for the user to know what went wrong. (den-api now persists the reason as `statusReason` — #2112.)
2. `workerStatusMeta` had a vocabulary mismatch with den-api: the default branch allowed opening workers with status `"ready"` — a status den-api never emits (`provisioning | healthy | failed` only) — while any genuinely unknown status fell through inconsistently.

## Fix

- `den.ts`: `DenWorkerSummary` now carries `statusReason` (parsed defensively from `/v1/workers`; older servers without the field simply yield `null`).
- `sections.tsx`:
  - `CloudWorkerListItem` renders the failure reason under the worker row (red text) when the status tone is error.
  - `workerStatusMeta`: `"ready"` is now an explicit alias of `healthy` (openable, green); the default branch for unknown statuses no longer allows opening.

Backward compatible: servers that don't send `statusReason` render exactly as before.

## Testing

- Reviewer repro: with #2112 deployed, force a provisioning failure (e.g. bad Daytona credentials) → Settings -> Cloud -> Workers shows the "Attention" worker with the actual provisioning error underneath instead of an unexplained badge.

Part of the provisioning-flow fix series (6/6). Related: #2108, #2109, #2110, #2111, #2112.

### DIFF
--- a/apps/app/src/app/lib/den.ts
+++ b/apps/app/src/app/lib/den.ts
@@ -97,6 +97,7 @@ export type DenWorkerSummary = {
   workerId: string;
   workerName: string;
   status: string;
+  statusReason: string | null;
   instanceUrl: string | null;
   provider: string | null;
   isMine: boolean;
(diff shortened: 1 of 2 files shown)


benjaminshafii:

PR #2149, opened 2026-06-10
feat(den-web): Analytics page in the Den org admin dashboard

benjaminshafii:
## What

Second PR of the **OpenWork Analytics** stack (Layers 1+2). Adds an admin-only **Analytics** page to the Den dashboard, consuming the `GET /v1/telemetry/analytics` endpoint from #2148.

> Stacked on #2148 (base branch is `analytics/layer12-server`). Retarget to `dev` once #2148 merges.

- New route `(admin)/analytics` + `AnalyticsScreen` (`ee/apps/den-web/app/(den)/dashboard/_components/analytics-screen.tsx`)
- Sidebar nav entry "Analytics" (admin-only group, same gating as Members/Billing) + page title
- **Layer 1** — summary cards: OpenWork users (+ pending invites), active this week (+ 30d)
- **Layer 2** — sessions this week, tasks this week + success rate; 12-week trend charts for weekly active users, sessions per week, tasks per week (completed vs failed); 30-day detail cards (avg task duration, tasks completed/failed)
- Dependency-free bar charts (pure Tailwind divs), empty-state overlay when no events exist yet, privacy note (event metadata only — never prompts/code/files)

## Tests

Commands run:

- `npx tsc --noEmit` in `ee/apps/den-web` — passes
- `pnpm --filter @openwork-ee/den-web build` — production build passes, `/dashboard/analytics` route emitted
- Manual end-to-end against the local Den stack (MySQL + den-api from #2148 + seeded `acme-robotics-demo` org with 12 weeks of telemetry events): signed in as the org owner via headless Chrome (CDP), navigated to `/dashboard/analytics`, verified all cards and charts render real aggregates (17 users, 14 active this week, 44 sessions, 40 tasks at 93% success, growing 12-week trends, 3m 22s avg duration). Screenshots captured (attaching below).

Non-admin access is blocked by the existing `(admin)` route-group layout guard.

## Screenshots

_Screenshots of the page rendering live seeded data are attached as a comment (captured via headless Chrome against the local stack)._

PR #2148, opened 2026-06-10
feat(den): analytics event model + /v1/telemetry/analytics endpoint (Layer 1+2)

benjaminshafii:
## What

First of three PRs implementing **OpenWork Analytics** Layers 1 (who is using AI) and 2 (how often).

Server-side foundation:

- **Event model** (`ee/packages/den-db/src/schema/telemetry.ts` + migration `0021`): adds `source`, `session_id`, `duration_ms`, `success` columns to `telemetry_event`, and expands the allowed event types to the PRD set: `user.active`, `session.active`, `session.started/ended`, `task.started/completed/failed`.
- **Ingest hardening** (`POST /v1/telemetry/ingest`): unknown event types are silently dropped, disallowed `source` values are nulled, `durationMs` is clamped. Still fire-and-forget (always 204, errors swallowed). No content fields exist anywhere in the pipeline — only event names, timestamps, opaque session ids, durations, and success flags.
- **New `GET /v1/telemetry/analytics`**: org-scoped Layer 1+2 aggregates — members, pending invites, active members (7d/30d), distinct sessions (7d/30d), tasks completed/failed (7d/30d), avg task duration (30d), and a 12-week trend of active members / sessions / tasks per week.

The existing `/v1/telemetry/adoption` endpoint is unchanged for compatibility.

## Stack

1. **This PR** — server event model + analytics endpoint
2. #TBD — Den dashboard Analytics page (stacked on this)
3. #TBD — desktop app emits task lifecycle events

## Tests

Commands run:

- `pnpm --filter @openwork-ee/den-db db:generate` — migration `0021_flaky_hex.sql` generated cleanly
- `npx tsc --noEmit` in `ee/apps/den-api` — passes
- End-to-end against the local Den stack (`dev:den:mysql` + drizzle push + `dev:den:api` + `seed:demo-org`):
  - signed in as the seeded demo owner, `POST /v1/telemetry/ingest` with 6 events → 204
  - unknown event type (`totally.unknown`) **not stored** (5 rows in DB), disallowed source (`evil-source`) stored as NULL
  - `GET /v1/telemetry/analytics` returned correct aggregates: `members: 17`, `activeMembers7d: 1`, `sessions7d: 2` (distinct session ids), `tasksCompleted7d: 2`, `tasksFailed7d: 1`, `avgTaskDurationMs30d: 30000` = (42000+18000)/2 ✓

Raw verification output:

```
ingest: 204
[...]

PR #2150, opened 2026-06-10
feat(app): emit task lifecycle telemetry events to Den (Analytics Layer 2)

benjaminshafii:
## What

Third PR of the **OpenWork Analytics** stack (with #2148 server and #2149 dashboard). The desktop/web app now emits real Layer 2 usage signals to `POST /v1/telemetry/ingest`:

- `apps/app/src/app/lib/den-telemetry.ts`: events carry `source: "app"`, optional `sessionId`, `durationMs`, `success`; new helpers `trackTaskStarted/Completed/Failed`
- Send path (`session-route.tsx`): `task.started` on each send; `session.active` now carries the session id so the server can count **distinct sessions per week**
- Run lifecycle (`session-sync.ts`): `task.completed` / `task.failed` with run duration on `session.idle` / `session.error`, reusing the existing `markTaskRunStart`/`takeTaskRunStart` bookkeeping (so it only fires for runs this client instrumented, deduped across workspace syncs)

Privacy unchanged: auth-gated (no-op when signed out of Den), fire-and-forget, batched; only event names, opaque session ids, durations, and success flags — never prompts, code, or file paths.

**Compatibility:** safe to merge independently of #2148 — the current server's zod ingest schema strips unknown fields, so older servers simply ignore the new fields; with #2148 deployed they're stored and aggregated.

## Tests

Commands run:

- `pnpm typecheck` (`@openwork/app`) — passes
- `bun test` in `apps/app` — 109 pass / 4 fail; the 4 failures (`artifact-spreadsheet`, `artifacts`) are **pre-existing on the base commit** (verified by running the same tests on an untouched `origin/dev` worktree — same failures)
- The exact event payload shape produced here (`task.started`/`task.completed`/`task.failed` with `sessionId`/`durationMs`/`success`) was verified end-to-end against the new server in #2148: ingested via `POST /v1/telemetry/ingest` (204), stored, and correctly aggregated by `GET /v1/telemetry/analytics` (distinct sessions, success rates, avg duration)

No video for this PR: the change is invisible in the UI (background telemetry). To reproduce manually: run the local Den stack (`pnpm dev:den`) with #2148, sign in to Den in the app, send a prompt, and watch rows appear in `telemetry_event` and the numbers move on the #2149 Analytics page.

### DIFF
--- a/apps/app/src/app/lib/den-telemetry.ts
+++ b/apps/app/src/app/lib/den-telemetry.ts
@@ -16,9 +16,16 @@ import { type DenSettings, readDenSettings, resolveDenBaseUrls } from "./den";
 const INGEST_PATH = "/v1/telemetry/ingest";
 const INGEST_TIMEOUT_MS = 5_000;
 
-type TelemetryEvent = {
+type TelemetryEventFields = {
+  sessionId?: string;
+  durationMs?: number;
+  success?: boolean;
+};
+
+type TelemetryEvent = TelemetryEventFields & {
   type: string;
   timestamp: string;
+  source: "app";
 };
 
 let pendingEvents: TelemetryEvent[] = [];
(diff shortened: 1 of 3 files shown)


benjaminshafii:
Rebased onto `dev` — applies cleanly over the session-route decomposition rounds (#2200–#2204); verified the `trackTaskStarted` call still sits in the live send choke point and `trackTaskCompleted/Failed` next to the existing run-lifecycle analytics in session-sync.

**Re-tested:** `pnpm typecheck` clean; `bun test` in apps/app: 138 pass / 2 fail — both failures (`artifact-spreadsheet`) are pre-existing on untouched `origin/dev` (verified).

benjaminshafii:
Note: the `openwork-tests (macos-14)` failure is the pre-existing flake `reload event API > exposes runtime config content changes as reload events` — it fails intermittently on `dev` itself (e.g. runs 27450298191 / 27448739832 for unrelated CI-only commits) and passes locally on macOS (224 pass / 0 fail in apps/server) and on the ubuntu runner. This PR touches only `apps/app`.

PR #1977, opened 2026-05-27
feat(app): add new message list

JeanMeijer:
## Summary
- Adds a new chat message list implementation with shared message/provider utilities.
- Adds reusable markdown, artifact, tool-call, reasoning, source, steps, and response-stream UI components.
- Adds renderers for built-in tools like bash, file, grep, glob, web search/fetch, apply patch, questions, skills, and todos.
- Wires the session surface/sync adapter to the new message rendering path.
- Adds artifact/tool helper libraries and updates app dependencies/lockfile for the new UI stack.

PR #2165, opened 2026-06-10
fix(app): make message branch/revert actually work and add right-click edit message

benjaminshafii:
## Problem

Three message-action bugs, all reproduced end-to-end in the real Electron app on a Daytona sandbox (opencode 1.16.2, GPT-5.5, multi-turn session with file edits):

1. **Branching produced a "weird message".** "Branch in new chat" on an assistant turn created a session that dropped the final answer and ended mid-turn on a tool step followed by a literal **"Empty message"** carcass. Cause: OpenCode's `session.fork` copies messages strictly *before* the given id, and the UI passed the id of the clicked message itself (or worse, a synthetic `session-error:*` id that doesn't exist server-side, which corrupts the boundary and copies everything).
2. **Revert didn't work at all.** Clicking Revert succeeded server-side (verified `session.revert` was set via the opencode API) but the UI never changed. Cause: nothing invalidated the snapshot cache or truncated the transcript cache, so the revert cursor never reached the renderer. After a manual reload there was *also* an off-by-one: `applyRevertCursor` kept the reverted message visible (`slice(0, idx + 1)`), while OpenCode treats `revert.messageID` as the *first reverted* message.
3. **No way to edit a message.** Right-click on a message did nothing; there was no edit/rewire feature at all.

## Fix

- **Branch includes the branch point**: `resolveForkBoundaryId` passes the *next real* message id to `session.fork` (or forks the whole session at the tail), skipping synthetic `session-error:*` ids. Group actions target the last real message of the turn.
- **Revert updates the UI immediately**: after a successful `session.revert`, the returned revert cursor is stamped into the snapshot cache, the live transcript cache is truncated, and the snapshot refetches (`applySessionRevert`). Fixed the cursor off-by-one (`slice(0, idx)`).
- **Caches stay consistent with the server**: `session.updated` SSE events now sync the snapshot cache's revert cursor (set *and* clear), and `message.removed` events prune both transcript and snapshot caches. Without this, sending a prompt after a revert left a stale cursor that froze the transcript until reload (caught during e2e validation). `seedSessionState` also applies the snapshot's revert cursor so reverted messages can't be resurrected by merges.
- **New: Edit message**: right-click a user message (or use the hover pencil) → context menu with **Edit message / Copy / Branch in new chat / Revert**. Edit reverts the session to just before that prompt and restores the text into the composer so it can be rewritten and resent.
- Revert/branch failures now surface as toasts instead of silent `console.warn`.

## Tests run

- `cd apps/app && bun test scripts/session-render-state.test.ts tests/session-sync-permissions.test.ts` → **37 pass, 0 fail** (includes 13 new assertions for `applyRevertCursor`, `resolveForkBoundaryId`, and revert-aware `deriveRenderedSessionMessages`)
- `cd apps/app && tsc -p tsconfig.json --noEmit` → clean
- Note: `tests/session-sync-tool-parts.test.ts` fails on `dev` independently of this PR (it imports `shouldDeferInProgressTool`, which `parse-tool-parts.ts` never exports — drift from #1977).

## End-to-end validation (Daytona, real Electron via CDP)

Sandbox: `bash .devcontainer/test-on-daytona.sh`, real OpenAI GPT-5.5 session in `/workspace/hello`, driven via CDP with message-id-level assertions.

**Before (dev @ b471f7d9a):**
- Branch at assistant turn → forked session ended `[user, assistant "TURN ONE OK", user, assistant "Apply patch"]` + **"Empty message"**, final answer missing.
- Revert on a user message → server `session.revert` set, **UI unchanged** (all 8 messages still rendered); after manual reload the reverted prompt was still visible (off-by-one).
- Right-click on messages → no menu (0 menus opened).

**After (this branch, same sandbox, same session):**
- Revert on the "TURN FOUR" prompt → transcript went **7 → 5 messages instantly**, no reload.
- Right-click user message → menu `["Edit message", "Copy", "Branch in new chat", "Revert"]`; **Edit message** rewound the transcript to turn 1 and filled the composer with the original prompt text; resending produced a fresh turn that rendered live (validates the post-revert prompt path).
- Branch at assistant turn → new session contains the **complete history including the final "TURN TWO OK" answer**, no "Empty message".
- Branch at a user message → branch now includes the clicked prompt (previously silently dropped).

Before/after screenshots captured during the runs are available on request (branch-empty-message before, full-turn branch + edit-composer after).

## How to reproduce manually

1. Create a workspace, run 2–3 prompts (one that edits a file).
2. Hover a user message → Revert: the turn should disappear immediately.
3. Right-click a user message → Edit message: transcript rewinds, composer refills; edit and resend.
4. Hover an assistant answer → Branch in new chat: the new session must end with that answer (no "Empty message").

### DIFF
--- a/apps/app/scripts/session-render-state.test.ts
+++ b/apps/app/scripts/session-render-state.test.ts
@@ -6,7 +6,11 @@ import {
   deriveRenderedSessionMessages,
   resolveRenderedSessionSnapshot,
 } from "../src/react-app/domains/session/surface/session-render-state";
-import { reconcileTranscriptMessages } from "../src/react-app/domains/session/sync/transcript-reconcile";
+import {
+  applyRevertCursor,
+  reconcileTranscriptMessages,
+  resolveForkBoundaryId,
+} from "../src/react-app/domains/session/sync/transcript-reconcile";
 import { describeOpencodeSessionError } from "../src/react-app/domains/session/sync/usechat-adapter";
 
 function snapshotWithMessages(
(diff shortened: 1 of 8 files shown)


benjaminshafii:
## E2E evidence (Daytona, real Electron + opencode 1.16.2 + GPT-5.5)

**Screen recording of the fixed flows** (~106s, recorded on the sandbox while driving the real app):
https://8090-xs4o8qogluioq6m0.daytonaproxy01.net/recordings/message-fixes-after.mp4

What the recording shows, in order:
1. A real session (turn 1 + turn 2 with an `Apply patch` file edit).
2. **Right-click a user message** → context menu (`Edit message / Copy / Branch in new chat / Revert`).
3. **Edit message** → transcript rewinds instantly to turn 1, composer refills with the original prompt.
4. **Resend** → new turn streams in and completes (validates prompting after revert — previously the transcript froze on a stale revert cursor).
5. **Branch in new chat** on the assistant answer → branched session contains the complete history **including the final answer** — no "Empty message" carcass.

CDP assertions captured during validation (message-id level):
- BEFORE (dev @ b471f7d9a): revert left all 8 messages rendered while the server's `session.revert` was confirmed set via the opencode API; branch at an assistant turn produced a session ending on a tool step + "Empty message"; right-click opened 0 menus.
- AFTER (this branch): revert went 7 → 5 messages instantly with no reload; branch produced `[user, "TURN ONE OK", user, "Apply patch", "TURN TWO OK"]` with `hasEmptyMessage: false`; the context menu rendered all 4 items; edit rewound to 2 messages and filled the composer with the exact original prompt text.

Note: the recording URL is served from a temporary Daytona sandbox (`openwork-test-20260610-145719`) and dies when the sandbox is deleted. A local copy of the recording plus before/after screenshots is preserved at `/tmp/message-fixes-evidence/` and can be re-attached on request.

PR #1890, opened 2026-05-22
fix(den-api): build on Windows

pascalandr:
## Summary

Updates `ee/apps/den-api/scripts/build.mjs` for Windows-compatible Den API builds.

## Verification

- `pnpm install --frozen-lockfile` — passed.
- `pnpm --filter @openwork-ee/den-api build` — passed after frozen install restored missing workspace dependencies.
- `git diff --name-status upstream/dev...pr/den-api-windows-build` reviewed; one intended file only.

PR #2172, opened 2026-06-11
fix(den-api): build on Windows

pascalandr:
Replaces #1890.

Summary:
- Makes the Den API build script Windows-safe.
- Covers pnpm spawn input construction with a targeted test.
- Aligns generated latest app version to 0.16.0.

Verification:
- pnpm --dir "ee/apps/den-api" exec bun test --timeout 30000 test/build-script.test.ts
- pnpm --dir "ee/apps/den-api" exec tsc -p tsconfig.json --noEmit


Pablosinyores (commented):
Solid refactor. Guarding `main()` behind `import.meta.url === pathToFileURL(process.argv[1]).href` is the right way to make `build.mjs` importable from the test without firing the build, and going through `pathToFileURL` rather than a raw string compare is exactly what keeps that working on Windows drive paths. Preferring `npm_execpath` via `node <npmExecPath> ...args` with `shell: false` is also the most robust branch, and it is the one that will actually run under `pnpm run build`.

One nit on the win32 shell fallback: `quoteShellArg` escapes an embedded quote as `\"`, which is the MSVCRT/POSIX convention rather than cmd.exe's, which doubles them (`""`). It is moot for the current callers since every arg is a static token with no quotes, but if an arg ever carried a literal quote the `shell: true` path would mis-parse it. Worth a short comment noting the assumption, or doubling quotes in the win32 branch to be safe.

PR #2184, opened 2026-06-11
fix(mcp): stop enabled MCPs from reverting to off after engine disposes

benjaminshafii:
## Problem

Users report MCP toggles — including **OpenWork UI control** — keep turning off, and MCP installs failing (see also user report on X). Traced it: SQLite runtime state never flips to `false`; the **engine's view** reverts.

## Root causes & fixes

1. **Direct dispose with no re-sync** — provider auth flows (`refreshProviders({dispose:true})`) called `instance.dispose()` directly on the opencode engine. The engine rebuilds from the `OPENCODE_CONFIG_CONTENT` snapshot frozen at spawn, dropping every runtime-DB MCP enabled since. Now routed through the OpenWork server's `POST /workspace/:id/engine/reload` (which re-registers runtime MCPs), with fallback to direct dispose when no server is reachable.
2. **Sync aborted on first failure** — `syncRuntimeMcpToOpencodeEngine` threw on the first non-OK entry, so one dead remote MCP (e.g. an unreachable `posthog`/`openwork-cloud`) silently blocked re-registration of everything after it — including `openwork-ui` — on every reload. It now continues past per-entry failures and aggregates errors. This also explains "can't install any MCPs" when one configured MCP is down.
3. **Spawn injection only covered `workspaces[0]`** — runtime-DB MCPs for other workspaces were invisible until a manual reload. `embedded.ts`/`cli.ts` now push every workspace's runtime MCPs into the engine at startup (best-effort, exported `syncAllWorkspacesRuntimeMcpToEngine`).
4. **Renderer fallback hid runtime-DB MCPs** — when the OpenWork server was briefly unreachable (startup race), the file-only fallback dropped `config.remote` entries, rendering enabled MCPs as off. Last-known runtime entries are now kept in the fallback.

## Cloud safety

- Den/cloud servers run the same `startServer`; the reload route + sync fix apply there too.
- Startup sync only fires when a managed engine was spawned (`managedOpencode` non-null), so cloud servers pointing at external engines are unaffected.
- Provider-auth reload falls back to direct dispose when the OpenWork server isn't connected.

## Tests

- `bun test src/mcp.engine-sync.e2e.test.ts` — 7 pass (2 new: partial-failure reload sync, multi-workspace startup sync)
- `bun test` (apps/server) — 406 pass / 7 fail; the same 7 failures pre-exist on clean `dev` (verified via `git stash`)
- `pnpm typecheck` (apps/app) — clean
- `pnpm typecheck` (apps/server) — 1 pre-existing error on `dev` (`SessionMessagesResponse2` SDK export, unrelated)
- `bun test scripts/desktop-cloud-sync.test.ts` (apps/app) — 8 pass

No video captured: the trigger (engine dispose during provider auth) is exercised by the new e2e tests against a mock engine. To reproduce manually: enable OpenWork UI control, then connect/disconnect any model provider — before this fix the MCP drops from the engine.

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -285,6 +285,12 @@ export type OpenworkMcpItem = {
   disabledByTools?: boolean;
 };
 
+export type OpenworkMcpEngineSync = {
+  status: "ok" | "failed";
+  at: number;
+  failures: Array<{ name: string; status?: number; message?: string }>;
+};
+
 export type OpenworkWorkspaceExport = {
   workspaceId: string;
   exportedAt: number;
@@ -1369,7 +1375,11 @@ export function createOpenworkServerClient(options: { baseUrl: string; token?: s
         },
       ),
     listMcp: (workspaceId: string) =>
-      requestJson<{ items: OpenworkMcpItem[] }>(baseUrl, `/workspace/${workspaceId}/mcp`, { token, hostToken }),
+      requestJson<{ items: OpenworkMcpItem[]; engineSync?: OpenworkMcpEngineSync | null }>(
+        baseUrl,
+        `/workspace/${workspaceId}/mcp`,
+        { token, hostToken },
+      ),
     addMcp: (workspaceId: string, payload: { name: string; config: Record<string, unknown> }) =>
       requestJson<{ items: OpenworkMcpItem[] }>(baseUrl, `/workspace/${workspaceId}/mcp`, {
         token,
--- a/apps/app/src/react-app/domains/connections/provider-auth/store.ts
+++ b/apps/app/src/react-app/domains/connections/provider-auth/store.ts
@@ -1168,10 +1168,34 @@ export function createProviderAuthStore(options: CreateProviderAuthStoreOptions)
     if (!c) return null;
 
     if (optionsArg?.dispose) {
+      // Prefer the OpenWork server engine reload: it disposes the engine AND
+      // re-registers runtime-DB MCPs. A direct dispose makes the engine
+      // rebuild from the OPENCODE_CONFIG_CONTENT snapshot frozen at spawn,
+      // silently dropping any MCPs enabled since (toggles "turn off").
+      let reloaded = false;
       try {
-        unwrap(await c.instance.dispose());
+        const openworkSnapshot = options.openworkServer.getSnapshot();
+        const openworkClient = openworkSnapshot.openworkServerClient;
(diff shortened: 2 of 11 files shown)


benjaminshafii:
## Round 2: structural fixes (items 2 + 3 from the hardening plan)

### Server-managed `OPENCODE_CONFIG` file (replaces frozen `OPENCODE_CONFIG_CONTENT`)
The engine re-reads the `OPENCODE_CONFIG` file from disk on every instance rebuild (verified in opencode source: `loadInstanceState` → `loadFile(Flag.OPENCODE_CONFIG)`, merged on top of global config). OpenWork now writes the runtime config to `<config-dir>/runtime-opencode-config.json` (atomic temp+rename, serialized writes) and rewrites it on **every runtime-DB write** via a store write-listener. Result: even engine disposes OpenWork doesn't control pick up current MCP/plugin state instead of reverting to the spawn snapshot. This also means plugin changes load on the next instance rebuild instead of requiring a full app restart.

Precedence note: `OPENCODE_CONFIG` sits below project `opencode.jsonc` (CONTENT sat above). Name collisions between project-config MCPs and runtime-DB MCPs resolve to project config at instance build; the dynamic `/mcp` re-sync still applies runtime values afterwards, same as before.

### Engine sync failures are no longer silent
- Per-entry retry (once, on 5xx/network errors — engine is often mid-rebuild after a dispose)
- Last sync outcome recorded per workspace, exposed as `engineSync` on `GET /workspace/:id/mcp`, and logged
- The Extensions UI now shows *which* MCPs failed to register and suggests an engine reload

### Evidence
**Real-engine smoke test** (opencode 1.16.2, the exact bug scenario — direct dispose bypassing the OpenWork server, which used to wipe MCP state):
```
add MCP: 200
config file has smoke-mcp: true (.../runtime-opencode-config.json)
direct engine dispose: 200
engine MCPs after dispose: exa, smoke-mcp
PASS: smoke-mcp survived an untracked engine dispose via the fresh config file
```
(`exa` is the user's global-config MCP — confirms global config still merges and is not clobbered.)

**Tests run:**
- `bun test src/mcp.engine-sync.e2e.test.ts src/openwork-runtime-config.test.ts` — all pass (3 new tests: file contents, rewrite-on-write, workspace isolation; plus engineSync surfacing + retry assertions)
- `bun test` (apps/server) — 409 pass / 7 fail, same 7 pre-existing on `dev`
- `pnpm typecheck` (apps/app) — clean; (apps/server) — only the pre-existing `SessionMessagesResponse2` error

PR #2185, opened 2026-06-11
feat(extensions): Claude Code plugin bundle compatibility (MCP + skills + commands from GitHub)

benjaminshafii:
> Stacked on #2184 (uses its engine-sync changes). Retarget to `dev` after that merges.

## What

Makes OpenWork compatible with the **Claude Code plugin bundle format** — the way Anthropic ships integrations like Slack by Salesforce: one repo with `.claude-plugin/plugin.json` bundling an MCP server (connectivity) + skills (expertise) + commands (guided first-run recipes). Includes an Anthropic-style **"Will install" disclosure** before anything is written.

## How

**Server** — `claude-plugin-bundle.ts` resolves a GitHub repo (`owner/repo`, full URLs, `/tree/<ref>/<subdir>`, `.git` suffix) into the existing `CloudPluginResolved` shape. Install therefore reuses `installCloudPlugin` wholesale: namespacing, Claude frontmatter translation (tools/model/agent), MCP registration into the runtime DB, the install registry, and uninstall via the existing `DELETE /cloud-plugins/:pluginId`.

- `POST /workspace/:id/claude-plugins` — `{ url, ref?, dryRun? }`. `dryRun` returns the preview (components grouped by type + warnings) with zero writes; install goes through approval gating, audit, reload events, and hot-registers bundled MCPs with the running engine.
- Graceful degradation with explicit warnings instead of silent failure: `hooks` (unsupported), MCP servers using `\${CLAUDE_PLUGIN_ROOT}` (plugin-local binaries), multi-file skill extras (only SKILL.md installed for now).
- Mock-able GitHub endpoints via `OPENWORK_GITHUB_API_BASE` / `OPENWORK_GITHUB_RAW_BASE` for tests.

**App** — "From GitHub" button on the Extensions custom-app card opens the disclosure modal: paste URL → Preview → plugin name/version/source + grouped "Will install" list (MCP servers / Skills / Commands / Agents, with descriptions) + warnings → Install. Installed bundles show up as extension cards with the existing uninstall path.

## Evidence

**Real-world install of the official Slack plugin** (`bun` smoke against live GitHub):
```
preview status: 200  → name: slack, v1.0.0
components: 1 MCP (slack), 2 skills (slack-messaging, slack-search),
            5 commands (channel-digest, draft-announcement, find-discussions, standup, summarize-channel)
install status: 200
installed: .opencode/skills/slack-plugin/{slack-messaging,slack-search}/SKILL.md
           .opencode/commands/slack-plugin/{5 commands}.md
           opencode.jsonc#mcp.slack
workspace MCPs: exa (config.global), slack (config.remote)
```

## Tests

- `bun test src/claude-plugin-bundle.e2e.test.ts` — 3 pass (URL parsing variants; dryRun preview incl. \${CLAUDE_PLUGIN_ROOT} skip warning; full install/uninstall cycle against a mock GitHub + mock engine, asserting written files, runtime-DB MCP registration, engine push, and cleanup)
- `bun test` (apps/server) — 412 pass / 7 fail (same 7 pre-existing on `dev`)
- `pnpm typecheck` — app clean; server has only the pre-existing `SessionMessagesResponse2` error

No video captured — the end-to-end flow is covered by the live-GitHub smoke above and the e2e tests; to reproduce in-app: Settings → Extensions → "From GitHub" → paste `https://github.com/slackapi/slack-mcp-plugin` → Preview → Install.

## Follow-ups (out of scope)

- Multi-file skills (scripts/, references/) need directory-aware install records
- `userConfig` prompts at enable time (sensitive values → keychain)
- Surfacing bundle commands as composer example prompts
- Marketplace repos with `marketplace.json` listing multiple plugins

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -258,6 +258,22 @@ export type OpenworkCloudPluginsResult = {
   plugins: Record<string, CloudImportedPlugin>;
 };
 
+export type OpenworkClaudePluginComponent = {
+  type: "mcp" | "skill" | "command" | "agent";
+  name: string;
+  description: string | null;
+};
+
+export type OpenworkClaudePluginPreview = {
+  pluginId: string;
+  name: string;
+  description: string | null;
+  version: string | null;
+  source: { owner: string; repo: string; ref: string; dir: string | null };
+  components: OpenworkClaudePluginComponent[];
+  warnings: string[];
+};
+
 function arrayBufferToBase64(data: ArrayBuffer): string {
   const bytes = new Uint8Array(data);
   let binary = "";
@@ -1259,6 +1275,22 @@ export function createOpenworkServerClient(options: { baseUrl: string; token?: s
         token,
         hostToken,
         method: "DELETE",
+        timeoutMs: timeouts.config,
+      }),
+    previewClaudePlugin: (workspaceId: string, payload: { url: string; ref?: string }) =>
+      requestJson<{ preview: OpenworkClaudePluginPreview }>(baseUrl, `/workspace/${encodeURIComponent(workspaceId)}/claude-plugins`, {
+        token,
+        hostToken,
+        method: "POST",
+        body: { ...payload, dryRun: true },
+        timeoutMs: timeouts.config,
+      }),
+    installClaudePlugin: (workspaceId: string, payload: { url: string; ref?: string }) =>
+      requestJson<OpenworkCloudPluginInstallResult & { preview: OpenworkClaudePluginPreview }>(baseUrl, `/workspace/${encodeURIComponent(workspaceId)}/claude-plugins`, {
+        token,
+        hostToken,
+        method: "POST",
+        body: payload,
         timeoutMs: timeouts.config,
       }),
     readOpencodeConfigFile: (workspaceId: string, scope: "project" | "global" = "project") => {
--- a/apps/app/src/react-app/domains/connections/modals/claude-plugin-import-modal.tsx
+++ b/apps/app/src/react-app/domains/connections/modals/claude-plugin-import-modal.tsx
@@ -1,0 +1,228 @@
+/** @jsxImportSource react */
+import { useReducer } from "react";
+import { Download, Loader2, Search } from "lucide-react";
+
+import {
+  Dialog,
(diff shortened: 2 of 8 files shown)


benjaminshafii:
Addressed the cubic review in 75bc25ef9:

- **P1 stuck installing state**: `handleInstall` now wraps `onInstall` in try/catch — thrown errors render in the modal instead of locking it (close is blocked while installing).
- **P1 stale preview race**: the preview is now bound to the URL it was generated from (`previewedUrl`); install always targets that URL, and preview responses that resolve after the user edited the field are discarded.
- **P2 query/hash URLs**: `parseClaudePluginSource` strips `?…`/`#…` before parsing, with new test cases (`?tab=readme-ov-file#install`, tree URL + query).
- **P2 refs containing `/`** (e.g. `/tree/release/v1.0/dir`): deferred — disambiguating ref vs subdir requires listing repo refs via the GitHub API. Documented as a known limitation in the parser docstring; will revisit if someone hits it.

Verification: `bun test src/claude-plugin-bundle.e2e.test.ts` → 3 pass / 0 fail (26 assertions); `tsc --noEmit` clean for both `@openwork/app` and `apps/server`.

benjaminshafii:
Addressed all 4 cubic findings (latest two commits):

1. **P1 — stuck installing state**: `onInstall` is wrapped in try/catch; thrown errors render in the modal instead of locking it (close is blocked while installing).
2. **P1 — stale preview race**: previews are bound to the URL they were generated from and discarded if the input changed while in flight; **Install always targets the previewed URL**, never the live input value.
3. **P2 — slash branch refs**: actually fixed rather than documented — `/tree/<...>` URLs are now resolved by trying progressively longer refs against the trees API (`release` → `release/v1` → …) and using the first that exists; raw URLs keep ref slashes as path segments. Covered by a new e2e test with a `release/v1` branch mock.
4. **P2 — query/hash in URLs**: stripped before parsing (`?tab=readme-ov-file#readme` now parses), with test coverage.

Re-verified: `bun test src/claude-plugin-bundle.e2e.test.ts` — 4 pass; full server suite 413 pass / 7 pre-existing fails; both typechecks clean; live-GitHub smoke against slackapi/slack-mcp-plugin still previews + installs (slack MCP registered as `config.remote`).

PR #2226, opened 2026-06-13
feat(app): sync session groups server-side

src-opn:
## Summary
- Store session groups and assignments server-side in the runtime SQLite DB
- Add session group APIs plus update events so clients can sync group state
- Add client-side migration from existing local group state and show "Create a Group" next to "Show x more" when no groups exist

## Tests
- pnpm --filter openwork-server test src/session-groups.e2e.test.ts
- pnpm --filter openwork-server typecheck
- Manual desktop smoke: ran pnpm dev from the worktree with Electron CDP on http://127.0.0.1:9830; verified migration PUT, session-groups/events polling, "Show 2 more ⋅ Create a Group", creating a Test Group, and server persistence/events in the UI

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -500,6 +500,27 @@ export type OpenworkReloadEvent = {
   timestamp: number;
 };
 
+export type OpenworkSessionGroupDefinition = {
+  id: string;
+  label: string;
+};
+
+export type OpenworkSessionGroupState = {
+  groups: OpenworkSessionGroupDefinition[];
+  assignments: Record<string, string>;
+};
+
+export type OpenworkSessionGroupEvent = {
+  id: string;
+  seq: number;
+  workspaceId: string;
+  type: "session_groups.updated";
+  action: "created" | "updated" | "deleted" | "assigned" | "reordered" | "imported";
+  groupId?: string;
+  sessionId?: string;
+  timestamp: number;
+};
+
 // Fallback for explicit server-mode URL derivation. Desktop local workers replace this
 // with the persisted runtime-discovered port once the host reports it.
 export const DEFAULT_OPENWORK_SERVER_PORT = 8787;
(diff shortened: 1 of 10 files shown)


src-opn:
Addressed the four issues identified by cubic in b5bc18c4:

- Added per-workspace client sync versions so stale session-group responses cannot overwrite newer optimistic/server state.
- Made session-group event polling single-flight to avoid overlapping refresh loops.
- Serialized server read-modify-write session-group updates per workspace to avoid lost concurrent writes.
- Changed the session-group event buffer to be per-workspace so busy workspaces cannot evict quiet workspace events.

Validation:
- pnpm --filter openwork-server test src/session-groups.e2e.test.ts
- pnpm --filter openwork-server typecheck

src-opn:
Addressed the new issue identified by cubic in f8c316c8:

- Reworked client mutation response handling so successful older server responses are retained while newer mutations are pending.
- If a newer mutation fails, the latest successful pending response is applied once all pending mutations settle, preventing UI/server divergence while still avoiding stale overwrites of newer optimistic state.
- Full background/poll syncs now skip applying while local mutations are pending.

Validation:
- pnpm --filter openwork-server test src/session-groups.e2e.test.ts
- pnpm --filter openwork-server typecheck

src-opn:
Addressed the latest two issues identified by cubic in 0538d5fc:

- Added a separate server-sync freshness version so out-of-order getSessionGroups responses cannot apply stale data.
- Buffered the latest server sync result while local mutations are pending, then apply it if all pending mutations settle without a newer successful mutation response.
- Successful mutation responses still take precedence over deferred server sync data.

Validation:
- pnpm --filter openwork-server test src/session-groups.e2e.test.ts
- pnpm --filter openwork-server typecheck

src-opn:
Addressed the latest issue identified by cubic in 89d5ccfe:

- Deferred server sync state is now applied after pending mutations finish even if a newer sync has since started but not applied yet.
- The freshness guard still prevents stale sync responses from becoming deferred in the first place; this only prevents dropping the last valid deferred state during mutation settlement.

Validation:
- pnpm --filter openwork-server test src/session-groups.e2e.test.ts
- pnpm --filter openwork-server typecheck

PR #2229, opened 2026-06-13
fix(app): hide default-only agent picker

src-opn:
## Summary
- hide the composer agent picker when only the default OpenWork agent is available
- filter the default OpenWork agent out of agent picker lists so non-default agents are the only selectable alternatives
- refresh agent availability after engine reloads so newly added agents reveal the picker

## Tests

## Evidence
- Not captured (code-only conditional UI change).

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src-opn:
Addressed the issue identified by cubic: when the agent picker should no longer show, the composer now closes `agentMenuOpen` so the hidden picker cannot continue capturing keyboard shortcuts.

Verification: `pnpm --filter @openwork/app typecheck`

PR #2281, opened 2026-06-15
feat(composer): use a friendlier label for the built-in build agent

### LINKED ISSUE #1065: Use friendlier agent labels in the composer
## Context
Enterprise readiness polish follow-up from the Mar 19 enterprise review.

## Problem
The remaining composer issue is mostly naming clarity, not a broad controls redesign.

## Scope
- rename agent choices to friendlier labels
- specifically replace `Build` with `General purpose agent`
- keep the composer understandable without raw agent jargon

## Out of scope
- `show configured model first` (handled elsewhere)
- simplified thinking controls (handled elsewhere)
- review / run-task affordance cleanup (handled elsewhere)
- the broader criteria work from PR #1086 (not continuing in this issue)

## Acceptance criteria
- agent labels are friendly and self-explanatory
- non-technical users can pick the general-purpose option without needing agent vocabulary

sanidhyasin:
## Problem

Closes #1065.

The composer agent picker rendered each agent's **raw name**, so OpenCode's built-in `build` agent showed up as **"Build"** in the picker and in the selected-agent pill. As called out in the issue, that's developer jargon — a non-technical user can't tell that "Build" is the general-purpose option.

## Change

- Add `formatComposerAgentLabel(name)` (`.../composer/agent-label.ts`): maps known built-in agents to friendly, **translatable** labels and falls back to the existing capitalized name for every other (custom) agent. Today the only mapping is `build → "General purpose agent"`, matching the issue's explicit direction.
- Add the `composer.agent_general_purpose` key to `en.ts`. Other locales resolve it through the existing locale → English fallback, so no other locale files need to change.
- Wire the helper into both agent-picker menus (`composer.tsx`) and the selected-agent pill (`session-route.tsx`), replacing the inline `name.charAt(0).toUpperCase() + name.slice(1)`.

The `@`-mention list is intentionally left alone — its label doubles as the value the engine matches on, so it must stay the raw agent name.

## Scope

Minimal and behavior-preserving for every agent except `build`: custom agents and `plan` still render exactly as before.

## Testing

- `bun test tests/agent-label.test.ts` — added unit test for the helper (built-in relabel + capitalization fallback). **2 pass.**
- I did **not** run the full `pnpm typecheck`/build or capture a UI video: this is a fresh clone without the monorepo `node_modules` installed, and a full install was avoided as too heavy for a one-line-helper change. The edits are type-preserving (`string → string`) substitutions.

### Reviewer repro (UI)
1. Connect any provider and open a session.
2. Open the agent picker in the composer.
3. The `build` entry now reads **"General purpose agent"** instead of "Build"; selecting it shows the same friendly label in the agent pill. Other agents are unchanged.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/i18n/locales/en.ts
+++ b/apps/app/src/i18n/locales/en.ts
@@ -77,6 +77,7 @@ export default {
   "common.something_went_wrong": "Something went wrong",
   "common.submit": "Submit",
   "common.unknown": "Unknown",
+  "composer.agent_general_purpose": "General purpose agent",
   "composer.agent_label": "Agent",
   "composer.agent_selected": "Agent set to {agent}.",
   "composer.agent_selected_default": "Agent reset to default.",
--- a/apps/app/src/i18n/locales/ca.ts
+++ b/apps/app/src/i18n/locales/ca.ts
@@ -75,6 +75,7 @@ export default {
   "common.something_went_wrong": "Hi ha hagut un problema",
   "common.submit": "Envia",
   "common.unknown": "Desconegut",
+  "composer.agent_general_purpose": "Agent de propòsit general",
   "composer.agent_label": "Agent",
   "composer.any_file_type_supported": "S'admet qualsevol tipus de fitxer.",
   "composer.attach_files": "Adjuntar fitxers",
--- a/apps/app/src/i18n/locales/es.ts
+++ b/apps/app/src/i18n/locales/es.ts
@@ -75,6 +75,7 @@ export default {
   "common.something_went_wrong": "Algo salió mal",
   "common.submit": "Enviar",
   "common.unknown": "Desconocido",
+  "composer.agent_general_purpose": "Agente de propósito general",
   "composer.agent_label": "Agente",
   "composer.any_file_type_supported": "Se admite cualquier tipo de archivo.",
   "composer.attach_files": "Adjuntar archivos",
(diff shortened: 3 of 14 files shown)


apps/app/src/i18n/locales/en.ts
```diff
@@ -77,6 +77,7 @@ export default {
   "common.something_went_wrong": "Something went wrong",
   "common.submit": "Submit",
   "common.unknown": "Unknown",
+  "composer.agent_general_purpose": "General purpose agent",
```
Dhanus3133, line 80: Other language translations are missing. Anyway, It will safely fall back to english for other languages for now, but just adding this in case translations need to be added. 

Image

Dhanus3133 (commented):
Tested locally and the changes works!

sanidhyasin:
Good catch, thanks @Dhanus3133! Added translations for all locales in b04eefd.

PR #2293, opened 2026-06-16
fix(settings): align authorized folders empty state

### LINKED ISSUE #1101: [Bug]: Authorized folders empty state renders with broken light overlay in dark settings
**Title:** [Bug]: Authorized folders empty state renders with broken light overlay in dark settings

**Summary:** *
In `Settings > General`, the `Authorized folders` section renders with a large light/gray overlay that looks visually broken in the dark theme. The empty state appears disabled or covered, and the input/footer styling does not match the rest of the settings UI.

**Steps to reproduce:** *
1. Open OpenWork.
2. Go to `Settings`.
3. Open the `General` tab.
4. Scroll to the `Authorized folders` section when no external folders are authorized.

**Expected behavior:** *
The empty state should match the dark settings theme, with readable text and controls that look active and intentionally styled.

**Actual behavior:** *
The section shows a large light/gray panel inside the dark UI, making the area look disabled or visually broken. The empty-state text has low contrast, and the input/footer row looks detached from the card styling.

---
Label: bug

Evidence: screenshot provided in chat for this report, but not attached via CLI.

Iflaqbhat:
## Summary
- Fixes #1101
- Restyles the Authorized folders empty state to match the Settings dark theme
- Uses OpenWork settings DLS surface, border, and text tokens instead of the generic Empty defaults

## Verification
- pnpm --filter @openwork/app test
- git diff --check

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/react-app/domains/settings/panels/authorized-folders-panel.tsx
+++ b/apps/app/src/react-app/domains/settings/panels/authorized-folders-panel.tsx
@@ -154,10 +154,11 @@ export function AuthorizedFoldersPanel(props: AuthorizedFoldersPanelProps) {
   const canPickAuthorizedFolder =
     isDesktopRuntime() && canWriteConfig && props.activeWorkspaceType === "local";
   const workspaceRootFolder = serverWorkspaceRoot || props.selectedWorkspaceRoot.trim();
-  const visibleAuthorizedFolders = useMemo(() => {
+  const externalAuthorizedFolders = useMemo(() => {
     const root = workspaceRootFolder;
-    return root ? [root, ...authorizedFolders.filter((folder) => folder !== root)] : authorizedFolders;
+    return root ? authorizedFolders.filter((folder) => folder !== root) : authorizedFolders;
   }, [authorizedFolders, workspaceRootFolder]);
+  const hasVisibleFolderRows = Boolean(workspaceRootFolder) || externalAuthorizedFolders.length > 0;
 
   useEffect(() => {
     const openworkClient = props.openworkServerClient;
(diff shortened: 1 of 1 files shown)


sid-valecha (commented):
I checked this locally at `aa040c4c`.

What I could verify:
- `git diff --check dev...HEAD` -> pass
- The patch is limited to `apps/app/src/react-app/domains/settings/panels/authorized-folders-panel.tsx`
- The new empty-state styling uses the existing `Empty variant="ghost"` path and DLS tokens, so the code-level change looks consistent with the settings surface.

One thing I would not call fully verified yet: this PR says `Fixes #1101`, but the block being restyled only renders when `visibleAuthorizedFolders.length === 0`. In current code, `visibleAuthorizedFolders` prepends `workspaceRootFolder` whenever the server/selected workspace root is known, so the normal “no external folders authorized” state should still render the workspace-root row rather than this `Empty` block. That matches the earlier issue comment where the broken overlay could not be reproduced on `dev` without a more specific state.

Could you attach a screenshot or the exact app state where this `Empty` block renders after the change, for example no runtime workspace root / empty selected workspace root? Without that, the fallback restyle looks reasonable, but I would not treat the #1101 repro as fully validated yet.

Iflaqbhat:
@sid-valecha Updated in 6eb552e6.

Thanks, you were right. I updated the implementation so the empty state is now reachable in the normal no-external-folders state.

The workspace root row is rendered separately, and external authorized folders are tracked separately. When there are no external folders, the DLS-styled empty state renders below the workspace root row.

Verified locally in Settings > Permissions with a workspace root and no external folders: the root row remains visible and the empty state renders naturally without demo flags.

Checks:
- `git diff --check`
Screenshot 2026-06-16 at 9 58 28 PM

Pablosinyores:
The split between the always-on workspace-root row and the external-folders list reads much cleaner, and scoping the empty state to `externalAuthorizedFolders.length === 0` with the `no_external_folders` copy is a sensible reframing.

Just confirming the intended UX: when a workspace root is set but there are no external folders, `hasVisibleFolderRows` is true (root row renders) and the empty block also renders since `externalAuthorizedFolders.length === 0` — so both show together. That looks deliberate given the external-specific copy, but worth a quick visual check that the dashed empty block sits naturally beneath the root row rather than reading like a half-empty panel. @Iflaqbhat

Iflaqbhat:
@Pablosinyores Thanks, yes, that is the intended UX.

I verified visually in Settings > Permissions with a workspace root and no external folders. The workspace root row stays visible, and the external-folders empty state renders below it. The dashed empty block reads as a separate external-folder state rather than a half-empty root panel.

The copy is external-specific: “No external folders authorized”, so showing both together feels correct.

PR #2294, opened 2026-06-16
feat: add support for custom openai compatible provider (fixes #2102)

### LINKED ISSUE #2102: Support Nous provider / custom OpenAI-compatible endpoint
## Summary\r\n\r\nOpenWork currently supports a fixed set of providers (OpenRouter, OpenAI, Anthropic, Google, NVIDIA). For users who want to use Nous Research's OpenAI-compatible inference API () with their own API key, there is no way to configure it because the app does not expose a custom  or generic OpenAI-compatible provider entry.\r\n\r\n## Request\r\n\r\n- Add support for a **custom OpenAI-compatible provider** where users can set:\r\n  - \r\n  - \r\n  - model id(s)\r\n- Or add **Nous** as a first-class provider option using the Nous inference endpoint.\r\n\r\n## Context\r\n\r\n- Hermes Agent already uses Nous successfully with OpenAI-compatible clients.\r\n- Many smaller providers expose an OpenAI-compatible API; supporting one custom endpoint makes OpenWork useful beyond the currently bundled providers.

ashishSoni1234:
## Summary
- Added support for a custom OpenAI-compatible provider. Users can now input a Base URL, Model Name, and API Key to connect any OpenAI-compatible API endpoint via the AI Providers settings UI.

## Why
- To address the need for using custom endpoints and API keys, specifically allowing users to connect to Nous Research's OpenAI-compatible inference API or other smaller providers.

## Issue
- Closes #2102

## Scope
- Modified `store.ts` to handle custom provider storage in `opencode.jsonc`.
- Updated `provider-auth-modal.tsx` to display the "Custom OpenAI Compatible" option.
- Hooked up routing changes in `session-route.tsx` and `settings-route.tsx`.

## Out of scope
- N/A

### Ran
- `pnpm run dev`

### Result
- pass/fail: pass

## CI status
- pass: Yes (Assumed pending CI run)
- code-related failures: N/A

## Manual verification
1. Opened the AI Providers section in Settings.
2. Clicked on "Custom OpenAI Compatible".
3. Entered Base URL, Model Name, and API key.
4. Saved and verified the connection successfully.

## Evidence
- N/A

## Risk
- Low. Only adds a new provider option without modifying existing provider logic.

## Rollback
- Simply revert the added provider ID and related UI options. No breaking changes introduced.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/react-app/domains/connections/provider-auth/provider-auth-modal.tsx
+++ b/apps/app/src/react-app/domains/connections/provider-auth/provider-auth-modal.tsx
@@ -57,6 +57,7 @@ const PROVIDER_LABELS: Record<string, string> = {
 };
 
 const OPENWORK_MODELS_PROVIDER_ID = "openwork";
+const CUSTOM_OPENAI_PROVIDER_ID = "__custom_openai_compatible__";
 
 export type ProviderAuthModalProps = {
   open: boolean;
@@ -79,6 +80,11 @@ export type ProviderAuthModalProps = {
   onRefreshProviders?: () => Promise<unknown>;
   showOpenWorkModelsSubscribe?: boolean;
   onSubscribeOpenWorkModels?: () => void | Promise<void>;
+  onSubmitCustomProvider?: (opts: {
+    baseUrl: string;
+    modelId: string;
+    apiKey: string;
+  }) => Promise<string | void>;
   onClose: () => void;
 };
 
@@ -87,7 +93,7 @@ export default function ProviderAuthModal(props: ProviderAuthModalProps) {
   const isRemoteWorker = workerType === "remote";
 
   const [view, setView] = useState<
-    "list" | "method" | "api" | "cloud" | "oauth-code" | "oauth-auto" | "openwork-subscribe"
+    "list" | "method" | "api" | "cloud" | "oauth-code" | "oauth-auto" | "openwork-subscribe" | "custom"
   >("list");
   const [selectedProviderId, setSelectedProviderId] = useState<string | null>(null);
   const [selectedCloudMethod, setSelectedCloudMethod] = useState<ProviderAuthMethod | null>(null);
@@ -101,6 +107,10 @@ export default function ProviderAuthModal(props: ProviderAuthModalProps) {
   const [oauthAutoBusy, setOauthAutoBusy] = useState(false);
   const [oauthCodeCopied, setOauthCodeCopied] = useState(false);
   const [oauthBrowserOpened, setOauthBrowserOpened] = useState(false);
+  // Custom OpenAI-Compatible provider form state
+  const [customBaseUrl, setCustomBaseUrl] = useState("");
+  const [customModelId, setCustomModelId] = useState("");
+  const [customApiKey, setCustomApiKey] = useState("");
 
   const searchInputRef = useRef<HTMLInputElement | null>(null);
   const providerPollRef = useRef<number | null>(null);
@@ -196,6 +206,14 @@ export default function ProviderAuthModal(props: ProviderAuthModalProps) {
       })
       .sort(compareProviders);
 
+    const customEntry: ProviderAuthEntry = {
+      id: CUSTOM_OPENAI_PROVIDER_ID,
+      name: "Custom OpenAI Compatible",
+      methods: [{ type: "api", label: "Configure" }],
+      connected: false,
+      env: [],
+    };
+
     if (props.showOpenWorkModelsSubscribe) {
       const connectedToOpenWork = connected.has(OPENWORK_MODELS_PROVIDER_ID);
       return [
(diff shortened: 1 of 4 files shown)


ashishSoni1234:
Hi @SongotenU, I have implemented the custom OpenAI-compatible provider as requested in issue #2102. The PR is ready for review. Could you please take a look and authorize the Vercel CI workflows whenever you have time? Let me know if any changes are needed. Thanks!

Pablosinyores:
Useful feature. A few questions from the modal flow:
- API key storage: does `onSubmitCustomProvider` persist the key through the same secure store the built-in providers use, or does it land in plaintext config? It's a user-supplied secret, so good to confirm it gets the same treatment.
- `modelId` is optional in `handleCustomSubmit` (only URL + key required). If left blank, what does the request send as the model — a sensible default, or does the first completion just fail? Might be worth requiring it or defaulting explicitly.
- `new URL(trimmedUrl)` accepts any scheme, so `ftp://`/`javascript:` pass. A light `https?:` check keeps it to real endpoints.

View wiring and form reset on close look clean.

ashishSoni1234:
Hi @obchain and @Pablosinyores, thanks for the review! I've pushed a commit addressing all the feedback:

'Model Name' is now properly validated as a required field.
The 'Base URL' strictly requires http:// or https:// protocol now.
Passed a fallbackUpdate down to the config update logic so the custom provider correctly persists for remote workers.
Checked on the API key storage, it uses the existing secure credential store logic (c.auth.set), so it's safely tucked away and not exposed in the workspace config file.

Please let me know if anything else is needed or if this is good to merge!

Abhijeet1005:
Pulled the latest (`12512226`) on Windows and traced through it — the commit resolves all the points raised above. Sharing what I checked in case it helps move it along.

**Verified**
- **API key storage:** the key is persisted via `c.auth.set({ providerID, auth: { type: "api", key } })` — the same secure credential path the built-in single-key providers use (the existing `c.auth.set` calls at `store.ts:1346`/`1394`). The block written to `opencode.jsonc` is only `{ id, name, api, npm, models }` — no key in plaintext config.
- **Model Name required:** validated in both the modal (`handleCustomSubmit`) and the store (`submitCustomProvider`), label is now `*`, and Save is disabled without it. `modelEntry` is therefore always populated, so the "blank model → zero selectable models" path can't happen anymore.
- **Remote workers:** `submitCustomProvider` now passes a `fallbackUpdate` to `updateProjectConfigFile`. When `readProjectConfigFile()` returns null it no longer bails — it falls through to `c.config.get()` → `fallbackUpdate(config)` → `c.config.update()`, and the fallback preserves existing config via spreads.
- **URL scheme:** `new URL()` now rejects anything but `http:`/`https:` in both the modal and the store, so `ftp://`/`javascript:` are out. Allowing `http://` (not just https) is the right call for local inference servers (Ollama/LM Studio on `http://localhost`).

`pnpm --filter @openwork/app typecheck` → clean. Couldn't exercise the modal live (needs the full app), but the logic and wiring check out. Looks mergeable to me.

One tiny non-blocking aside: the new validation messages are hardcoded English, but that matches every other error string in this modal, so it's consistent — a future i18n pass could lift them all together.

Pablosinyores (commented):
Nice addition — the UX flow and validation are solid (URL parsed + http/https enforced, required-field guards, key masked, keys-stored-locally note). Two things worth a look on the persistence side in `store.ts`:

**1. `providerId` derived from the hostname slug can collide and silently overwrite (medium).** `submitCustomProvider` builds `providerId = custom-<hostname-slug>` and then does `modify(updated, ["provider", providerId], providerConfig, …)`, which *replaces* the whole provider block. Two consequences:
- Two different endpoints on the same host — e.g. `https://api.foo.com/v1` and `https://api.foo.com/v2`, or `api.foo.com` vs `inference-api.foo.com` (both slug → `foo`) — resolve to the same `providerId`, so the second save clobbers the first (api URL + models + stored key).
- Even for the *same* endpoint, `models` is set to a single `{ [modelId]: … }`, so adding a second model on that provider overwrites the first model rather than merging. Adding two models from one custom host loses the first.

If multiple custom models/endpoints are meant to be supported, consider merging into the existing `provider[providerId].models` block, and/or disambiguating the id when a different `api` URL already exists under the derived id.

**2. Minor:** the URL validation is duplicated in `handleCustomSubmit` (modal) and `submitCustomProvider` (store) — fine as defense-in-depth, just noting it so the two don't drift. Allowing `http` is reasonable here (local/LAN endpoints).

Non-blocking overall; (1) is the one I'd confirm is intended before merge.

PR #2300, opened 2026-06-17
fix(composer): expand pasted text from chip label

### LINKED ISSUE #1843: [Feature]: The pasted text will be folded and cannot be expanded and adjusted.
### Summary

The pasted text will be folded and cannot be expanded and adjusted.

### Problem / goal

I hope that like previous versions, I can click on the label of pasted text and modify the text after unfolding it. This will be more convenient.

### Primary user(s)

- [ ] Bob (IT / power user)
- [ ] Susan (non-technical)
- [ ] Other team roles

No

No

### Testability

Click on the label where the text is pasted to see if it can expand

No

### Additional context

_No response_

Iflaqbhat:
Fixes #1843

## Summary
- Makes the pasted-text chip itself clickable, not only the small arrow icon
- Keeps the existing expand affordance while allowing users to click the `Pasted · n lines` label
- Adds keyboard support so focused pasted-text chips expand with Enter or Space

## Verification
- Verified locally by pasting multi-line text into the composer and clicking the pasted-text label to expand it
- `git diff --check`

https://github.com/user-attachments/assets/e8a967e0-6e5f-4ef3-aecf-487a9692a1fe

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
+++ b/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
@@ -302,20 +302,21 @@ function pastedTextChipLabel(lines: number) {
 
 function createPastedTextChipDom(label: string, lines: number) {
   const dom = document.createElement("span");
-  dom.className = "inline-flex items-center gap-1 rounded-full border border-amber-6/35 bg-amber-3/15 px-2.5 py-1 text-xs font-medium text-amber-11";
+  dom.className = "group inline-flex cursor-pointer items-center gap-1 rounded-full border border-amber-6/35 bg-amber-3/15 px-2.5 py-1 text-xs font-medium text-amber-11 transition-colors hover:bg-amber-4/20 focus-visible:outline focus-visible:outline-2 focus-visible:outline-amber-8";
   dom.contentEditable = "false";
   dom.setAttribute("spellcheck", "false");
+  dom.setAttribute("role", "button");
+  dom.setAttribute("aria-label", `Expand pasted text ${label}`);
+  dom.tabIndex = 0;
   dom.title = `Pasted text · ${label}`;
+  dom.dataset.pastedExpandLabel = label;
 
   const text = document.createElement("span");
   text.textContent = pastedTextChipLabel(lines);
 
-  const button = document.createElement("button");
-  button.type = "button";
-  button.className = "ml-0.5 inline-flex h-4 w-4 items-center justify-center rounded-full text-amber-10 transition-colors hover:bg-amber-4 hover:text-amber-12";
-  button.title = "Expand pasted text";
-  button.setAttribute("aria-label", "Expand pasted text");
-  button.dataset.pastedExpandLabel = label;
+  const icon = document.createElement("span");
+  icon.className = "ml-0.5 inline-flex h-4 w-4 items-center justify-center rounded-full text-amber-10 transition-colors group-hover:text-amber-12";
+  icon.setAttribute("aria-hidden", "true");
 
   const svg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
   svg.setAttribute("viewBox", "0 0 16 16");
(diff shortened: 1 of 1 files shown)


Pablosinyores:
The refactor to `pastedTextExpandLabelFromTarget` and moving the affordance onto the whole chip reads well — icon `aria-hidden` with the label on the `role="button"` host is the right a11y shape.

Two things in the editor context:
- The chip now has `tabIndex={0}` inside the Lexical `contentEditable`. Does Tab now stop on every pasted-text chip while navigating the composer? In a draft with several pastes that could get noisy / pull focus out of the typing flow — worth checking how it feels.
- For `role="button"`, Space normally activates on keyup (keydown just prevents scroll). Handling both on keydown with `preventDefault` stops the scroll so it's fine in practice — just flagging for strict button semantics.

Click/keyboard expand path otherwise looks consistent.

Iflaqbhat:
@Pablosinyores Thanks, that makes sense.

For `tabIndex={0}`, I kept the chip keyboard-focusable because the whole chip is now the interactive expand control. Removing it would make the chip nicer for Tab flow, but worse for keyboard accessibility because mouse users could expand it while keyboard users could not.

The Space key point is fair. Since we prevent default scrolling, the current keydown behavior works in practice, but I agree it is slightly different from native button semantics. I can adjust Space to activate on keyup if maintainers prefer that.

I’ll do one more local keyboard pass with multiple pasted chips and can tweak if the Tab flow feels too noisy.

Abhijeet1005:
Pulled this on Windows and traced through it — clean a11y/UX fix, looks correct.

**What I checked**
- The refactor makes the whole chip one interactive control (`role="button"`, `tabIndex=0`, `aria-label`, focus-visible ring) and turns the old inner `<button>` into an `aria-hidden` decorative icon. Good call — one clear keyboard target, no nested-interactive element.
- The thing I most wanted to confirm: a new capture-phase key handler in the composer is exactly where you'd risk breaking Enter-to-send. It doesn't — `handlePastedTextExpandKeyDown` only acts when `event.target` is inside `[data-pasted-expand-label]` (a focused chip). While typing, the target is the contentEditable root, so it returns early without `preventDefault`, and Enter-to-send / Shift+Enter / spaces are untouched. Only a Tab-focused chip responds to Enter/Space — which is the intent. Same scoping holds for the click/mousedown capture handlers.
- `pnpm --filter @openwork/app typecheck` -> clean.

I couldn't exercise the click/focus behavior live (that needs the full app running), so this is a static + typecheck pass — but the logic is sound and matches the recording. LGTM.

Iflaqbhat:
@Abhijeet1005 Thanks for the feedback and for checking this on Windows!

Pablosinyores (commented):
Nice a11y upgrade — making the whole chip the target (role=button + aria-label + keyboard) instead of the tiny arrow is a real usability win, and folding the pointer/keydown detection into `pastedTextExpandLabelFromTarget` reads well. Since it's an ARIA `role="button"` on a span (not a native button), the explicit Enter/Space keydown handler is exactly what's needed — no synthetic click to worry about, so no double-fire.

Two small non-blockers:
- Native buttons fire on Space at *keyup* to avoid auto-repeat; keydown + preventDefault here is fine inside the editor, just noting the convention.
- `tabIndex = 0` now inserts the chip into the editor's Tab order. If that's intended for keyboard reach, good; if you only wanted pointer + Enter from focus, worth a second look so Tab-through of a long draft doesn't stop on every chip.

PR #2302, opened 2026-06-17
fix(desktop): recover empty workspace state

src-opn:
## Summary
- recover desktop workspace state at startup when openwork-workspaces.json has an empty workspaces array
- prefer persisted OpenWork server config, then fall back to openwork-server-tokens.json path recovery
- add Electron workspace-store coverage for both recovery paths

## Tests
- pnpm --filter @openwork/desktop test ✅
- pnpm --filter @openwork/desktop typecheck:electron ✅

## Evidence
- No video captured; this is a startup migration/state recovery path verified with focused Electron tests.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/desktop/electron/workspace-store.mjs
+++ b/apps/desktop/electron/workspace-store.mjs
@@ -511,7 +511,7 @@ export function createWorkspaceStore({ app, defaultDenBaseUrl, defaultRequireSig
 
   async function readWorkspaceState() {
     const state = await readJsonFile(workspaceStatePath(), EMPTY_WORKSPACE_LIST);
-    const selectedId =
+    let selectedId =
       typeof state?.selectedId === "string"
         ? state.selectedId
         : typeof state?.selectedWorkspaceId === "string"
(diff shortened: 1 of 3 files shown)


src-opn:
Addressed the issue identified by cubic: recovered workspaces now flow through the existing remote OpenWork migration/dedup path instead of returning early. Added coverage for recovered duplicate legacy remote entries normalizing to one rem_<workspaceId> entry.\n\nTests run:\n- pnpm --filter @openwork/desktop test ✅\n- pnpm --filter @openwork/desktop typecheck:electron ✅

PR #2309, opened 2026-06-18
fix(desktop): include mac x64 node-pty package

### LINKED ISSUE #2190: OpenWork app 0.16.2 is not working on mac with intel chip
I am trying to install 0.16.2 version mac (intel) - openwork-mac-x64-0.16.2.dmg

After installation when i am launching app, i am getting below error

----
Uncaught Exception:
Error: The @lydell/node-pty package supports your platform (darwin-x64), but it could not find the platform-specific package for it: @lydell/node-pty-darwin-x64

This can happen if you use the "--omit=optional" (or "--no-optional") npm flag.
The "optionalDependencies" package.json feature is used to install the correct
binary executable for your current platform. Remove that flag to use @lydell/node-pty.

This can also happen if the "node_modules" folder was copied between two operating systems
that need different binaries - including "virtual" operating systems like Docker and WSL.
If so, try installing with npm rather than copying "node_modules".
at requirePlatformSpecificPackage (/Applications/OpenWork.app/Contents/Resources/app.asar/node_modules/node-pty/index.js:19:13)
at Object.<anonymous> (/Applications/OpenWork.app/Contents/Resources/app.asar/node_modules/node-pty/index.js:47:18)
at Module._compile (node:internal/modules/cjs/loader:1738:14)
at Module._extensions..js (node:internal/modules/cjs/loader:1904:10)
at Module.load (node:internal/modules/cjs/loader:1472:32)
at Module._load (node:internal/modules/cjs/loader:1289:12)
at c._load (node:electron/js2c/node_init:2:17950)
at TracingChannel.traceSync (node:diagnostics_channel:322:14)
at wrapModuleLoad (node:internal/modules/cjs/loader:242:24)
at Module.require (node:internal/modules/cjs/loader:1494:12)
----

Iflaqbhat:
Fixes #2190

## Summary
- Adds pnpm supportedArchitectures so macOS Electron release installs include both arm64 and x64 native optional packages
- Ensures the Intel macOS app can resolve @lydell/node-pty-darwin-x64 at runtime
- Adds a desktop packaging regression test for the architecture config

## Verification
- Verified with a clean temp install that both @lydell/node-pty-darwin-arm64 and @lydell/node-pty-darwin-x64 are installed and resolvable
- pnpm install --frozen-lockfile
- pnpm --filter @openwork/desktop test
- pnpm --filter @openwork/desktop typecheck:electron
- git diff --check

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release-macos-aarch64.yml
+++ b/.github/workflows/release-macos-aarch64.yml
@@ -326,7 +326,20 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-electron-pnpm-
 
+      - name: Install dependencies (macOS dual arch)
+        if: matrix.os_type == 'macos'
+        run: |
+          set -euo pipefail
+          # The x64 macOS app is built on an arm64 runner, but node-pty
+          # resolves its CPU-specific optional package at runtime.
+          {
+            echo "supportedArchitectures.cpu[]=x64"
+            echo "supportedArchitectures.cpu[]=arm64"
+          } >> .npmrc
+          pnpm install --frozen-lockfile
+
       - name: Install dependencies
+        if: matrix.os_type != 'macos'
         run: pnpm install --frozen-lockfile
 
       - name: Write Electron notary API key (macOS)
--- a/apps/desktop/electron/packaging.test.mjs
+++ b/apps/desktop/electron/packaging.test.mjs
@@ -1,0 +1,23 @@
+import { readFileSync } from "node:fs";
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { resolve } from "node:path";
+import { parse } from "yaml";
+
+const repoRoot = resolve(import.meta.dirname, "..", "..", "..");
+
+describe("desktop packaging dependencies", () => {
+  it("installs both CPU variants of native packages for Electron release builds", () => {
+    const workflow = parse(
+      readFileSync(resolve(repoRoot, ".github/workflows/release-macos-aarch64.yml"), "utf8"),
+    );
+    const steps = workflow.jobs["publish-electron"].steps;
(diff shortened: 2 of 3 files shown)


Pablosinyores:
Clean fix — pinning `supportedArchitectures` so both CPU variants install for release builds is the right lever for the mac x64 node-pty gap, and good to see it locked down with `packaging.test.mjs`.

One question: with `os: [current]` and `cpu: [current, x64, arm64]`, does this also pull the x64 native binaries on Linux/Windows dev installs (not just the mac release), and if so is the extra install size/time acceptable? If the dual-arch fetch only needs to matter for the mac release build, scoping it there would keep everyday installs lean — but if one global setting is preferred for simplicity, that's a reasonable trade too. @Iflaqbhat

Iflaqbhat:
@Pablosinyores Thanks, agreed. I scoped the dual-arch install to the macOS Electron release job instead of keeping it in the global workspace config, so everyday Linux/Windows installs stay lean.

The release workflow now writes the temporary pnpm supportedArchitectures CPU config only for macOS release packaging, where the x64 app is built on the arm64 runner. I also updated the packaging test to cover that workflow guard.

Verification:
- pnpm --filter @openwork/desktop test
- pnpm --filter @openwork/desktop typecheck:electron
- git diff --check

PR #2314, opened 2026-06-18
ci: enforce signed commits

OmarMcAdam:
## Summary
- sign automated repository/AUR commits with the OPENWORK_GH_AGENT GPG key
- add a pull_request_target workflow that closes PRs with unsigned commits
- use OPENWORK_GH_AGENT secrets for bot identity, token, and GPG material

## Required secrets
- OPENWORK_GH_AGENT_NAME
- OPENWORK_GH_AGENT_EMAIL
- OPENWORK_GH_AGENT_TOKEN
- OPENWORK_GH_AGENT_GPG_PRIVATE_KEY
- OPENWORK_GH_AGENT_GPG_PASSPHRASE

## Tests
- git diff --check
- ruby YAML parse for modified workflow files
- actionlint on modified workflows, if installed

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/aur-validate.yml
+++ b/.github/workflows/aur-validate.yml
@@ -134,6 +134,8 @@ jobs:
       INPUT_ARTIFACT_NAME: ${{ inputs.artifact_name || 'openwork-electron-linux-x64' }}
       PUSH_TO_AUR: ${{ inputs.push_to_aur && 'true' || 'false' }}
       AUR_REPO: ${{ inputs.aur_repo || 'openwork' }}
+      OPENWORK_GH_AGENT_NAME: ${{ secrets.OPENWORK_GH_AGENT_NAME }}
+      OPENWORK_GH_AGENT_EMAIL: ${{ secrets.OPENWORK_GH_AGENT_EMAIL }}
     steps:
       - name: Validate workflow inputs
         shell: bash
@@ -173,6 +175,7 @@ jobs:
             base-devel \
             curl \
             git \
+            gnupg \
             jq \
             namcap \
             openssh \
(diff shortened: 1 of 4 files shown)


Pablosinyores (commented):
The `pull_request_target` trigger here is used safely — the job only calls the GitHub API (`listCommits`, `createComment`, `pulls.update`) and never checks out or runs PR-controlled code, so the usual `pull_request_target` code-injection risk does not apply.

One concern on the close behavior. The workflow runs on `synchronize` and `reopened` as well as `opened`, and closes the PR whenever any commit fails `verification.verified`. So a contributor who pushes a follow-up commit without signing, or force-pushes mid-review, gets the PR closed from under them even after a maintainer has started reviewing. Reopening works but it is a jarring loop.

Two options that keep the enforcement without the hostile UX:
- fail the check (`core.setFailed`) without `pulls.update({ state: "closed" })`, so branch protection blocks the merge but the PR stays open for the contributor to fix; or
- keep the close only for `types: [opened, reopened]` and just fail (not close) on `synchronize`.

Minor: `verification.verified === true` also covers SSH-signed and GitHub-web-UI commits, not only the GPG key imported in the release workflows here. That is probably the intent, just worth confirming that is the accepted signing surface.

PR #2329, opened 2026-06-21
fix(settings): keep theme previews independent of active theme

### LINKED ISSUE #2328: [Bug]: Theme picker previews inherit the active theme
### Summary

In Settings > Appearance, the Light, Dark, and System preview cards do not consistently show their intended colors.

### Steps to reproduce

1. Run OpenWork from current `dev`.
2. Open Settings > Appearance.
3. Select Light.
4. Look at the Dark and System theme previews.

### Expected behavior

- Light preview stays white.
- Dark preview stays black.
- System preview stays split white and black.
- These previews should be stable regardless of the active theme.

### Actual behavior

When Light is selected, the Dark preview appears light/white and the System preview loses its black half. The preview colors inherit the active app surface instead of showing fixed examples.

### Root cause

`ThemePreview` uses `bg-zinc-950`, but OpenWork's Tailwind color configuration does not include the `zinc` palette. The class is not emitted, so that area becomes transparent and inherits the active theme background.

### Evidence

Screen recording attached below.

https://github.com/user-attachments/assets/a74491a6-75ab-413e-b2e3-c226f2c2e08e

Iflaqbhat:
## Summary
- Replaces unsupported `bg-zinc-950` utilities with configured `bg-black` utilities in the Appearance theme picker.
- Keeps the Light, Dark, and System preview swatches visually stable, regardless of the active app theme.
- Uses fixed pure white and black preview examples by design; the swatches show the available modes rather than tracking a custom active-theme surface.
- Adds regression coverage for both the intended preview colors and their configured Tailwind palette entries.

## Why
The app overrides Tailwind colors and does not define the Zinc palette. The old `bg-zinc-950` class therefore produced no background rule, letting the preview inherit its parent theme.

## Issue
- Closes #2328

## Scope
- `apps/app/src/react-app/domains/settings/appearance/theme-section.tsx`
- `apps/app/tests/theme-preview.test.ts`

## Out of scope
- Changing the selected theme behavior or application theme tokens.

### Ran
- `git diff --check`

### Result
- pass: 73 tests, typecheck, production build, and whitespace check.
- build emitted existing `use memo` and chunk-size warnings only.

## CI status
- pass: local verification above.
- code-related failures: none.
- external/env/auth blockers: none.

## Manual verification
1. Open Settings > Appearance.
2. Select Light, Dark, and System in turn.
3. Confirm the previews remain: Light = white, Dark = black, System = white/black split.

## Evidence
- [Original reproduction recording and report](https://github.com/different-ai/openwork/issues/2328)

## Risk
- Low. This changes only fixed preview swatch classes.

## Rollback
- Revert commits `fac8b55b` and `b6d5995`.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/react-app/domains/settings/appearance/theme-section.tsx
+++ b/apps/app/src/react-app/domains/settings/appearance/theme-section.tsx
@@ -14,6 +14,11 @@ import {
 
 type ThemeMode = AppearanceViewProps["themeMode"];
 
+export const THEME_PREVIEW_CLASSES: Record<"light" | "dark", string> = {
+  light: "bg-white",
+  dark: "bg-black",
+};
+
 interface ThemeSectionProps
   extends Pick<AppearanceViewProps, "busy" | "themeMode" | "setThemeMode"> {}
 
@@ -69,14 +74,14 @@ function ThemePicker(props: ThemePickerProps) {
         value="light"
         label={t("settings.theme_light")}
       >
-        <ThemePreview value="light" className="bg-white" />
+        <ThemePreview value="light" className={THEME_PREVIEW_CLASSES.light} />
         <ThemePickerLabel>{t("settings.theme_light")}</ThemePickerLabel>
       </ThemePickerItem>
       <ThemePickerItem
         value="dark"
         label={t("settings.theme_dark")}
       >
-        <ThemePreview value="dark" className="bg-zinc-950" />
+        <ThemePreview value="dark" className={THEME_PREVIEW_CLASSES.dark} />
         <ThemePickerLabel>{t("settings.theme_dark")}</ThemePickerLabel>
       </ThemePickerItem>
     </ToggleGroup>
@@ -116,8 +121,8 @@ function ThemePreview(props: ThemePreviewProps) {
     >
       {props.value === "system" && (
         <div className="flex h-full">
-          <div className="w-1/2 bg-white" />
-          <div className="w-1/2 bg-zinc-950" />
+          <div className={cn("w-1/2", THEME_PREVIEW_CLASSES.light)} />
+          <div className={cn("w-1/2", THEME_PREVIEW_CLASSES.dark)} />
         </div>
       )}
     </div>
--- a/apps/app/tests/theme-preview.test.ts
+++ b/apps/app/tests/theme-preview.test.ts
@@ -1,0 +1,10 @@
+import { describe, expect, test } from "bun:test";
+
+import { THEME_PREVIEW_CLASSES } from "../src/react-app/domains/settings/appearance/theme-section";
+
+describe("theme picker previews", () => {
+  test("uses fixed light and dark preview colors", () => {
+    expect(THEME_PREVIEW_CLASSES.light).toBe("bg-white");
+    expect(THEME_PREVIEW_CLASSES.dark).toBe("bg-black");
+  });
+});
--- a/apps/app/tests/theme-preview.test.ts
+++ b/apps/app/tests/theme-preview.test.ts
@@ -1,10 +1,19 @@
 import { describe, expect, test } from "bun:test";
 
+import tailwindConfig from "../tailwind.config";
 import { THEME_PREVIEW_CLASSES } from "../src/react-app/domains/settings/appearance/theme-section";
 
 describe("theme picker previews", () => {
   test("uses fixed light and dark preview colors", () => {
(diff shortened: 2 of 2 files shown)


evanklem:
Pulled this branch and verified the fix. Reads correct to me.

What I checked:

- Root cause matches the description.
  - The dark previews used `bg-zinc-950`.
  - `apps/app/tailwind.config.ts` overrides `theme.colors` with the Radix palette, which has no `zinc`.
  - Because of that, `bg-zinc-950` compiled to no CSS rule and the dark swatch rendered transparent, so the card showed through.
  - Switching to `bg-black` / `bg-white`, which are in the palette, fixes it.

- Built the app CSS with the real config to confirm emission:

  ```bash
  cd apps/app

  npx @tailwindcss/cli@4.1.18 \
    -i src/app/index.css \
    -o /tmp/out.css \
    --content "./src/**/*.{ts,tsx}"
  ```

  Results:

  - `bg-zinc-950` produces no rule.
  - `.bg-black` is emitted.
  - `.bg-white` is emitted.

- Both affected spots are covered:
  - The Dark swatch.
  - The dark half of the System split.

Both were previously using `bg-zinc-950`.

The Light swatch was already using `bg-white`, so there is no visual change there.

## Screenshot

image

## One small note

Not a blocker: the added test pins the constant strings, `bg-white` / `bg-black`.

That guards the literal values, but it does not assert the thing that actually broke here: whether the class resolves to a real palette color.

A class like `bg-zinc-950` could pass a same-shaped test while still rendering nothing. It might be worth adding a check that ties the class to the configured palette rather than only checking the literal string.

Also FYI, the same root cause, a color class that is not in the Radix palette, shows up in a handful of other components. I put the full list in #2341 so it does not widen this PR.

Iflaqbhat:
@evanklem Thanks for the thorough verification and for documenting the broader issue in #2341.

I pushed b6d5995e to strengthen the regression test: it now checks that the preview color utilities resolve through the configured Tailwind palette, in addition to asserting the intended black/white values.

`pnpm --filter @openwork/app test` and typecheck pass.

Pablosinyores (commented):
Reasonable fix — extracting `THEME_PREVIEW_CLASSES` and adding the regression test are good.

One thing to verify: the rationale is that the app Tailwind override drops `bg-zinc-950`. If that override replaces the palette rather than just removing zinc, it is worth confirming `bg-black`/`bg-white` are actually in the configured palette too — otherwise the dark swatch can hit the same "unsupported utility" problem the PR is fixing. The test asserts the class names but not that they resolve to a real color; a check against the Tailwind config (or a computed-style assertion) would lock that down.

Also worth a line in the PR description: the swatches are now fixed pure black/white rather than the live theme background. That matches the "independent of active theme" goal, but means they will not track a custom dark palette — fine if intentional.

Iflaqbhat:
@Pablosinyores Thanks, good call. I pushed b6d5995 to strengthen the regression test: it now imports the configured Tailwind palette and verifies the preview utilities resolve to real colors. `bg-black` and `bg-white` pass; the prior `bg-zinc-950` does not.

The previews intentionally use fixed pure black/white examples rather than live theme-surface tokens, so they remain stable while selecting Light, Dark, or System. I added that design intent to the PR description as well.

Pablosinyores (commented):
Good cleanup — `bg-zinc-950` not being in the configured palette meant the dark swatch was effectively unstyled, so switching to `bg-black` and centralizing in `THEME_PREVIEW_CLASSES` is the right call.

The stated goal is that previews stay fixed "regardless of the active theme". That only fully holds if `white`/`black` in this Tailwind config resolve to static values rather than theme-driven CSS variables — worth a quick check that they aren't remapped per theme, otherwise the swatches would still track the active surface.

Minor: the test reads `tailwindConfig.theme.colors` directly, which works while colors live at `theme.colors`. If the config ever moves them under `theme.extend.colors`, the assertion silently breaks — `resolveConfig(tailwindConfig).theme.colors` would be more robust.

PR #2339, opened 2026-06-22
Merge plugin system prompts for provider compatibility

### LINKED ISSUE #2099: [Bug]: opencode customized model api calling router failure
### Summary

I used CC-Switch to have added a custom model api for OpenCode. I can interact in opencode cli with the model correctly. However, it does not work in OpenWork when using the model api. It just shows as below: 

System message must be at the beginning.
Status: 400
Response: {"error":{"message":"System message must be at the beginning.","type":"BadRequestError","param":null,"code":400}}

### To Reproduce

envs:

- MacOS
- OpenCode: 1.16.2
- OpenWork: v0.15.2
- CC-Switch: 3.16.1
  - API Format: OpenAI Compatible
  - Base URL: http://[my model IP]:[PORT]/v1
 
steps:

1. cd to/your_project_dir
2. use `opencode` to activate opencode and change the model to the one I just added, and start conversation. Everything goes right. `/exit` it.
3. start OpenWork and make sure the customized provider and model are connected. Enter anything in TUI, and you will be responsed with the error.

### Expected behavior

OpenWork did recognize the config of OpenCode and the customized model api.
OpenCode CLI works normally as well as OpenWork desktop app.

### Actual behavior

OpenCode CLI works normally while OpenWork throws an exception.

### Screenshots (optional)

Image

_No response_

_No response_

VamsiKrishna0101:
## Summary
- Fixes strict OpenAI-compatible provider failures caused by multiple plugin-added system messages.
- Merges OpenWork plugin system guidance into a single system prompt while preserving the existing agent instructions.
- Adds tests covering system prompt merging for OpenWork capabilities and extension guidance plugins.

## Why
- Some OpenAI-compatible providers/proxies reject requests when more than one `system` message is sent.
- Issue #2099 reports this with a custom OpenAI-compatible API through CC-Switch: OpenCode CLI worked, but OpenWork failed with `System message must be at the beginning`.
- OpenWork plugins were appending additional system entries, which could produce provider payloads that strict APIs reject.
- The fix keeps the OpenWork guidance but sends it as one provider-safe system prompt.

## Issue
- Closes #2099

## Scope
- Updates `apps/server/src/opencode-plugins/openwork-capabilities-knowledge.ts`.
- Updates `apps/server/src/opencode-plugins/openwork-extensions-preview.ts`.
- Preserves OpenWork capabilities, docs, UI-control, extension, browser, voice, cloud, and session-memory guidance.
- Flattens non-empty system prompt entries into one string separated by blank lines.
- Mutates the system prompt array in place so hook callers that keep the original array reference see the merged prompt.
- Adds focused unit coverage for both plugin transforms.

## Out of scope
- No provider configuration changes.
- No model/provider SDK changes.
- No changes to OpenCode CLI behavior.
- No UI changes.
- No new dependencies.
- No changes to message persistence or SQLite handling.

### Ran
- `pnpm --filter openwork-server exec bun test src/opencode-plugins/openwork-capabilities-knowledge.test.ts src/opencode-plugins/openwork-extensions-preview.test.ts`

### Result
- pass:
  - `openwork-capabilities-knowledge.test.ts`: 2 passed
  - `openwork-extensions-preview.test.ts`: 1 passed
  - `tsc -p tsconfig.json --noEmit`: passed
  - N/A

## CI status
- pass:
  - Not run locally beyond targeted tests and server typecheck.
- code-related failures:
  - N/A
- external/env/auth blockers:
  - N/A

## Manual verification
1. Configured a local OpenAI-compatible test provider against a fake `/v1` server.
2. Sent a message from the OpenWork desktop app using the custom provider.
3. Verified the provider request contains a single leading `system` message instead of multiple plugin-added system messages.

## Evidence
- Local fake-provider request body confirmed the final `messages` payload contains only one `system` message at the beginning.
- Screenshot/video: N/A

## Risk
- Low-to-medium.
- The change is limited to plugin system prompt transforms.
- The agent still receives the existing OpenWork guidance, but now as one merged system prompt.
- Main behavioral risk is prompt-size/order sensitivity from concatenating existing system entries into one string.

## Rollback
- Revert this PR to restore the previous behavior where these plugins append separate system prompt entries.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -10,7 +10,7 @@
     "dev": "OPENWORK_DEV_MODE=1 bun src/cli.ts",
     "test": "bun test",
     "test:artifacts": "bun test src/artifact-files.e2e.test.ts src/workspace-init.test.ts src/server.normalizeWorkspaceRelativePath.test.ts",
-    "build": "tsc -p tsconfig.json && bun build src/opencode-plugins/openwork-extensions-preview.ts src/opencode-plugins/openwork-capabilities-knowledge.ts src/opencode-plugins/openwork-anthropic-adaptive-thinking.ts src/opencode-plugins/openwork-anthropic-tool-schema.ts --outdir dist/opencode-plugins --target node --format esm",
+    "build": "tsc -p tsconfig.json && bun build src/opencode-plugins/openwork-extensions-preview.ts src/opencode-plugins/openwork-capabilities-knowledge.ts src/opencode-plugins/openwork-anthropic-adaptive-thinking.ts src/opencode-plugins/openwork-anthropic-tool-schema.ts src/opencode-plugins/openwork-system-prompt-normalizer.ts --outdir dist/opencode-plugins --target node --format esm",
     "build:bin": "bun build --compile src/cli.ts --outfile dist/bin/openwork-server",
     "build:bin:all": "bun ./script/build.ts --outdir dist/bin --target bun-darwin-arm64 --target bun-darwin-x64 --target bun-linux-x64 --target bun-linux-arm64 --target bun-windows-x64",
     "start": "bun dist/cli.js",
--- a/apps/server/src/opencode-plugins/openwork-capabilities-knowledge.test.ts
+++ b/apps/server/src/opencode-plugins/openwork-capabilities-knowledge.test.ts
@@ -3,6 +3,23 @@ import { resolve } from "node:path";
 import { OpenWorkCapabilitiesKnowledge } from "./openwork-capabilities-knowledge.js";
 
 describe("OpenWork capabilities knowledge plugin", () => {
+  test("adds capabilities knowledge to the system prompt", async () => {
+    const plugin = await OpenWorkCapabilitiesKnowledge();
+    const output = {
+      system: [
+        "You are OpenWork.",
+        "",
+      ],
+    };
+
+    await plugin["experimental.chat.system.transform"]({}, output);
+
+    expect(output.system).toHaveLength(3);
+    expect(output.system[0]).toBe("You are OpenWork.");
+    expect(output.system[2]).toContain("You are running inside OpenWork");
+    expect(output.system[2]).toContain("OpenWork product questions");
+  });
+
   test("retrieves Slack connection guidance from bundled docs", async () => {
     process.env.OPENWORK_DOCS_DIR = resolve(import.meta.dir, "../../../../packages/docs");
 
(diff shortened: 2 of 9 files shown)


Pablosinyores (commented):
The in-place collapse is correct for a single transform, but the composition is worth a look. Both `OpenWorkCapabilitiesKnowledge` and `OpenWorkExtensionsPreview` register `experimental.chat.system.transform` and each flattens the entire `output.system` array. That only yields a single system message at send time if these two transforms run last. Any other plugin (or the base agent) that pushes a system entry after them reintroduces multiple system messages and the strict-proxy failure comes back. The guarantee is order-dependent and breaks silently if plugin order changes.

A single centralized flatten applied once, after all `system.transform` hooks have run, would be more robust than each plugin mutating the shared array, and it avoids duplicating the same six lines per plugin.

On the tests: both new cases exercise one plugin against a fresh array, so they do not cover the multi-system-message scenario the fix actually targets. A test that runs both transforms in sequence on the same `output.system` (or one transform over a pre-seeded multi-entry array) would prove the provider-compat behavior rather than the isolated case.

VamsiKrishna0101:
> The in-place collapse is correct for a single transform, but the composition is worth a look. Both `OpenWorkCapabilitiesKnowledge` and `OpenWorkExtensionsPreview` register `experimental.chat.system.transform` and each flattens the entire `output.system` array. That only yields a single system message at send time if these two transforms run last. Any other plugin (or the base agent) that pushes a system entry after them reintroduces multiple system messages and the strict-proxy failure comes back. The guarantee is order-dependent and breaks silently if plugin order changes.
> 
> A single centralized flatten applied once, after all `system.transform` hooks have run, would be more robust than each plugin mutating the shared array, and it avoids duplicating the same six lines per plugin.
> 
> On the tests: both new cases exercise one plugin against a fresh array, so they do not cover the multi-system-message scenario the fix actually targets. A test that runs both transforms in sequence on the same `output.system` (or one transform over a pre-seeded multi-entry array) would prove the provider-compat behavior rather than the isolated case.

Good point, thanks. I updated the PR to avoid per-plugin flattening.

The OpenWork guidance plugins now only append their own system context, and I added a dedicated `openwork-system-prompt-normalizer` plugin that runs after the OpenWork/runtime plugin list. That final plugin merges non-empty system entries into one provider-safe system prompt in place.

I also updated the tests so the provider-compat behavior is covered at composition level: the new test runs both OpenWork system transforms on the same `output.system`, verifies multiple system entries exist before normalization, then runs the final normalizer and asserts the result is exactly one system message while preserving the capabilities and extension guidance.

I also added the new normalizer plugin to the server build entrypoints so it is bundled for packaged builds.

ryan-stellargrammar:
Still encountering this while trying to use OpenWork 0.18.32 with locally hosted HF model `mlx-community/Qwen3.6-35B-A3B-OptiQ-4bit` running under `optiq serve`. Why was this PR closed w/out merge?

PR #2066, opened 2026-06-02
Store runtime OpenCode config in OpenWork state

benjaminshafii:
## Summary
- Stop OpenWork bootstrap and desktop workspace creation from writing user-owned `opencode.jsonc` for runtime defaults.
- Store OpenWork-managed runtime OpenCode config in an OpenWork-owned Drizzle-backed SQLite table keyed by workspace id.
- Merge persisted user OpenCode config + runtime DB config for MCP/plugin/authorized-folder reads, and inject runtime DB config through `OPENCODE_CONFIG_CONTENT` for managed OpenCode.
- Tolerate malformed user-owned `opencode.jsonc` so corrupted user config does not take down OpenWork-managed runtime MCP/plugin/folder flows.
- Avoid native SQLite ABI mismatch in Electron by using `node:sqlite` in non-Bun runtime paths while keeping Drizzle/Bun SQLite for Bun tests/server paths.
- Add an Advanced settings migration action that moves legacy OpenWork-owned runtime config from `.opencode/openwork.json` into the runtime DB while preserving workspace metadata.
- Extend the migration action to move safe OpenWork-managed top-level keys from `opencode.jsonc` into runtime DB: `default_agent` when it is `openwork`, `plugin`, `disabled_providers`, and `provider`.
- Add an Advanced settings runtime config status panel showing runtime DB keys/JSON, legacy `.opencode/openwork.json` migratable keys, parse warnings, user-owned `opencode.jsonc` keys, and migratable `opencode.jsonc` keys.
- Make Advanced settings useful when OpenCode is unavailable by distinguishing OpenCode engine health from OpenWork server/runtime config diagnostics.

## Transition Path
- Existing `opencode.jsonc` remains user-owned by default, but users can explicitly move safe OpenWork-managed runtime keys from it using Advanced settings.
- The explicit migration removes only migrated keys from `opencode.jsonc`; `$schema` and unrelated user keys are preserved.
- Prior OpenWork-owned runtime keys in `.opencode/openwork.json` (`plugin`, `mcp`, `permission`, `provider`) can also be migrated explicitly from Advanced settings.
- The migration keeps existing runtime DB values on conflict, imports missing legacy/user-file state, then strips only migrated keys from their source file.
- Workspace metadata in `.opencode/openwork.json` is preserved.
- Advanced settings now makes the source split visible so users can see what is in runtime DB versus user `opencode.jsonc`.
[...]

PR #2343, opened 2026-06-22
fix(server): migrate project opencode.jsonc MCPs into runtime store

### LINKED ISSUE #2342: [Bug]: MCPs in a project opencode.jsonc can't be removed from the UI after the runtime-store move
## Summary

If a workspace has MCP servers defined in its project `opencode.jsonc`, the app lists them but you can't remove them. Clicking remove returns 200 and the entry comes back after the engine reloads. This mostly hits agents/MCPs that were added before MCP storage moved to the runtime config store (#2066), because the old add flow wrote them straight into `opencode.jsonc`.

## To Reproduce

1. Add an mcp entry to a workspace `opencode.jsonc`:
   ```jsonc
   { "mcp": { "nova-mail": { "type": "remote", "url": "https://example.com/mcp/mail", "enabled": true } } }
   ```
2. Open that workspace in OpenWork. The MCP shows up in the list.
3. Remove it from the UI.
4. It disappears for a second, then reappears after the engine reloads.

## Expected behavior

Removing it from the UI actually removes it, or there's a path that does.

## Actual behavior

`DELETE /workspace/:id/mcp/:name` returns 200 but only writes the runtime config store. The entry still lives in `opencode.jsonc`, and `listMcp` unions the project file back in, so it comes back.

There's a migrate route (`POST /workspace/:id/runtime-config/migrate`) that's supposed to lift legacy config into the runtime store, but it doesn't help here: `userRuntimeConfigFromOpencodeConfig` reads `default_agent`, `plugin`, `disabled_providers` and `provider` from the project `opencode.jsonc`, but not `mcp`. So project-file MCPs are never copied into the store and never stripped from the file. The other legacy source, `legacyRuntimeConfigFromOpenworkConfig`, does handle `mcp`, so the user-config path looks like an oversight.

## OW version & Desktop info

OpenWork server 0.17.1, on macOS.

## Additional context

Opening a PR for this.

Grantmartin2002:
## Summary
- the runtime-config migrate route now lifts `mcp` entries out of a project `opencode.jsonc` into the runtime config store, the same way it already does for plugins, providers and default_agent
- add an e2e test for the migrate route that asserts a project-file mcp lands in the runtime store and is stripped from the file

## Why
MCPs written into a project `opencode.jsonc` (e.g. by the add flow before MCP storage moved to the runtime store in #2066) can't be removed from the UI. Delete only touches the runtime store, but `listMcp` unions the project file back in, so the entry returns 200 on delete and reappears on reload. The migrate route is the intended fix-up path, but `userRuntimeConfigFromOpencodeConfig` read default_agent/plugin/disabled_providers/provider and skipped `mcp`, while `legacyRuntimeConfigFromOpenworkConfig` already handled `mcp`. The merge and the opencode.jsonc strip both key off the returned key list, so adding `mcp` to the extraction and the key tuple makes both the copy-into-store and the remove-from-file work.

## Issue
- Closes #2342

## Scope
- `apps/server/src/server.ts`: `userRuntimeConfigFromOpencodeConfig` + `USER_OPENCODE_RUNTIME_CONFIG_KEYS`
- new `apps/server/src/runtime-config-migrate.e2e.test.ts`

## Out of scope
- the `DELETE /mcp` route itself
- the global `~/.config/opencode` case (delete still can't touch a global entry; that's a separate limitation)

### Ran
- `./node_modules/.bin/tsc -p tsconfig.json --noEmit` (in `apps/server`)

### Result
- pass: typecheck clean
- I could not run the new e2e locally. Every `startServer`-based e2e (including the existing `authorized-folders.e2e`) fails on this machine because `better-sqlite3`'s native binary won't build on my Node (v26; repo pins 24, no prebuild). The new test copies the `authorized-folders.e2e` harness, so it should run under CI. To run it: `bun test src/runtime-config-migrate.e2e.test.ts` from `apps/server`.

## CI status
- pass: relying on CI for the e2e
- external/env/auth blockers: local `better-sqlite3` build (Node version), not the change

## Manual verification
1. Hit this on a real workspace whose `opencode.jsonc` had mcp entries that wouldn't delete from the UI (200, then reappear on reload). Confirmed they live in the project file and `listMcp` tags them `config.project`.
2. Traced the migrate route and confirmed `userRuntimeConfigFromOpencodeConfig` never reads `mcp`.
3. After the change, typecheck passes and the new e2e encodes the expected before/after (entry in runtime store, gone from `opencode.jsonc`).

## Evidence
- N/A (server-side fix, no UI change)

## Risk
- Low. It only adds `mcp` to a migration path that already moves the other keys the same way, and it matches how the openwork-legacy source already migrates `mcp`.

## Rollback
- Revert this commit.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/src/runtime-config-migrate.e2e.test.ts
+++ b/apps/server/src/runtime-config-migrate.e2e.test.ts
@@ -1,0 +1,117 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { startServer } from "./server.js";
+import type { ServerConfig } from "./types.js";
+import { readRuntimeOpencodeConfig } from "./runtime-opencode-config-store.js";
+
+type Served = {
+  port: number;
+  stop: (closeActiveConnections?: boolean) => void | Promise<void>;
+};
+
+const CLIENT_TOKEN = "owt_runtime_migrate_client";
+const HOST_TOKEN = "owt_runtime_migrate_host";
+const stops: Array<() => void | Promise<void>> = [];
+const roots: string[] = [];
+const priorDataDir = process.env.OPENWORK_DATA_DIR;
+const priorTokenStore = process.env.OPENWORK_TOKEN_STORE;
+
+function asRecord(value: unknown): Record<string, unknown> {
+  if (!value || typeof value !== "object" || Array.isArray(value)) return {};
+  return value as Record<string, unknown>;
+}
+
+function clientAuth() {
+  return { authorization: `Bearer ${CLIENT_TOKEN}`, "content-type": "application/json" };
+}
+
+async function createTempRoot(prefix: string) {
+  const root = await mkdtemp(join(tmpdir(), prefix));
+  roots.push(root);
+  return root;
+}
+
+async function startOpenworkServer(workspaceRoot: string) {
+  const config: ServerConfig = {
+    host: "127.0.0.1",
+    port: 0,
+    configPath: join(workspaceRoot, "server.json"),
+    token: CLIENT_TOKEN,
+    hostToken: HOST_TOKEN,
+    approval: { mode: "auto", timeoutMs: 1000 },
+    corsOrigins: ["*"],
+    workspaces: [{ id: "ws_1", name: "Workspace", path: workspaceRoot, preset: "starter", workspaceType: "local" }],
+    authorizedRoots: [workspaceRoot],
+    readOnly: false,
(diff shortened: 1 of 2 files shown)


benjaminshafii (approved):
Thank you for the clear fix and test coverage. I validated the migration path and MCP delete symptom locally and on Daytona; this looks good.

PR #2359, opened 2026-06-24
fix(server): don't let remote workspaces block engine startup

joniler:
## Summary
- A configured **remote** workspace stops the local managed OpenCode engine from starting, so **local** workspaces can no longer connect (while the remote keeps working). Fixed by only provisioning workspace files for workspaces that actually have local files to set up.

## Why
- Remote workspaces are persisted with an empty `path`. Both `startEmbeddedServer` (desktop) and the CLI ran `for (const workspace of config.workspaces) await ensureWorkspaceFiles(workspace.path, ...)`.
- `ensureWorkspaceFiles()` throws `ApiError(400, "invalid_workspace_path", "workspace path is required")` on a blank path, which aborts startup. The managed engine never binds, so every local workspace fails to connect.
- Remote workspaces are unaffected (they reach their host directly), producing the confusing **"remote works, local is broken"** symptom whenever both are configured. `embedded.ts` already special-cases `workspaceType === "remote"` a few lines below the offending loop — the provisioning loop just didn't.

## Scope
- `apps/server/src/workspace-init.ts`: new `ensureLocalWorkspaceFiles()` helper. It skips workspaces that have no local files to provision — **remote** workspaces (which live on a host and may even carry a non-empty remote `directory`) **and** any workspace **without a resolved local path** (e.g. a legacy entry that defaults to `workspaceType: "local"` with an empty path, per `buildWorkspaceInfos`). Both conditions are needed: a `path`-only guard would wrongly provision local files into a remote's `directory`, and a `workspaceType`-only guard would miss a pathless local entry.
- `apps/server/src/embedded.ts`, `apps/server/src/cli.ts`: call the helper instead of the duplicated inline loop.
- `apps/server/src/workspace-init.test.ts`: regression tests.

## Out of scope
- `ensureWorkspaceFiles()` still throws on a blank path — that validation is correct for the single-path callers (`routes/workspaces.ts`, `server.ts`) which pass a resolved local path.
- `reload-watcher.ts` also iterates workspaces, but already wraps per-workspace startup in try/catch (degrades gracefully, doesn't abort), so it's left as-is.

### Ran (in `apps/server`)
- `pnpm install --filter "openwork-server..."`
- `pnpm run typecheck`
- `bun test src/workspace-init.test.ts`

### Result
- pass/fail: **pass**
- `typecheck` exits 0; **11 tests pass** (3 new, covering: local provisioned + remote skipped, remote with empty path doesn't throw, pathless non-remote skipped).
- Full `pnpm test:e2e` not run here (needs the OpenCode runtime + a display); the change is covered by the new server-level tests + typecheck.

## CI status
- pass: pending CI

## Manual verification
1. Configure one local + one remote workspace (the remote is stored with an empty `path` in `server.json`).
2. **Before:** desktop boot fails; managed OpenCode never binds; the local workspace cannot connect to the engine. The remote still works.
3. **After:** boot succeeds, the engine binds, the local workspace connects, and the remote is unchanged.

## Evidence
No video/screenshot attached (the repro uses a private remote worker, and this is a server-side boot crash best shown via logs). Before the fix (OpenWork desktop v0.17.1):
```
ApiError: workspace path is required
    at ensureWorkspaceFiles (resources/app.asar/server/dist/workspace-init.js:42:15)
    at async engineStart (resources/app.asar/electron/runtime.mjs)
  code: 'invalid_workspace_path'
```
After (local-patched build, blank remote path present), the managed engine binds and serves the local workspace:
```
opencode-x86_64 ... serve --hostname 127.0.0.1 --port 41169
GET /workspace/<local-id>/events 200
```
Exact repro for a reviewer: add a local workspace and an OpenWork remote worker, restart the app.

## Risk
- Low. Skipped workspaces have no local files to provision, so this only avoids a guaranteed throw. Validation for real local paths is unchanged, and the duplicated loop is now a single tested helper.

## Rollback
- Revert this commit.

### DIFF
--- a/apps/server/src/cli.ts
+++ b/apps/server/src/cli.ts
@@ -5,7 +5,7 @@ import { mkdir } from "node:fs/promises";
 import { parseCliArgs, printHelp, resolveServerConfig } from "./config.js";
 import { createManagedOpencodeServer, type ManagedOpencodeServer } from "./managed-opencode.js";
 import { createServerLogger, startServer, syncAllWorkspacesRuntimeMcpToEngine } from "./server.js";
-import { ensureWorkspaceFiles } from "./workspace-init.js";
+import { ensureLocalWorkspaceFiles } from "./workspace-init.js";
 import { keepOpenworkRuntimeConfigFileFresh, writeOpenworkRuntimeConfigFile } from "./openwork-runtime-config.js";
 import pkg from "../package.json" with { type: "json" };
 
@@ -27,9 +27,7 @@ const serverUrl = `http://${config.host === "0.0.0.0" ? "127.0.0.1" : config.hos
 let managedOpencode: ManagedOpencodeServer | null = null;
 
 if (!config.readOnly) {
-  for (const workspace of config.workspaces) {
-    await ensureWorkspaceFiles(workspace.path, workspace.preset ?? "starter");
-  }
+  await ensureLocalWorkspaceFiles(config.workspaces);
 }
 
 if (!config.opencodeBaseUrl && process.env.OPENWORK_MANAGE_OPENCODE === "1") {
--- a/apps/server/src/embedded.ts
+++ b/apps/server/src/embedded.ts
@@ -9,7 +9,7 @@ import { mkdir } from "node:fs/promises";
 import { resolveServerConfig, type CliArgs } from "./config.js";
 import { createManagedOpencodeServer, type ManagedOpencodeServer, type OpencodeExecutionSnapshot } from "./managed-opencode.js";
 import { startServer, syncAllWorkspacesRuntimeMcpToEngine } from "./server.js";
-import { ensureWorkspaceFiles } from "./workspace-init.js";
+import { ensureLocalWorkspaceFiles } from "./workspace-init.js";
 import { keepOpenworkRuntimeConfigFileFresh, writeOpenworkRuntimeConfigFile } from "./openwork-runtime-config.js";
 import type { ServeResult } from "./serve-node.js";
 import type { ServerConfig } from "./types.js";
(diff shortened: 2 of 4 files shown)


benjaminshafii:
## Independent verification ✅

Reproduced and validated this fix in a fresh worktree off `dev`.

**Bug proven on base (`dev`):** Wrote a test mirroring the `embedded.ts`/`cli.ts` boot loop (`for (const workspace of config.workspaces) await ensureWorkspaceFiles(workspace.path, ...)`) with a local + remote(empty path) workspace. It throws `workspace path is required`, aborting startup — exactly the reported symptom (remote works, local can't connect to the managed engine).

**Fix verified on PR branch:**
- `bun test src/workspace-init.test.ts` → **11 pass** (8 existing + 3 new regression tests; the local-provisioned-plus-remote-skipped case is the boot-loop equivalent and now succeeds).
- `pnpm run typecheck` → exit 0.
- Broader `src/workspace-init.test.ts + src/server.normalizeWorkspaceRelativePath.test.ts` → 22 pass.

**Safety review:**
- Minimal, well-scoped: skips remote workspaces and pathless entries — the only entries with no local files to provision.
- Local workspaces always carry a validated path, so they're provisioned unchanged.
- Single-path callers (`routes/workspaces.ts`, `server.ts`) keep their correct `invalid_workspace_path` validation. Confirmed no other boot loops iterate all workspaces unconditionally.
- No `any`/casts; uses `Pick<WorkspaceInfo, ...>`.

Only failing checks are Vercel deploy authorizations (auth/env blockers for a first-time contributor, not code failures), and they target app/den/worker-proxy which this server-only change doesn't touch.

Merging.

PR #2360, opened 2026-06-24
fix(server): boot managed engine in first local workspace, not workspaces[0]

joniler:
## Summary
- The server-managed OpenCode engine boots in `config.workspaces[0]`. When index 0 is a remote workspace with no local path, the engine **skips startup entirely**, leaving local workspaces with nothing to connect to. Select the first non-remote workspace with a local path instead.

## Why
- A remote worker added via "Add a worker → Connect remote" is **prepended** to `config.workspaces` (`routes/workspaces.ts`), so index 0 can be a remote workspace whose `path` is `""`. The managed-engine block guards on `workspaces[0]?.path`, so it silently doesn't spawn the engine — and every local workspace then has no engine to reach. The engine serves all workspaces but only needs one local directory to start in; that should be a real local workspace, not whatever happens to sit at index 0.

## Issue
- Companion to #2359 (the remote-workspace boot crash). Surfaced while fixing that; independently correct, since index 0 is not reliably the local/active workspace. With #2359 alone a path-less remote at index 0 no longer crashes provisioning, but the engine could still skip startup — this closes that gap.

## Scope
- `apps/server/src/workspaces.ts`: new `findManagedEngineWorkspace()` — first non-remote workspace with a resolved local path, else `undefined`.
- `apps/server/src/embedded.ts`, `apps/server/src/cli.ts`: use it for the managed-engine cwd / runtime-config selection instead of `config.workspaces[0]`.
- `apps/server/src/workspaces.test.ts`: unit tests.

## Out of scope
- Other `config.workspaces[0]` uses (active-workspace fallbacks, proxy default, image generation) — those are "default/active workspace" selection, a separate concern; left untouched.

### Ran (in `apps/server`)
- `pnpm install --filter "openwork-server..."`
- `pnpm run typecheck`
- `bun test src/workspaces.test.ts`
- `bun test` over the workspace-related non-e2e files (`workspace-init`, `workspace-import-preview`, `workspace-export-safety`, `reload-watcher`, `validators`)

### Result
- pass/fail: **pass**. `typecheck` exits 0; new selection tests **6/6**; **61/61** across the workspace-related suite — no regressions.
- Full `pnpm test:e2e` not run here (needs the OpenCode runtime + a display).

## CI status
- pass: pending CI

## Manual verification
1. Config with a path-less remote worker at index 0 and a local workspace after it.
2. **Before:** managed engine doesn't spawn (`workspaces[0]?.path` is empty); the local workspace has no engine.
3. **After:** the engine boots in the local workspace; remote-only configs still (correctly) spawn no local engine.

## Evidence
No video — this is boot-path/selection logic, deterministically covered by unit tests. The tests assert selection against real config shapes: local+remote in both orders (incl. a path-less remote at index 0), remote-only, a remote that carries a non-empty `directory`, and blank entries.

## Risk
- Low. For the common case (a local workspace at index 0) selection is identical to before. Behavior only changes when index 0 has no local path — where the engine previously failed to start at all.

## Rollback
- Revert this commit.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2360?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/src/cli.ts
+++ b/apps/server/src/cli.ts
@@ -6,6 +6,7 @@ import { parseCliArgs, printHelp, resolveServerConfig } from "./config.js";
 import { createManagedOpencodeServer, type ManagedOpencodeServer } from "./managed-opencode.js";
 import { createServerLogger, startServer, syncAllWorkspacesRuntimeMcpToEngine } from "./server.js";
 import { ensureLocalWorkspaceFiles } from "./workspace-init.js";
+import { findManagedEngineWorkspace } from "./workspaces.js";
 import { keepOpenworkRuntimeConfigFileFresh, writeOpenworkRuntimeConfigFile } from "./openwork-runtime-config.js";
 import pkg from "../package.json" with { type: "json" };
 
@@ -31,8 +32,8 @@ if (!config.readOnly) {
 }
 
 if (!config.opencodeBaseUrl && process.env.OPENWORK_MANAGE_OPENCODE === "1") {
-  const workspace = config.workspaces[0];
-  if (workspace?.path) {
+  const workspace = findManagedEngineWorkspace(config.workspaces);
+  if (workspace) {
     // Server-managed config file: the engine re-reads it from disk on every
     // instance rebuild, and keepOpenworkRuntimeConfigFileFresh rewrites it
     // on every runtime-DB write — so disposes always pick up current state.
--- a/apps/server/src/embedded.ts
+++ b/apps/server/src/embedded.ts
@@ -10,6 +10,7 @@ import { resolveServerConfig, type CliArgs } from "./config.js";
 import { createManagedOpencodeServer, type ManagedOpencodeServer, type OpencodeExecutionSnapshot } from "./managed-opencode.js";
 import { startServer, syncAllWorkspacesRuntimeMcpToEngine } from "./server.js";
 import { ensureLocalWorkspaceFiles } from "./workspace-init.js";
+import { findManagedEngineWorkspace } from "./workspaces.js";
 import { keepOpenworkRuntimeConfigFileFresh, writeOpenworkRuntimeConfigFile } from "./openwork-runtime-config.js";
 import type { ServeResult } from "./serve-node.js";
 import type { ServerConfig } from "./types.js";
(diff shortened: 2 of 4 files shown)


benjaminshafii:
## End-to-end verification ✅ (real server boot)

Beyond the unit tests, I booted the **real `openwork-server` (`bun src/cli.ts`)** with a config matching the exact bug scenario — a path-less remote worker at **index 0** followed by a local workspace — using a stub `opencode` engine binary (announces `opencode server listening on …`) via `OPENWORK_OPENCODE_BIN`. Ran the identical boot on both branches.

**server.json used:**
```
workspaces:
  [0] remote (path: "", workspaceType: remote)   <- prepended remote worker
  [1] local  (path: /…/local-ws, workspaceType: local)
```

**BASE (`dev` @ 752f40922) — bug reproduced:**
```
[base] managed-engine=SKIPPED  fake-engine-process=no
```
Server log shows `OpenWork server listening` + `Workspaces: 2` but **no `Managed OpenCode listening`** line — the `config.workspaces[0]?.path` guard sees the remote's empty path and skips engine startup entirely. The local workspace has no engine to reach.

**PR branch (767924add) — fixed:**
```
[pr] managed-engine=STARTED  fake-engine-process=yes
```
Server log: `Managed OpenCode listening on http://127.0.0.1:50138`. The engine boots in the **local** workspace — confirmed by the provisioned `local-ws/.opencode/openwork.json` whose `authorizedRoots` points at the local path (not the remote at index 0).

**Plus:** `pnpm run typecheck` → exit 0; `bun test src/workspaces.test.ts` → 6/6; broader workspace suite (workspaces + workspace-init + reload-watcher + validators) → 44/44, no regressions.

**Safety:** change is scoped to the two managed-engine boot blocks (`embedded.ts`, `cli.ts`); all other `config.workspaces[0]` uses (active-workspace fallbacks, proxy default, image gen) are untouched. Common case (local at index 0) selects identically to before; behavior only changes when index 0 has no local path — where the engine previously failed to start at all.

Only failing checks are Vercel deploy authorizations (auth/env blockers, not code; target app/den/worker-proxy which this server-only change doesn't touch).

Merging.

PR #2399, opened 2026-06-30
feat(den-api): additive search_capabilities tool on the Cloud Control MCP

benjaminshafii:
## What

Adds one new tool, `search_capabilities`, to the existing, already-shipped Den `/mcp` endpoint ("OpenWork Cloud Control" as seen by the desktop app). It ranks the same OpenAPI-derived catalog already registered there by keyword, so a harness can search instead of being handed the full ~129-tool list, then call whichever real tool it matched via the normal MCP `tools/call` protocol.

## Why

This is the smallest real step toward a "search then execute" facade on top of infrastructure that already exists and is already live in the desktop app. It also directly addresses the original OpenCode-reload pain point: because the harness can rely on `search_capabilities` instead of needing every tool individually surfaced, capabilities can grow behind that one stable tool without the tool list itself needing to change.

## Change

- `ee/apps/den-api/src/mcp/search.ts` (new): pure `searchCapabilities(catalog, query, limit)` — tokenized keyword scoring against tool name (camelCase-aware), summary, and path. No new infra, no new dependencies.
- `ee/apps/den-api/src/mcp/index.ts`: registers `search_capabilities` as one additional `server.registerTool(...)` call, alongside the existing catalog loop. **Purely additive — 25 lines, nothing removed or changed.**
- `evals/flows/mcp-search-capabilities.flow.mjs` (new): a real, end-to-end fraimz flow (see Tests run below).

No change to auth (`auth.ts`), policy (`policy.ts`), or the existing in-process execute path (`invoke.ts`). The existing Cloud Control connection and every existing catalog tool are unaffected — verified explicitly in the flow (`tools/list` still includes `getOrg` etc. alongside the new tool).

## Tests run

`npx tsc -p ee/apps/den-api/tsconfig.json --noEmit` — clean.

Ran the real coded flow (`pnpm fraimz --flow mcp-search-capabilities`) against:
- the real Den API (local, same MySQL-backed demo org seed used elsewhere in `evals/`)
- the real desktop Electron app, signed in as the seeded demo owner

All steps passed:
1. App boots, signs in (skipped if already signed in), resolves the active org, creates a workspace if needed.
2. **OpenWork Cloud Control still shows "Connected"** — proves the addition didn't break the existing, already-shipped surface.
3. Mints a real org-scoped MCP token via the existing `/v1/mcp/token` route.
4. `tools/list` over real JSON-RPC → `search_capabilities` present alongside the full, unchanged 128-tool catalog.
5. `tools/call search_capabilities({query:"list organization"})` → real ranked matches from the live catalog, top hit `getOrg`.
6. `tools/call getOrg` (the matched tool, called via the normal protocol) → runs through the **unchanged** `invoke.ts` path and returns genuine organization data.
7. That same organization name is then confirmed visible in the actual desktop UI (Cloud Account screen) — ties the protocol proof to a real visual fact, not just a JSON blob.
8. **A real chat message in the desktop app** ("...first call search_capabilities... then call whichever tool it matches...") causes the actual OpenCode agent — using a real, already-connected model provider (no manual provider setup needed) — to call `search_capabilities`, then call the matched `getOrg`, with the real result surfaced back in the chat transcript. This is the strongest proof: not a direct protocol call from a test script, but the real chat → real agent → real cloud MCP path.

fraimz: `evals/results/2026-06-30T22-41-53-138Z/fraimz.html` (local; not committed, gitignored results dir). 10/10 steps passed.

## Honest limitations

- The harness still sees the full catalog (`search_capabilities` was added *alongside* the existing 128 tools, not in place of them). In my first manual chat test, the model skipped search entirely and called `getOrg` directly by name, because it already had it in its tool list — I had to explicitly instruct it to use search first to specifically exercise the new tool. Getting the harness down to a true minimal two-tool surface (search + execute only, with the other 128 tools no longer individually registered) is a deliberate follow-up, not included here.
- The chat-triggered step depends on a live LLM call, so it's real but not as deterministic as the pure protocol steps.
- Only tested against one organization (the seeded demo org).
- Screenshots in the local fraimz run were captured using the light-mode runner default from #2398; that PR is not a code dependency of this one (this PR's diff stands alone and was typechecked independently), but pairs well with it for readable evidence.

<a href="https://cubic.dev/pr/different-ai/openwork/pull/2399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ee/apps/den-api/src/mcp/index.ts
+++ b/ee/apps/den-api/src/mcp/index.ts
@@ -1,10 +1,12 @@
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js"
 import { StreamableHTTPTransport } from "@hono/mcp"
 import type { Hono } from "hono"
+import { z } from "zod"
 import { publicRoute, tokenRoute } from "../middleware/index.js"
 import { getMcpResourceUrl, verifyMcpRequest } from "./auth.js"
 import { buildMcpCatalog, getToolDescription, loadOpenApiDocument, type McpToolOperation } from "./catalog.js"
 import { invokeMcpOperation } from "./invoke.js"
+import { SEARCH_CAPABILITIES_TOOL_NAME, searchCapabilities } from "./search.js"
 
 const CATALOG_CACHE_TTL_MS = 5 * 60 * 1000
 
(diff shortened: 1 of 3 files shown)


benjaminshafii:
Closing in favor of a PR from a branch on this repo directly (no fork needed).

PR #2416, opened 2026-07-02
feat(app): manage cloud providers via runtime config, not opencode.jsonc

benjaminshafii:
## Why

Cloud provider imports edited the user's project `opencode.jsonc` — jsonc block inserts/removes with comment preservation. Consequences seen on a real customer machine and reproduced in our evals:
- the user's config file becomes a dumping ground for `lpr_*` blocks
- partial failures leave **stale orphan blocks** that accumulate across imports (v1/v2/v3 provider copies)
- removal correctness depends on fragile client-side read-modify-write of a file the user also owns

## What

The OpenWork server already has the right primitive — `PATCH /workspace/:id/config` with **per-key provider merge** (`mergeRuntimeProviderUpdate`: record values upsert, explicit `null` deletes), feeding the runtime SQLite → `OPENCODE_CONFIG`. Its code comment literally says it exists so *"clients can remove runtime-managed providers (e.g. cloud imports) without read-modify-write races"* — the client just never adopted it. Now it does:

- **Import/reconcile**: `buildRuntimeProviderPatch` upserts the `lpr_*` entry (and null-deletes a renamed predecessor) via `patchConfig`. Zero `opencode.jsonc` edits.
- **Removal**: per-key `null` delete — idempotent, race-free.
- **Migration**: legacy `opencode.jsonc` blocks for the imported ids are stripped best-effort, so blocks written by pre-runtime builds stop shadowing state (this also cleans up the orphan accumulation existing users have).
- **Orphan sweep** (sign-out/startup) now covers both runtime entries and legacy file blocks.
- Dead code removed: `formatConfigWithCloudProvider` + the comment writer.

## Tests / evidence

- `pnpm typecheck` + `bun test tests/` — clean / 86 pass, 0 fail on latest `dev` (reimport diff tests rewritten around the runtime patch builder, incl. renamed-predecessor deletion)
- fraimz e2e (`cloud-config-sync-latency`, extended) against the real Electron app + real local Den — **PASSED**:
  1. palette "Sync cloud config" imports a fresh Den provider in ~1.2s
  2. **runtime injection proof**: the engine serves the provider while `opencode.jsonc` contains no block for it — including a **deliberately pre-seeded legacy block for the same id, which the migration stripped** (asserted via the config-file API)
  3. Den-side model add lands hands-free in **2.7s**
  4. runtime `null` delete removes the provider from the engine

## Notes

- The `cloudImports` **import-record** race (extensions store vs provider-auth store clobbering each other's sections) is still open and tracked separately — this PR removes the *provider block* half of the file-write problem and makes removal idempotent, which stops the orphan accumulation.
- Cloud provider management now requires the OpenWork server connection (it already did for import records; error messaging unchanged).


benjaminshafii:
Rebased onto current dev and re-validated (conflicts resolved):

**Conflict resolution**
- `provider-auth/store.ts`: rebased over #2430 (multi-env credentials) — Omar's `envEntries`/`upsertUserEnv` apply logic untouched; only the provider-block write swapped from jsonc edits to the runtime patch, exactly as before.
- `cloud-config-sync-latency.flow.mjs`: that flow was added by #2414, which was reverted on dev — so the palette "Sync cloud config" + 30s auto-sync + "Synced ago" assertions are gone with it. Replaced by `cloud-runtime-provider-config.flow.mjs`, which proves this PR's actual claims through the surviving end-user path (Settings -> Cloud): Import -> runtime injection (engine serves the provider, zero opencode.jsonc footprint, pre-seeded legacy block migrated), Den model add -> "Out of sync" -> Sync upsert, runtime null-delete removal. The latency assertions return with the #2414 re-land.

**Also folded in**
- Stale copy fix: "Import managed LLM providers into opencode.jsonc..." and the sync detail string no longer mention opencode.jsonc (they were factually wrong with this PR).

**Re-validation on the rebased branch**
- `pnpm typecheck` clean; `bun test tests/` 94 pass, 0 fail (includes #2430's credential tests).
- fraimz `cloud-runtime-provider-config` against the real Electron app + real local Den (`--stack den`): **PASSED** — import lands in ~1.0s, runtime injection + legacy-block migration asserted via the config-file API, Sync upsert in ~0.8s, null-delete removes the provider from the engine.

PR #2429, opened 2026-07-02
feat(server): portable export of skills and runtime MCPs for agent packaging

benjaminshafii:
## Problem

When a user asks the agent to *"grab this skill and MCP X, add it as a plugin, and put it on marketplace Y"*, the plugin/marketplace half already works — but the agent has no canonical way to reference OpenWork-managed runtime extensions:

- Skills are plain files (discoverable), but **MCP servers added through OpenWork live in the runtime DB** (`runtime.sqlite`) and are injected into the engine via the `OPENCODE_CONFIG` file / dynamic `POST /mcp`. There is no read surface for their definitions.
- Reading `$OPENCODE_CONFIG` directly is wrong: it only covers the first workspace, it is derived state, and it can contain secrets.

## What this adds

1. **`POST /workspace/:id/extensions/export`** (client route, read-only) — returns requested skills (full `SKILL.md` content) and MCP servers (from any source, including runtime-DB-managed ones) in one portable shape. Secret-bearing values (`headers.*`, `environment.*`, `env.*`) are **always redacted** and listed as `redactedKeys`, so exports never leak credentials; installers declare them as required inputs instead. (`apps/server/src/extensions-export.ts`)
2. **`openwork_extensions_export` tool** in the bundled `openwork-extensions-preview` plugin — resolves the workspace from the session context (or an explicit `workspaceId`) and calls the endpoint via `OPENWORK_SERVER_URL`/`OPENWORK_SERVER_TOKEN`, matching the server-consumption-first principle.
3. **Capabilities-knowledge prompt section** teaching the agent: runtime extensions are not files, don't read `OPENCODE_CONFIG`/runtime DB directly, use the export tool, never inline redacted secrets, then publish through the existing Cloud MCP / marketplace flow.

The harness task now decomposes cleanly: `openwork_extensions_export(skill x, mcp x)` → portable bundle → existing "create plugin + publish to marketplace" flow.

## Tests

Ran in `apps/server`:

- `bun test src/extensions-export.test.ts` — **6 pass, 0 fail** (redaction unit tests, export module tests against a real runtime DB, HTTP route test against a real started server, and an end-to-end test that loads the actual bundled plugin factory, points it at the live server via env, and calls `openwork_extensions_export` — asserting the secret never appears in output).
- `bun test src` — **257 pass, 0 fail** (full server src suite).
- `pnpm typecheck` — clean.
- `bun build` of both touched plugins — clean.

## Validation note

No UI surface changes, so no fraimz was produced. The observable behavior (server API + agent tool) is covered end-to-end by the plugin-tool test above, which exercises the real server, real runtime SQLite DB, and the real bundled plugin over HTTP. To reproduce: `cd apps/server && bun test src/extensions-export.test.ts`. A full desktop fraimz would additionally require a live model provider driving the agent; happy to follow up with one if desired.

## Security

- Export is read-only; no approval flow needed (same access level as existing `GET /workspace/:id/skills/:name` and `GET /workspace/:id/mcp`).
- All header/environment values are redacted unconditionally — there is no opt-out.


benjaminshafii:
## fraimz: Passed

Ran the full fraimz loop against the real Electron app built from this branch (separate `dev2` profile, CDP on `:9826`), flow committed as `evals/flows/extensions-export-portable.flow.mjs`.

```
▶ extensions-export-portable — Skill + runtime MCP export as a portable, secret-redacted bundle
  ✓ App boots clean
  ✓ User installs a skill and sees it in Settings > Extensions > Skills
  ✓ User connects an MCP server carrying a secret header
  ✓ Export surface returns a portable bundle with the secret redacted
  ✓ Agent exports both via the openwork_extensions_export tool
fraimz: evals/results/2026-07-02T19-21-56-640Z/fraimz.html
```

Frame-by-frame (each claim bound to an action, observable assertion, and validated screenshot):

| Frame | Claim | Witness |
|---|---|---|
| `01-booted` | App boots clean | route reported; no error text |
| `02-skill-installed` | Installed skill visible in Settings > Extensions | screenshot requires `release-notes-eval`; row shows "Connected" with its description |
| `03-mcp-connected` | Runtime-managed MCP (secret `Authorization` header) visible under YOUR APPS | screenshot requires `eval-export-mcp` (shows Paused row), rejects the secret string |
| *(no frame image — read-only API)* | `POST /workspace/:id/extensions/export` round-trips both; secret redacted | asserts: skill content round-trips, `source: config.remote`, `headers.Authorization === "<redacted>"`, `redactedKeys` declares it, raw response never contains the secret |
| `04-agent-export-proof` | **The real agent calls `openwork_extensions_export` and reports the result** | agent reply must be `EXPORT-RESULT <url> <redactedKeys>` — the url and key names exist **only in the tool result, not the prompt**, so the line cannot be produced without actually running the tool; screenshot requires `EXPORT-RESULT` + the exported url, rejects the secret |

Anti-fabrication note: an earlier draft of the agent frame matched marker text that also appeared in the user's own message; it was reworked so the assertion is unguessable from the prompt (echoes tool-result-only values). Also verified visually that the transcript shows the "Running openwork extensions export" tool call.

Repro:
```bash
pnpm install
OPENWORK_ELECTRON_REMOTE_DEBUG_PORT=9826 OPENWORK_ELECTRON_APP_IDENTIFIER=com.differentai.openwork.dev2 \
  OPENWORK_DATA_DIR=$HOME/.openwork/openwork-orchestrator-dev2 PORT=5183 pnpm dev &
pnpm fraimz --flow extensions-export-portable --cdp-url http://127.0.0.1:9826
```

(`evals/results/` is gitignored; screenshots available on request — 1180x820 PNGs, all validations passing per report.json.)

PR #2430, opened 2026-07-02
feat(providers): support multiple env var credentials per cloud provider

OmarMcAdam:
## Problem

Some providers (e.g. AWS Bedrock) need several env vars — `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, `AWS_BEARER_TOKEN_BEDROCK` — but the cloud provider flow only supported a single "API key / credential" field in the dashboard, one secret in the DB, and one `auth.set` on the desktop.

## Approach

**Storage (no migration).** The encrypted `api_key` column is now self-describing: providers with one env key keep storing the bare credential string (every existing row is untouched and keeps working), providers with several env keys store a JSON object of env name → value. Decode is strict shape-based (non-empty JSON object, all string values), so legacy plain keys can't be misread as maps.

**den-api** (`src/llm/provider-credentials.ts`, `routes/org/llm-providers.ts`)
- New `apiKeys: Record<string, string>` write field alongside the untouched legacy `apiKey`. Values merge into the stored credential: non-empty sets, empty string clears, absent keeps — so an admin can rotate one AWS key without re-entering the rest. Unknown env names → 400.
- Legacy strings migrate into the map when a provider's env list grows; maps collapse back to a bare string when it shrinks to one.
- `/connect` stays additive: single-env providers return `apiKey` exactly as before; multi-env providers return `apiKeys` with `apiKey: null`, so **old desktop clients hit their existing "no stored credential" error instead of applying a JSON blob as the key**.
- Responses expose `configuredEnvKeys` (names only, never values) for the editor.

**den-web**
- The Credential section renders one password input per env key when the provider declares several, with "Saved" badges from `configuredEnvKeys` and leave-blank-to-keep placeholders. Single-env providers keep the exact existing UX.
- The guided form no longer bails to the JSON editor for multi-env configs (`envName` → `envNames`), and the JSON editor path leniently extracts env names so it gets per-key inputs too.

**desktop (apps/app)**
- `DenOrgLlmProviderConnection` gains `apiKeys`; `resolveCloudProviderCredentials` splits the payload: every value is upserted as a user env var via `upsertUserEnv` (same mechanism as the existing `OPENAI_API_KEY` flow), and the first env-ordered value becomes the opencode auth.json entry (models.dev `env[0]` convention). Legacy single-key payloads follow the unchanged auth-only path.
- Connect fails explicitly if env vars are needed but the OpenWork server client is unavailable, rather than half-applying the import.

## Testing

- `ee/apps/den-api/test/provider-credentials.test.ts` — 18 tests: codec round-trips, merge/clear/keep semantics, legacy↔map migration in both directions, JSON-lookalike safety.
- `ee/apps/den-web/tests/llm-provider-guided.test.ts` — multi-env guided round-trips and env-name extraction.
- `apps/app/tests/cloud-provider-credentials.test.ts` — credential splitting, env ordering, primary-key selection.
- `tsc --noEmit` clean on den-api, den-web, and apps/app. Remaining suite failures are pre-existing on dev (route-guard list includes routes added by other recent merges; artifact-spreadsheet tests).

### DIFF
--- a/apps/app/src/app/lib/den.ts
+++ b/apps/app/src/app/lib/den.ts
@@ -176,6 +176,12 @@ export type DenOrgLlmProvider = {
 
 export type DenOrgLlmProviderConnection = DenOrgLlmProvider & {
   apiKey: string | null;
+  /**
+   * Per-env-var credential values for providers whose config declares several
+   * env keys (e.g. AWS Bedrock). Servers send either `apiKey` or `apiKeys`,
+   * never both.
+   */
+  apiKeys: Record<string, string> | null;
 };
 
 export type DenOrgMarketplaceResolved = {
@@ -1161,7 +1167,20 @@ function getDenOrgLlmProviderConnection(payload: unknown): DenOrgLlmProviderConn
   return {
     ...provider,
     apiKey: typeof payload.llmProvider.apiKey === "string" ? payload.llmProvider.apiKey : null,
+    apiKeys: parseApiKeysRecord(payload.llmProvider.apiKeys),
   };
+}
+
+function parseApiKeysRecord(value: unknown): Record<string, string> | null {
+  if (!isRecord(value)) {
+    return null;
+  }
+
+  const entries = Object.entries(value).filter(
+    (entry): entry is [string, string] =>
+      typeof entry[1] === "string" && entry[1].trim().length > 0,
+  );
+  return entries.length > 0 ? Object.fromEntries(entries) : null;
 }
 
 function parsePluginConfigObjectType(value: unknown): DenPluginConfigObjectType | null {
--- a/apps/app/src/react-app/domains/connections/provider-auth/cloud-provider-config.ts
+++ b/apps/app/src/react-app/domains/connections/provider-auth/cloud-provider-config.ts
@@ -62,6 +62,33 @@ const addCloudProviderComment = (
 export const getCloudProviderEnv = (config: Record<string, unknown>) =>
   getStringList(config.env);
 
+/**
+ * Split a connect payload's credential into the opencode auth.json entry and
+ * the env vars to upsert. Multi-env providers (`apiKeys`) set every value as
+ * an env var and use the first env-ordered value as the auth entry, following
+ * the models.dev convention that `env[0]` is the primary credential. Legacy
+ * single-credential payloads (`apiKey`) keep today's auth-only behaviour.
+ */
+export const resolveCloudProviderCredentials = (
+  provider: Pick<
+    DenOrgLlmProviderConnection,
+    "apiKey" | "apiKeys" | "providerConfig"
+  >,
+) => {
+  const apiKeys = provider.apiKeys ?? {};
+  const envNames = getCloudProviderEnv(provider.providerConfig);
+  const orderedNames = [
+    ...envNames.filter((name) => name in apiKeys),
+    ...Object.keys(apiKeys).filter((name) => !envNames.includes(name)),
+  ];
+  const envEntries = orderedNames.flatMap((name) => {
+    const value = apiKeys[name]?.trim();
+    return value ? [{ key: name, value }] : [];
+  });
+  const primaryApiKey = provider.apiKey?.trim() || envEntries[0]?.value || "";
+  return { envEntries, primaryApiKey };
+};
+
 export const getCloudManagedProviderId = (
   provider: Pick<DenOrgLlmProvider, "id" | "providerId" | "source">,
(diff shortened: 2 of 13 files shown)


benjaminshafii:
## Manual E2E validation — dashboard + API (all 6 items pass)

Tested this branch locally on macOS: fresh worktree of `cloud-providers-multiple-env-vars` (b329a15), `pnpm install`, stack via `DEN_API_PORT=8790 pnpm dev:web-local` (Docker MySQL + drizzle push + den-api:8790 + den-web:3005), `pnpm dev:den:seed-demo`, signed in as the seeded admin `alex@acme.test`, drove the real dashboard through Chrome/CDP.

### 1. Multi-env catalog provider (Amazon Bedrock) — ✅
Add Provider → Catalog → Amazon Bedrock renders **4 labeled password inputs** (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, `AWS_BEARER_TOKEN_BEDROCK`) instead of the single "API key / credential" field. Filled 2 of 4, picked a model, saved → `POST /v1/llm-providers 201`.

Bedrock four env inputs

### 2. Saved badges + server-side merge on edit — ✅
Reopening the editor shows "Saved" badges + "Leave blank to keep current value" placeholders for exactly the 2 filled keys; the other 2 keep "Paste the … value". Filled only `AWS_REGION`, saved (`PATCH 200`), reopened → 3 "Saved" badges, and the connect endpoint later returned all three original values intact (blank ≠ wipe).

Two saved badges
Three saved badges after merge

Note: the first PATCH hit the expected re-auth guard (403 → "Sign in again to continue" modal → verify password → 200). Not a bug.

### 3. Single-env regression (Anthropic) — ✅
Classic single credential field. Saved with a key; reopen shows the "Existing credential saved" chip; saving again with the field blank keeps the credential — list endpoint still reports `hasApiKey: true` and connect returns the original key.

Anthropic single field
Existing credential chip

### 4. Guided custom form ↔ JSON — ✅
Guided form shows the single field with the auto-derived env. Setting `"env": ["FOO_KEY", "FOO_REGION"]` in JSON mode live-switches the credential section to two labeled inputs, and **"Use the guided form instead" now works on that multi-env JSON** — returns to the form with both env inputs, provider id, base URL, and model IDs preserved.

Guided single field
JSON mode two env inputs
Guided form after switch-back

Pre-existing caveat (not a regression of this PR): opening "Advanced: edit as JSON" *before* filling base URL/models generates the placeholder template whose example model has `name ≠ id`, which is itself not form-representable — switch-back bails with the "advanced fields" hint regardless of env (`readGuidedCustomProviderFields`, name/id check).

### 5. API contract — ✅
- Bedrock `GET /v1/llm-providers/:id/connect` → `"apiKey": null`, `"apiKeys": {"AWS_ACCESS_KEY_ID": "...", "AWS_SECRET_ACCESS_KEY": "...", "AWS_REGION": "..."}` — exactly the filled names/values; the unfilled `AWS_BEARER_TOKEN_BEDROCK` is absent.
- Anthropic connect → `"apiKey": "sk-ant-…"`, `"apiKeys": null`.
- `GET /v1/llm-providers?scope=manageable` → no `apiKey` values anywhere; `configuredEnvKeys` lists the right names for all providers, including legacy single-env rows mapping to their one env key.

### 6. Validation — ✅
```
PATCH /v1/llm-providers/:bedrockId  body: { …full write body…, "apiKeys": { "TOTALLY_BOGUS_ENV": "x" } }
→ 400 {"error":"invalid_api_keys","message":"TOTALLY_BOGUS_ENV is not one of this provider's env keys (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_BEARER_TOKEN_BEDROCK)."}
```
Provider state unchanged afterwards (connect still returns the 3 configured keys).

### Unit tests run
| Command | Result |
| --- | --- |
| `ee/apps/den-web` → `bun test tests/llm-provider-guided.test.ts` | 17 pass, 0 fail |
| `ee/apps/den-api` → `bun test test/provider-credentials.test.ts test/llm-custom-provider-guided.test.ts` | 18 pass, 0 fail |
| `apps/app` → `bun test tests/cloud-provider-credentials.test.ts tests/cloud-provider-reimport.test.ts` | 11 pass, 0 fail |

No console errors during the flows; server log only shows unrelated local-dev `stripe_secret_key_missing` noise. Desktop connect flow not covered (out of scope — dashboard + API only).

PR #2432, opened 2026-07-02
feat(den-web): ask for the credential before the endpoint in the guided provider form

benjaminshafii:
Follow-up to #2431. In the guided custom-provider form, the credential is now asked for **before** the endpoint, and lives inline in the form (the separate Credential section remains for catalog providers and the JSON escape hatch).

Why: the endpoint probe needs URL + key. With the key below the endpoint (in a section further down the page), filling the form top-to-bottom left the probe idle with a "enter the API key below" hint pointing past the models the admin was about to pick. Now the natural top-to-bottom fill (name → id → key → URL) makes the probe fire the moment the URL is typed, and the model picker appears directly beneath it.

Changes:
- Inline `API key / credential` field between Provider ID and Base URL in the guided form (same `apiKey` state, same keep-current-credential placeholder on edit).
- Standalone Credential section hidden in guided form mode only.
- Idle hint updated ("above").
- The Azure Foundry fraimz flow now asserts the key field precedes the Base URL field in the DOM.

## Validation

- `tsc --noEmit` clean (den-web).
- Re-ran the end-to-end fraimz against a Daytona Den stack running this change and the **real** customer Azure resource:

```
pnpm fraimz --flow llm-provider-azure-foundry --cdp-url http://127.0.0.1:9333
```

Frame `01-probe-deployments` shows the new order (Provider ID → API key → Base URL) with the probe already green ("Endpoint reachable — 1 model available."); the remaining frames prove picker + verify-on-save + persisted `@ai-sdk/openai` config are unchanged.

### DIFF
--- a/ee/apps/den-web/app/(den)/dashboard/_components/llm-provider-editor-screen.tsx
+++ b/ee/apps/den-web/app/(den)/dashboard/_components/llm-provider-editor-screen.tsx
@@ -871,6 +871,26 @@ export function LlmProviderEditorScreen({
 
                         <label className="grid gap-3">
                             <span className="text-[14px] font-medium text-gray-700">
+                                API key / credential
+                            </span>
+                            <DenInput
+                                type="password"
+                                value={apiKey}
+                                onChange={(event) => setApiKey(event.target.value)}
+                                placeholder={
+                                    provider?.hasApiKey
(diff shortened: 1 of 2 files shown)


benjaminshafii:
Folding this into the re-land on top of #2430 (per team chat).

PR #2442, opened 2026-07-03
fix(app): fetch live LM Studio models in picker and settings

### LINKED ISSUE #2364: [Bug]: LM Studio showing wrong models always
### Summary

Image

**It should show these models instead**

Image

### To Reproduce

Go to model
Connect to LM Studio
It will give you three models as mentioned above that's looks hardcoded or from LMS

### Expected behavior

It should fetch current models from LMS

### Actual behavior

It should fetch current models from LMS:

From local LMS APIs: http://127.0.0.1:1234

### Screenshots (optional)

Image

Image

latest

_No response_

taneja-garvit:
Fixes #2364 by replacing the static models.dev catalog with models from the local LM Studio server when the provider is connected. Adds an LM Studio settings extension (mirroring Ollama) and unit tests for parsing and provider-list enrichment.

## Summary
- Fixes #2364 — LM Studio model picker showed 3 hardcoded catalog models instead of the user's downloaded models
- Fetches live models from `http://127.0.0.1:1234/api/v0/models` (fallback: `/v1/models`)
- Replaces static catalog in the **model picker** when LM Studio is connected (addresses UX gap noted on #2367)
- Adds LM Studio settings extension panel (Ollama-style) and syncs all live models into opencode config on install

## Related
Builds on the approach i with model-picker integration, stale-selection handling, embedding filtering on fallback, and unit tests.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm --filter @openwork/app exec bun test tests/lmstudio-models.test.ts`
- [ ] With LM Studio running on :1234, open Models → LM Studio shows live models (e.g. gemma), not gpt-oss/qwen3
- [ ] Settings → Extensions → LM Studio → refresh lists live models and Add to workspace works

## Comments
Working on #2364 in #. Related to #2367 - this also updates the model picker flow from the bug report, not only the extensions panel

### DIFF
--- a/apps/app/public/ext-lmstudio.svg
+++ b/apps/app/public/ext-lmstudio.svg
@@ -1,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" role="img" fill="none">
+  <title>LM Studio</title>
+  <rect width="24" height="24" rx="5" fill="#111827"/>
+  <text x="12" y="15.5" text-anchor="middle" font-family="Segoe UI, system-ui, sans-serif" font-size="7.5" font-weight="700" fill="#ffffff">LM</text>
+</svg>
--- a/apps/app/src/app/extensions.ts
+++ b/apps/app/src/app/extensions.ts
@@ -353,4 +353,31 @@ export const BUILT_IN_OPENWORK_EXTENSION_MANIFESTS: OpenWorkExtensionManifest[] 
     ],
     lifecycle: { reload: ["config"], detection: ["provider:ollama"] },
   },
+  {
+    schemaVersion: 1,
+    id: "lmstudio",
+    name: "LM Studio",
+    description: "Local model provider at http://127.0.0.1:1234.",
+    source: { format: "openwork-builtin", origin: "builtin", trusted: true },
+    icon: { src: "/ext-lmstudio.svg" },
+    composer: { prompt: "Use the LM Studio extension to " },
+    setup: {
+      instructions:
+        "Start LM Studio's local server (Developer tab), choose a downloaded model, then add it as an OpenCode provider.",
+      primaryCta: "Add LM Studio model",
+    },
+    resources: [
+      { type: "local-service", id: "lmstudio-api", label: "LM Studio API", description: "http://127.0.0.1:1234", required: true },
+      { type: "provider", id: "lmstudio", providerId: "lmstudio", packageName: "@ai-sdk/openai-compatible", required: true },
+    ],
+    contributions: [
+      { type: "settings-panel", ref: "openwork.lmstudio.settings", location: "settings-detail" },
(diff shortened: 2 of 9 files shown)


taneja-garvit:
Thanks for the thorough review on #2367 - this PR intentionally addresses a few of those points:

- Model picker path (#2364 / @Iflaqbhat):** When LM Studio is connected, `fetchProviderList` enriches the provider list with live models from `127.0.0.1:1234` so the Models modal shows downloaded models instead of the static catalog.
- Stale selection (@Pablosinyores):** `reconcileLMStudioSelectedModel()` resets selection when the chosen model disappears after refresh.
- Embedding fallback:** OpenAI-compatible `/v1/models` responses filter likely embedding model ids by name.

Pablosinyores (commented):
Reviewed the model-fetch and enrichment logic. Solid — it closes both gaps I raised on #2367: embeddings are filtered on the native path by exact \`type === "embeddings"\` and on the \`/v1\` fallback by the id heuristic, and the config panel reconciles the selection via \`reconcileLMStudioSelectedModel\` when the live list changes. Writing every live model through \`allModelIds\` (rather than a single id) is the right call too. Tests cover native parse, fallback filtering, embedding detection, reconcile, and the 404 fallback path.

One thing worth the maintainer noting: this overlaps directly with #2367 (smundhra-git), which targets the same "live LM Studio models" gap. This PR is the superset (adds the settings panel + fallback filtering + all-models write), so probably only one of the two should land — flagging so they do not both get merged.

Comments:

- [medium] `fetchProviderList` now `await`s `enrichProviderListWithLiveLMStudioModels` inline, so a provider-list load blocks on the LM Studio fetch whenever the provider is connected. The common offline case fails fast (connection refused), but a connected-but-hung server can add up to ~6s (3s native + 3s fallback timeout) to the list load, which gates the model picker UI. Consider enriching in the background (or a shorter timeout) so a stalled LM Studio does not stall provider loading.
- [minor] `buildLMStudioProviderModelsRecord` returns the existing models unchanged when the static catalog has no first entry to use as a template, so live models are silently dropped if the lmstudio catalog is ever empty. Works today because the catalog ships models, but a minimal fallback template (`{ id, name }`) would harden it.
- [nit] `isLikelyEmbeddingModelId` checks `embed || embedding || nomic-embed`; the first already subsumes the other two.

Nice work — this is close.

taneja-garvit:
Thanks for the thorough review! @Pablosinyores Makes sense on all three points:

- Blocking enrichment: agreed, will move this to background (or drop the timeout) so a hung LM Studio doesn't gate the provider list load.
- Empty catalog fallback: good catch, will add a minimal `{ id, name }` template so live models aren't silently dropped.
- Redundant embed check: will simplify to just the `embed` substring check.

Will push a follow-up commit addressing these shortly. Also noted your flag on the overlap with #2367 - happy to coordinate however the maintainer wants to handle that.

PR #2448, opened 2026-07-03
Redesign sign-in/sign-up screen; route desktop "Sign in" to sign-up tab

codemanshan:
## What & why

When a user clicks **"Sign in"** in the desktop app they land on the web login already forced into sign-in mode (`?mode=sign-in`). A **new** user then had no obvious way to create an account — the switch was buried at the bottom — so people got stuck. The screen also had a lot going on (animated hero + feature cards + an extra email-first step).

This PR fixes the trap and gives the login a cleaner, standard flow.

## Changes

**Web login (`ee/apps/den-web`)**
- Always-visible **Sign in / Create account** toggle at the top of the form — one tap to switch, so nobody gets stuck. `?mode=` now just preselects a tab instead of locking it.
- Removed the redundant **email-first sign-in step** (`continueSignInWithEmail`). SSO enforcement is unchanged — `submitAuth` and `beginSocialAuth` already call `redirectToRequiredSso` on every path, before any credential is accepted.
- **GitHub + Google** now show in both modes (GitHub was previously hidden in sign-in mode).
- Rebuilt as a single split card: gradient/brand panel (2/3) + form (1/3). On **mobile** the gradient is hidden and only the form shows, with an OpenWork logo header.
- Card height stays fixed when toggling modes. Removed now-unused toggle-copy props on the invite/claim screens.

**Desktop app (`apps/app`)**
- The standalone **"Sign in"** buttons (full-screen sign-in, chat status bar, Skills page) now open the **sign-up** tab (label unchanged), so new users aren't defaulted into sign-in; returning users just tap the toggle. Screens that already show a separate "Create account" button (Settings → Cloud, compact strip) are unchanged.

## Safety / testing
- **SSO enforcement verified preserved** across email/password, GitHub, and Google paths (plus the existing server-side backstop in `den-api`).
- OTP verification, password reset, invite, workspace-claim, and desktop-handoff flows all intact.
- Type-checks pass for both `den-web` and `apps/app`. No automated UI tests exist for this screen, so verification was type-check + two independent code reviews + a live click-through in the real Electron desktop app.

## Note
This branch spans **two areas** (web login + desktop buttons) on purpose — they're one change: the desktop buttons only make sense now that the web page has the toggle.

### DIFF
--- a/apps/app/src/react-app/domains/cloud/den-signin-surface.tsx
+++ b/apps/app/src/react-app/domains/cloud/den-signin-surface.tsx
@@ -319,7 +319,10 @@ export function DenSignInSurface(props: DenSignInSurfaceProps) {
               <button
                 type="button"
                 className="inline-flex h-12 w-full items-center justify-center gap-2 rounded-full bg-dls-accent text-sm font-semibold text-[var(--dls-accent-fg)] transition-all hover:bg-[var(--dls-accent-hover)] disabled:opacity-60 disabled:cursor-not-allowed"
-                onClick={() => props.onOpenBrowserAuth("sign-in")}
+                // Opens on the sign-up tab (label stays "Sign in"): most
+                // people hitting this are new, and the web page's tabs let
+                // returning users switch to sign-in in one tap.
+                onClick={() => props.onOpenBrowserAuth("sign-up")}
                 disabled={props.authBusy || props.sessionBusy}
               >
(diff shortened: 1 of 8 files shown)


kilicrukiye784-lab:
3 Tem 2026 Cum 22:40 tarihinde ben ***@***.***> şunu yazdı:

PR #2467, opened 2026-07-04
fix(app): use resolved Shiki theme for code blocks

### LINKED ISSUE #2440: Code block dark mode bug - Shiki theme hardcoded to github-light
## Describe the bug

When using Dark theme, code blocks inside assistant messages are rendered with Shiki's `github-light` theme. The code syntax colors are from the light palette (e.g. `#d73a49` for keywords), which are unreadable on a dark background.

## To Reproduce

1. Switch OpenWork to Dark theme (Settings > Appearance)
2. Send any message containing a code block
3. Observe the code tokens use light colors (low contrast on dark bg)

## Root cause

In `apps/app/src/components/markdown/markdown.tsx` line ~172, Shiki's theme is hardcoded to `"github-light"`:
```ts
theme: "github-light",
```
It should be dynamic based on the app's current theme (`github-dark` in dark mode).

## Workaround confirmed

Injecting `filter: invert(1) hue-rotate(180deg)` on `.shiki.github-light` elements via DevTools fixes the rendering.

## Suggested fix

```diff
- theme: "github-light",
+ theme: document.documentElement.classList.contains("dark") ? "github-dark" : "github-light",
```

## Environment

- OpenWork v0.17.x
- Windows
- Dark theme

taneja-garvit:
Fixes #2440

## Summary
- use the resolved app theme for Shiki code highlighting instead of hardcoding `github-light`
- fix unreadable code blocks in dark mode
- update shared markdown/code rendering paths so code blocks refresh when theme changes

Fixes #2440

## Test plan
- [x] Switch to dark mode and render a code block in an assistant message
- [x] Confirm syntax colors use the dark Shiki palette
- [x] Switch back to light mode and confirm code blocks update correctly
- [x] `pnpm typecheck`
- [x] `pnpm --filter @openwork/app exec bun test tests/theme.test.ts`

### DIFF
--- a/apps/app/src/app/theme.ts
+++ b/apps/app/src/app/theme.ts
@@ -1,5 +1,6 @@
 export type ThemeMode = "light" | "dark" | "system";
 export type ResolvedThemeMode = "light" | "dark";
+export type ShikiTheme = "github-light" | "github-dark";
 
 const THEME_PREF_KEY = "openwork.react.settings.theme-mode";
 const LEGACY_THEME_PREF_KEYS = ["openwork.themePref"];
@@ -95,6 +96,9 @@ export const bootstrapTheme = () => {
 export const getInitialThemeMode = () => getCurrentMode();
 
 export const getResolvedThemeMode = () => resolveMode(getCurrentMode());
+
+export const getResolvedShikiTheme = (): ShikiTheme =>
+  getResolvedThemeMode() === "dark" ? "github-dark" : "github-light";
 
 const persistThemeMode = (mode: ThemeMode) => {
   if (typeof window === "undefined") return;
--- a/apps/app/src/components/markdown/markdown.tsx
+++ b/apps/app/src/components/markdown/markdown.tsx
@@ -388,7 +388,7 @@ function MarkdownBlockInner({
     }
     return sanitizeMarkdownHtml(markdownParser.parse(text, { async: false }));
   }, [text]);
-  const [highlightedHtml, setHighlightedHtml] = useState<{ text: string; html: string } | null>(null);
+  const [highlightedHtml, setHighlightedHtml] = useState<{ text: string; theme: string; html: string } | null>(null);
 
   useEffect(() => {
     if (streaming || !hasFencedCodeBlock(text)) {
@@ -401,7 +401,7 @@ function MarkdownBlockInner({
       const sanitizedHtml = sanitizeMarkdownHtml(html);
 
       if (!cancelled && sanitizedHtml.trim()) {
-        setHighlightedHtml({ text, html: sanitizedHtml });
+        setHighlightedHtml({ text, theme: shikiTheme, html: sanitizedHtml });
       }
     }).catch(() => {
       if (!cancelled) {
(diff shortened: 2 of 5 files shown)


Pablosinyores (commented):
Read through this and the approach is right. `useSyncExternalStore(subscribeToTheme, getResolvedShikiTheme)` is the correct primitive for pulling theme into render without a hand-rolled listener, and adding `shikiTheme` to the highlight effect deps in both `MarkdownBlockInner` and `CodeBlockCode` is what actually makes blocks re-highlight on toggle. The `theme ?? resolvedTheme` fallback in `CodeBlockCode` keeps explicit-theme callers working while defaulting everything else to the resolved app theme — clean.

One minor thing in `markdown.tsx`: the cached-HTML shortcut

    const html = !streaming && highlightedHtml?.text === text ? highlightedHtml.html : syncHtml;

is keyed on `text` only, not on theme. On a light/dark toggle where the text is unchanged, `highlightedHtml.text === text` stays true, so the previously highlighted (old-theme) HTML shows until the effect's async re-highlight resolves and replaces `highlightedHtml`. It self-corrects because the effect now depends on `shikiTheme`, so it is a brief flash rather than a stuck state — just flagging in case you want to also gate that memo on theme to kill the transient. Not a blocker.

Mapping resolved light/dark to github-light/github-dark is exactly the right scope for #2440. Nice fix.

taneja-garvit:
Good catch - makes sense, the memo should be theme-aware too so we don't even get the transient. Pushing a fix now to also gate highlightedHtml on shikiTheme

taneja-garvit:
@Pablosinyores Pushed a small follow-up for this. I made the cached highlighted HTML theme-aware as well, so we don’t briefly reuse old-theme Shiki output during a light/dark toggle.

Pablosinyores (commented):
Good fix for #2440 — the core problems (hardcoded `github-light`, and cached HTML surviving a theme switch) are both addressed cleanly. Two things I like: driving `getResolvedShikiTheme` through `useSyncExternalStore(subscribeToTheme, ...)` makes the highlight reactive to theme changes instead of stale-on-mount, and adding `theme` to the cache record plus the `highlightedHtml.theme === shikiTheme` guard means a theme switch now correctly invalidates the memoized HTML rather than showing the old colors. The `useState` shape change to carry `theme` alongside `text`/`html` is the right way to key that.

One thing to double-check: `code-block.tsx` drops the `theme = "github-light"` default and makes `theme` a bare prop. Any call site that was relying on that default now passes `undefined` to Shiki. The markdown paths in this PR pass `getResolvedShikiTheme()`, but it's worth a quick grep for other `<CodeBlock>` / `codeToHtml` callers to confirm none are left handing in `undefined` (which would either throw or silently fall back inside Shiki). If code-block is only ever used through the markdown surfaces you updated, this is moot — just want to make sure removing the default doesn't leave a caller unthemed.

Test plan checkboxes in the description are still unchecked — assuming you've run the light/dark switch manually, might be worth ticking them so a reviewer knows it was exercised.

taneja-garvit:
Automated checks pass. I haven’t manually verified the light/dark toggle in the desktop app yet - happy to do that if needed before merge.

PR #2506, opened 2026-07-06
feat(app): copy/export sanitized diagnostics from the command palette

benjaminshafii:
## What

Sharing logs when something goes wrong is now one step: **Cmd/Ctrl+K → "logs"** offers **Copy diagnostics** and **Export diagnostics** — no Developer Mode, no digging through Settings → Debug. Both produce the same sanitized bundle: app build info, engine + server state (including the in-memory stdout/stderr tails), and recent dev/perf log tails.

## How

- New shared builder `apps/app/src/app/lib/diagnostics-bundle.ts`: pure `composeDiagnosticsBundleJson` core + async gatherer over the existing desktop IPC (`appBuildInfo`/`engineInfo`/`openworkServerInfo`).
- **Sanitization is two-layer**: IPC fields are allowlisted (`clientToken`/`ownerToken`/`hostToken`/`opencodePassword`/`opencodeUsername` never enter the bundle), and every known secret *value* is scrubbed from the serialized output — so a token leaking into a stdout tail still comes out `[redacted]`. Settings are recorded as `tokenPresent: true|false` only.
- Settings → Config "Diagnostics bundle" now reuses the same builder (the old inline builder is deleted); `downloadTextAsFile` moved to shared `app/lib/download.ts`.
- Behavior note: dev/perf log tails are now included in the bundle regardless of Developer Mode — that is the point of the feature (non-technical users share useful logs). Tails were already capped (8KB per stream) at the source.

## Spec & proof (demo-driven)

Approved voice-over: `evals/voiceovers/share-diagnostics.md` (4 frames), enforced by the coded flow `evals/flows/share-diagnostics.flow.mjs`. Frame 4 is non-vacuous: it fetches the live server tokens over the preload bridge and asserts none of them appear in the copied bundle (fails if the server is running with no secrets found).

## Tests run

- `pnpm --filter @openwork/app exec tsc --noEmit` — pass
- `pnpm --filter @openwork/app test` — 123 pass / 0 fail (includes new `tests/diagnostics-bundle.test.ts` covering value scrubbing incl. secrets planted in stderr tails, field-name exclusion, and the no-desktop-info case)
- `pnpm fraimz --flow share-diagnostics --cdp-url http://127.0.0.1:9826` — **PASSED, 4/4 frames** against a fresh isolated local Electron (dedicated Vite port/userdata/data dir; renderer provenance verified). Local run, not Daytona. Frame-by-frame proof comment follows below.


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-06T16-50-56-910Z`

Full frame proof with validated screenshots: `evals/results/2026-07-06T16-50-56-910Z/fraimz.html` (re-run: `pnpm fraimz --flow share-diagnostics`)

### ✅ share-diagnostics — Users can share sanitized diagnostics from the command palette
_User-facing flow demo_

1. **Copy diagnostics is available from Cmd/Ctrl+K search**
   > 🎙 Something in my workspace isn't behaving. I press Cmd+K and type "logs" — the palette offers Copy diagnostics right there, no developer mode, no digging through settings.
   - ✅ Visible text includes "Copy diagnostics"
   - 📸 `share-diagnostics-01-frame-1.png` — 4 validations passed
2. **Copy diagnostics puts a sanitized bundle on the clipboard**
   > 🎙 I pick it, a toast confirms, and the diagnostics bundle is on my clipboard — app version, engine and server state, recent log tails — ready to paste into an issue or a support thread.
   - 📸 `share-diagnostics-02-frame-2.png` — 4 validations passed
3. **Export diagnostics saves the same kind of bundle as a JSON file**
   > 🎙 Back in the palette, Export diagnostics saves the same bundle as a JSON file, for when a maintainer asks for an attachment.
   - 📸 `share-diagnostics-03-frame-3.png` — 4 validations passed
4. **Diagnostics record token presence without exposing token values**
   > 🎙 And it's safe to share by default: the bundle records whether a connection token exists — yes or no — never the value itself.
   - 📸 `share-diagnostics-04-frame-4.png` — 4 validations passed
   - ✅ Script frame 1 narrated: "Something in my workspace isn't behaving. I press Cmd+K and type \"logs\" — the palette of"
   - ✅ Script frame 2 narrated: "I pick it, a toast confirms, and the diagnostics bundle is on my clipboard — app version"
   - ✅ Script frame 3 narrated: "Back in the palette, Export diagnostics saves the same bundle as a JSON file, for when a"
   - ✅ Script frame 4 narrated: "And it's safe to share by default: the bundle records whether a connection token exists "


_(screenshot upload failed: BLOB_READ_WRITE_TOKEN is not set — fetch it with the get-env-var skill: export BLOB_READ_WRITE_TOKEN="$(infisical secrets get BLOB_READ_WRITE_TOKEN --plain --silent)" — images available in the run directory)_

PR #2480, opened 2026-07-04
feat(den): org install links — one generic installer, stamped per org at serve time

benjaminshafii:
> **Stacked on #2481** (`feat/org-capability-flags`) — merge that first. This PR now ships **capability-gated from day one**: install links are **dark for every org** until a platform admin enables the `installLinks` capability in `/admin`.

## What

PR-2 of the invite-to-desktop track (PR-1: #2471). An org admin copies an **install link** from the Members page; the link serves the **one generic installer binary, stamped per org at serve time** — zero per-org builds. The installer writes `desktop-bootstrap.json` (`requireSignin: true`, org's server URLs), so the desktop app's **first boot is the forced sign-in against the right deployment** instead of an empty local-first vessel. A completely bare installer doesn't refuse to run anymore — it asks for the install link (paste UI / `--install-link`).

### Pieces
- **capability gate (new, rides #2481)** — `POST /v1/orgs/:id/install-links` returns `403 { error: "capability_disabled", capability: "installLinks" }` unless the org's `capabilities.installLinks` is on (OpenAPI 403 schema updated); den-web renders **Copy install link** only when `/v1/org` reports the capability. Public token endpoints stay ungated — tokens only exist after a successful mint.
- **`packages/install-config`** — shared zod contract + filename-tag/config-URL helpers (used by den-api + installer)
- **den-api** — `InstallLinkTable` (sha256-hashed tokens, rotate-on-mint, revocable); `POST /v1/orgs/:id/install-links` (admin); `GET /v1/install-config?token=` (public, rate-limited); `GET /v1/install/:platform?token=`:
  - **mac**: zip with a stored `openwork-installer.json` sidecar appended — dependency-free `zip-append` util; existing members stay **byte-identical** (unit test extracts with system `unzip` and sha-compares)
  - **windows**: exe bytes untouched; org rides the `Content-Disposition` filename (`OpenWork-Installer--<host>--<token>.exe`)
  - **linux**: inspectable setup script (writes the XDG bootstrap config, points at the AppImage)
  - Artifacts resolve from `OPENWORK_INSTALLER_ARTIFACTS_DIR` (dev/eval); the **public generic release artifact ships in PR-3** — until then these endpoints 503 cleanly in prod
[...]

PR #2519, opened 2026-07-07
chore(ci): add manual den-api redeploy workflow

benjaminshafii:
## What

Adds `.github/workflows/den-api-deploy.yml` — a `workflow_dispatch`-only workflow that triggers a Render deploy for the production den-api service and polls until it's live, with no other side effects.

## Why

`den-api-env.yml` already triggers a deploy, but only as a side effect of setting an env var (it requires `key`/`value` inputs). Using it just to force a redeploy would mean mutating a production env var I don't intend to change. This adds a dedicated, minimal-diff workflow that reuses the exact same proven trigger-and-wait `curl`/Render API logic, with zero env var mutation.

## Context

Production den-api is currently running a build from before PR #2480 (org install-links) — confirmed via `/v1/app-version` reporting `0.17.9`, which predates that commit by 55 commits (verified with `git merge-base --is-ancestor`). This workflow lets us redeploy it to current `dev` without touching config.

## Tests run

- `python3 -c "import yaml; yaml.safe_load(...)"` — valid YAML.
- `actionlint` — clean except the expected unknown-custom-runner-label warning for `blacksmith-4vcpu-ubuntu-2204`, which is the same self-hosted runner label already used by `den-api-env.yml` in this repo.
- No app/runtime code touched — pure CI workflow addition. Real validation is running it for real against the actual Render service (planned right after merge).

### DIFF
--- a/.github/workflows/den-api-deploy.yml
+++ b/.github/workflows/den-api-deploy.yml
@@ -1,0 +1,76 @@
+name: Den API Deploy
+
+# Triggers a fresh deploy of the production den-api service (Render) from the
+# latest commit on its tracked branch. No env vars are changed — use
+# den-api-env.yml for that. Use this when production den-api has fallen
+# behind dev (e.g. after merges) and needs a manual redeploy.
+#
+# Required GitHub Actions secrets:
+#   RENDER_API_KEY                        Render API key
+#   RENDER_DEN_CONTROL_PLANE_SERVICE_ID   Render service id of den-api
+
+on:
+  workflow_dispatch: {}
+
+permissions:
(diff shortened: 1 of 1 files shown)


benjaminshafii:
Closing: this workflow solved the wrong problem. Render auto-deploy exists and has been triggering — the deploys were crash-looping on boot (ERR_MODULE_NOT_FOUND for @openwork/install-config/dist) so Render kept serving the last healthy pre-#2480 build. A manual redeploy would have crashed identically. Actual fix: #2520 (one line in den-api's build.mjs).

PR #2521, opened 2026-07-07
fix(den-web): reauth dialog now opens for every privileged action — detect step-up at the transport layer

benjaminshafii:
## What

Step-up auth ("confirm it's you") was architecturally right — `runReauthableAction` queues the privileged action, opens the security-check dialog, and auto-retries after verification — but **26 of 72 privileged call sites never triggered it**. They threw plain `Error`s instead of the `ReauthRequiredError` the provider detects, so for them (including Members → **Copy install link**) the 403 `{error:"reauth"}` response surfaced as a dead inline banner: *"For security, confirm it's you before changing workspace settings."* — with no dialog, no queued action, no way forward.

Two changes:

1. **`requestJson` now converts the 403-reauth shape into `ReauthRequiredError` at the transport layer** (`den-flow.ts`, +5 lines). One choke point; every current and future call site becomes reauth-aware regardless of how it throws. The 26 non-conforming throw sites become unreachable for reauth responses — no per-site rewrites.
2. **`retryReauthMutation` re-queues on repeat-reauth** (`org-dashboard-provider.tsx`, +5 lines): if the retried action *still* lacks a fresh session, the dialog reopens instead of rejecting the queued promise.

## Caller audit

Every `requestJson` caller in den-web was audited for the new throw: auth endpoints (`/api/auth/*`) and public routes never produce the `{error:"reauth"}` shape; all privileged dashboard surfaces are either wrapped by `runReauthableAction` or catch locally with no closure-level catch swallowing the error before the provider sees it. (Full table in the executor audit; no call-site changes were needed.)

## Proof — new den-stack eval `den-reauth-pending-action` (PASSED, frames below)

Real browser against the local den stack (MySQL + den-api + den-web + seeded Acme demo):
1. Sign in as the demo owner; age the session by 1h via SQL (server freshness window is 15 min); enable the `installLinks` org capability.
2. Members → click **Copy install link** → the **security-check dialog appears** and the raw server message is asserted NOT to render outside the dialog (this exact step dead-ended before the fix).
3. Type the password → dialog closes, the queued action completes **without another click**, button flips to Copied, and the clipboard is asserted to contain the real `/install?token=…` URL.

```
▶ den-reauth-pending-action — Den reauth keeps the pending install-link action alive
  ✓ Signing in and staging a stale session
  ✓ A privileged click opens the security check instead of a dead banner
  ✓ Verify password and the queued action completes on its own
  ✓ Voice-over script coverage
```

Also run: `pnpm --dir ee/apps/den-web exec tsc --noEmit` (pass), `pnpm --dir ee/apps/den-web build` (pass).

Repro:
```sh
pnpm dev:den:mysql && pnpm dev:den:db-push
DEN_WEB_PORT=3105 pnpm dev:den:api &
pnpm dev:den:seed-demo
WATCHPACK_POLLING=true DEN_WEB_PORT=3105 pnpm dev:den:web &
# Chrome with --remote-debugging-port=<cdp>, then:
OPENWORK_EVAL_DEN_WEB_URL=http://localhost:3105 OPENWORK_EVAL_DEN_MYSQL_CONTAINER=openwork-web-local-mysql \
  pnpm fraimz --flow den-reauth-pending-action --cdp-url http://127.0.0.1:<cdp>
```

Companion PR (in flight): popup-based verification for Google/GitHub/SSO so the queued action also survives OAuth — today the social path navigates the page away and loses it.

### DIFF
--- a/ee/apps/den-web/app/(den)/_lib/den-flow.ts
+++ b/ee/apps/den-web/app/(den)/_lib/den-flow.ts
@@ -1085,5 +1085,10 @@ export async function requestJson(path: string, init: RequestInit = {}, timeoutM
     }
   }
 
+  const reauthError = getReauthRequiredError(payload, response);
+  if (reauthError) {
+    throw reauthError;
+  }
+
   return { response, payload, text };
 }
--- a/ee/apps/den-web/app/(den)/dashboard/_providers/org-dashboard-provider.tsx
+++ b/ee/apps/den-web/app/(den)/dashboard/_providers/org-dashboard-provider.tsx
@@ -243,6 +243,11 @@ export function OrgDashboardProvider({
       await executeReauthableAction(pending.label, pending.action);
       pending.resolve();
     } catch (error) {
+      if (isReauthRequiredError(error)) {
+        setPendingReauthMutation(pending);
+        return;
+      }
+
       pending.reject(error);
     }
   }
--- a/evals/flows/den-reauth-pending-action.flow.mjs
+++ b/evals/flows/den-reauth-pending-action.flow.mjs
@@ -1,0 +1,353 @@
+import { execFile } from "node:child_process";
+import { promisify } from "node:util";
+import { loadVoiceoverParagraphs } from "../runner/voiceover.mjs";
+
+const FLOW_ID = "den-reauth-pending-action";
+const vo = await loadVoiceoverParagraphs(FLOW_ID);
+const execFileAsync = promisify(execFile);
+
+const DEMO_EMAIL = "alex@acme.test";
+const DEMO_PASSWORD = "OpenWorkDemo123!";
+const SIGN_IN_ROUTE = "/";
+const MEMBERS_ROUTE = "/dashboard/members";
+const COPY_INSTALL_LINK_SELECTOR = '[data-testid="copy-install-link"]';
+const RAW_REAUTH_MESSAGE = "For security, confirm it's you before changing workspace settings.";
+
+const state = {
+  clipboardText: "",
+};
+
+function routeUrl(ctx, path) {
+  return new URL(path, ctx.env.OPENWORK_EVAL_DEN_WEB_URL).toString();
+}
+
+function mysqlContainer(ctx) {
+  return ctx.env.OPENWORK_EVAL_DEN_MYSQL_CONTAINER || "openwork-web-local-mysql";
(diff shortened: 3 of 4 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-07T03-15-51-900Z`

Full frame proof with validated screenshots: `evals/results/2026-07-07T03-15-51-900Z/fraimz.html` (re-run: `pnpm fraimz --flow den-reauth-pending-action`)

### ✅ den-reauth-pending-action — Den reauth keeps the pending install-link action alive
_User-facing flow demo_

1. **Alex is signed in with a stale privileged session**
   > 🎙 Alex is an org admin whose sign-in is more than fifteen minutes old. We sign in as Alex, deliberately age the session in the eval database, and keep the dashboard visible so the account still feels signed in.
   - ✅ The signed-in dashboard is visible
   - 📸 `den-reauth-pending-action-01-reauth-stale-session-dashboard.png` — 5 validations passed

   

2. **A stale privileged click opens the security check instead of a dead banner**
   > 🎙 From Members, Alex clicks Copy install link. Instead of a dead error banner, OpenWork opens the security check and clearly asks Alex to confirm it is him before changing workspace settings.
   - ✅ The reauth dialog appears for the stale privileged action
   - ✅ The raw server reauth message is not rendered as a page banner outside the dialog
   - 📸 `den-reauth-pending-action-02-reauth-security-check-dialog.png` — 4 validations passed

   

3. **After password verification, the queued copy action completes automatically**
   > 🎙 Alex enters his password once. The dialog closes, OpenWork retries the original copy action automatically, and the button flips to Copied with the install URL already on the clipboard.
   - ✅ The dialog closes and the original copy button enters the Copied state without another click
   - ✅ The clipboard contains an install-link token URL
   - 📸 `den-reauth-pending-action-03-reauth-copied-after-retry.png` — 4 validations passed

   

   - ✅ Script frame 1 narrated: "Alex is an org admin whose sign-in is more than fifteen minutes old. We sign in as Alex,"
   - ✅ Script frame 2 narrated: "From Members, Alex clicks Copy install link. Instead of a dead error banner, OpenWork op"
   - ✅ Script frame 3 narrated: "Alex enters his password once. The dialog closes, OpenWork retries the original copy act"

PR #2523, opened 2026-07-07
fix(den-web): install-link copy falls back to a manual share row when the browser blocks the clipboard

benjaminshafii:
## What

Follow-up to #2521/#2522, fixing the production report: *"The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission."* rendered as a members-page banner after using **Copy install link**.

That message is the browser's `NotAllowedError` from `navigator.clipboard.writeText`. The write ran **inside** the queued reauth action, so after a step-up verification the automatic retry executed it **without transient user activation** (the retry fires from a verify/message chain, not a click) — and browsers that deniy clipboard permission hit the same. The mint itself had succeeded (and `rotate: true` had already revoked older links), but the user got a raw browser error and no link.

## Fix

The clipboard write is now a best-effort presentation step **outside** the queued action (`manage-members-screen.tsx`):

- Success → identical to today: button flips to **Copied**.
- Blocked → the minted link renders inline: readonly select-all input + **Copy** button (a fresh user gesture, so the retry typically succeeds) + **Done** dismiss, under the helper line *"Copy blocked by the browser — copy the link manually:"*. Never an error banner.
- Mint failures (HTTP errors) still surface via the page error as before; reauth still queues via the dialog.

## Proof — `den-reauth-pending-action` extended to 4 steps (PASSED, frames on the comment below)

Steps 1–3 unchanged (stale session → dialog → password verify → auto-retry → Copied with the token URL on the clipboard). New step 4: deniy `clipboard-read`/`clipboard-write` via CDP `Browser.setPermission`, mint again, assert the manual share row appears with a `/install?token=…` value while the raw browser message ("not allowed by the user agent") is absent from the page; permissions re-granted for idempotency.

```
▶ den-reauth-pending-action — Den reauth keeps the pending install-link action alive
  ✓ Signing in and staging a stale session
  ✓ A privileged click opens the security check instead of a dead banner
  ✓ Verify password and the queued action completes on its own
  ✓ Even when the browser blocks copying, the link is still usable
  ✓ Voice-over script coverage
```

Also run: `pnpm --dir ee/apps/den-web exec tsc --noEmit` (pass), `pnpm --dir ee/apps/den-web build` (pass). Repro identical to #2521.

Known follow-up (out of scope): a dozen other `writeText` call sites across den-web share the raw-error exposure on permission-denied browsers; they all run in direct click handlers so they don't hit the activation gap fixed here.

### DIFF
--- a/ee/apps/den-web/app/(den)/dashboard/_components/manage-members-screen.tsx
+++ b/ee/apps/den-web/app/(den)/dashboard/_components/manage-members-screen.tsx
@@ -163,6 +163,8 @@ export function ManageMembersScreen() {
   const [seatBillingDialogError, setSeatBillingDialogError] = useState<OrgPaymentRequiredError | null>(null);
   const [installLinkBusy, setInstallLinkBusy] = useState(false);
   const [installLinkCopied, setInstallLinkCopied] = useState(false);
+  const [installLinkShareUrl, setInstallLinkShareUrl] = useState<string | null>(null);
+  const [installLinkShareCopied, setInstallLinkShareCopied] = useState(false);
 
   const assignableRoles = useMemo(
     () => (orgContext?.roles ?? []).filter((role) => !role.protected),
@@ -252,13 +254,38 @@ export function ManageMembersScreen() {
       : null;
   }
 
+  function selectInstallLinkShareInput() {
+    const input = document.getElementById("install-link-share-url");
+    if (input instanceof HTMLInputElement) {
+      input.focus();
+      input.select();
+    }
+  }
+
+  async function handleInstallLinkShareCopy() {
+    if (!installLinkShareUrl) {
+      return;
+    }
+
+    try {
+      await navigator.clipboard.writeText(installLinkShareUrl);
+      setInstallLinkShareCopied(true);
+      window.setTimeout(() => setInstallLinkShareCopied(false), 1800);
+    } catch {
+      selectInstallLinkShareInput();
+    }
+  }
+
   async function handleCopyInstallLink() {
     if (!activeOrg) {
       return;
     }
 
+    let mintedInstallPageUrl: string | null = null;
     setPageError(null);
     setInstallLinkCopied(false);
+    setInstallLinkShareUrl(null);
+    setInstallLinkShareCopied(false);
     setInstallLinkBusy(true);
     try {
       await runReauthableAction("copy-install-link", async () => {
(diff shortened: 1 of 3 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-07T04-25-14-017Z`

Full frame proof with validated screenshots: `evals/results/2026-07-07T04-25-14-017Z/fraimz.html` (re-run: `pnpm fraimz --flow den-reauth-pending-action`)

### ✅ den-reauth-pending-action — Den reauth keeps the pending install-link action alive
_User-facing flow demo_

1. **Alex is signed in with a stale privileged session**
   > 🎙 Alex is an org admin whose sign-in is more than fifteen minutes old. We sign in as Alex, deliberately age the session in the eval database, and keep the dashboard visible so the account still feels signed in.
   - ✅ The signed-in dashboard is visible
   - 📸 `den-reauth-pending-action-01-reauth-stale-session-dashboard.png` — 5 validations passed

   

2. **A stale privileged click opens the security check instead of a dead banner**
   > 🎙 From Members, Alex clicks Copy install link. Instead of a dead error banner, OpenWork opens the security check and clearly asks Alex to confirm it is him before changing workspace settings.
   - ✅ The reauth dialog appears for the stale privileged action
   - ✅ The raw server reauth message is not rendered as a page banner outside the dialog
   - 📸 `den-reauth-pending-action-02-reauth-security-check-dialog.png` — 4 validations passed

   

3. **After password verification, the queued copy action completes automatically**
   > 🎙 Alex enters his password once. The dialog closes, OpenWork retries the original copy action automatically, and the button flips to Copied with the install URL already on the clipboard.
   - ✅ The dialog closes and the original copy button enters the Copied state without another click
   - ✅ The clipboard contains an install-link token URL
   - 📸 `den-reauth-pending-action-03-reauth-copied-after-retry.png` — 4 validations passed

   

4. **Even when the browser blocks copying, the link is still usable**
   > 🎙 Even when the browser refuses programmatic copying, Alex is not stuck. The freshly minted install link appears right on the page with a manual copy affordance, and no scary browser error is shown.
   - ✅ The inline manual copy row appears with an install-link token URL
   - ✅ The raw browser clipboard error is not rendered anywhere on the page
   - ✅ The page error banner stays empty after clipboard denial
   - 📸 `den-reauth-pending-action-04-reauth-copy-blocked-manual-install-link.png` — 4 validations passed

   

   - ✅ Script frame 1 narrated: "Alex is an org admin whose sign-in is more than fifteen minutes old. We sign in as Alex,"
   - ✅ Script frame 2 narrated: "From Members, Alex clicks Copy install link. Instead of a dead error banner, OpenWork op"
   - ✅ Script frame 3 narrated: "Alex enters his password once. The dialog closes, OpenWork retries the original copy act"
   - ✅ Script frame 4 narrated: "Even when the browser refuses programmatic copying, Alex is not stuck. The freshly minte"

PR #2528, opened 2026-07-07
feat(den-web): render LaTeX math in user message bubbles

### LINKED ISSUE #2501: [Feature Request] Native LaTeX/MathJax rendering for chat messages
### Description

Currently, OpenWork's chat interface renders LaTeX math expressions (wrapped in `$$...$$` or `$...$`) as plain monospace text instead of typeset mathematical notation. This makes it difficult for users to discuss mathematical content — formulas, equations, integrals, matrices — within sessions.

### Expected Behavior

When a user writes LaTeX math expressions in a chat message, they should be rendered as formatted mathematical notation, similar to how GitHub renders math in Markdown (via MathJax or KaTeX).

For example:
- `$$\int_0^\infty e^{-x^2} \, dx = \frac{\sqrt{\pi}}{2}$$` should render as a display equation
- `$E = mc^2$` should render as inline math

### Actual Behavior

LaTeX expressions appear as plain text like `$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$` without any rendering.

### Workaround (DevTools Snippet)

I created a JavaScript snippet that can be run from DevTools (Sources > Snippets) to inject MathJax 3 into the page. It works reasonably well for most formulas but has limitations (detailed below).

**How to use the workaround:**
1. Open DevTools: `Ctrl+Shift+I` / `Cmd+Option+I`
2. Sources tab → Snippets → New snippet
3. Paste the code from `openwork-mathjax-inject.js` → Save → Run

**Console output when running:**
```
[MathJax] Loading from CDN...
[MathJax] Ready!
[MathJax] Injected successfully!
[MathJax] Watching for new messages...
```

**Features of the workaround:**
- Renders `$$...$$` as display math, left-aligned
- Renders `$...$` as inline math
- MutationObserver with debounce (500ms) to auto-render new messages
- Exposes `window.renderMath()` for manual re-rendering in Console
- CSS overrides to prevent line breaks and ensure proper alignment
- Fallback CDN (`cdnjs.cloudflare.com`) if primary CDN is unreachable

**Workaround code:** (see attached file `openwork-mathjax-inject.js`)

### Known Limitations of the Workaround

Despite solving basic math rendering, the following LaTeX environments **fail to render multi-line content** — everything collapses into a single line:

- `\begin{cases} ... \\ ... \end{cases}` — system of equations
- `\begin{vmatrix} ... \\ ... \end{vmatrix}` — matrix determinant
- `\begin{bmatrix} ... \\ ... \end{bmatrix}` — bracketed matrix
- `\begin{pmatrix} ... \\ ... \end{pmatrix}` — parenthesized matrix

The `\\` line-break operator within these environments is not being processed correctly. The `\begin{aligned}` environment with `&` alignment markers works, but true matrix/cases environments do not render properly.

I suspect this is because MathJax is injected into a React-rendered DOM where text nodes are split across multiple elements, or the `\\` syntax is being mangled during the `$$ → \[` preprocessing step.

### Suggested Solution

It would be great if OpenWork could add **native MathJax or KaTeX support** to the chat message renderer. This would:

1. Eliminate the need for DevTools workarounds
2. Render all LaTeX environments correctly (matrices, cases, aligned, etc.)
3. Provide a seamless experience for users discussing mathematics in their sessions

A good reference implementation would be [Obsidian's MathJax support](https://help.obsidian.md/Editing+and+formatting/Advanced+formatting+syntax#Math) or GitHub's native math rendering.

### Attachments

- `openwork-mathjax-inject.js` — The workaround DevTools Snippet for MathJax injection

PriyeshPandey2000:
## Summary
- Replace the full markdown-parser approach with a dedicated math-only renderer for user message bubbles.
- Renderer handles $$...$$ display math and $...$ inline math. Inline $ requires the next character to be non-digit and non-whitespace; closer $ must not be followed by a digit — prevents "$5 and $10" pairing as a math span.
- Plain text is HTML-escaped verbatim with whitespace-pre-wrap, so formatting characters (#, *, _, >) are shown literally — no markdown surprises.
- Output runs through the same sanitizeMarkdownHtml DOMPurify config as assistant messages.
- Falls back to plain-text + skill-chip renderer when skill tokens are present.

## Why
KaTeX was already wired for assistant messages but user messages bypassed the markdown renderer entirely, causing LaTeX to appear as raw `$$...$$` text in user bubbles.

## Issue
Closes #2501

## Scope
- apps/app/src/components/markdown/markdown.tsx — renderUserMath function + mathOnly prop on MarkdownBlock
- apps/app/src/components/ui/message.tsx — propagate mathOnly through MessageContent
- apps/app/src/components/chat/message-list.tsx — use mathOnly for user bubbles

## Out of scope
Segment-based rendering to support LaTeX inside skill-chip messages.

### Ran
- `pnpm typecheck` — pass

### Manual
1. Send `What is $E = mc^2$?` — inline math renders in user bubble ✓
2. Send `$$\int_0^\infty e^{-x^2}dx = \frac{\sqrt{\pi}}{2}$$` — display math renders ✓
3. Send `$$\begin{cases} x+y=1 \\ x-y=0 \end{cases}$$` — multi-line environment renders ✓ (this was the known failure case in the DevTools workaround from the issue)
4. Send plain multi-line text — line breaks preserved ✓
5. Ask assistant to show LaTeX — assistant rendering unaffected ✓

## Evidence
Screenshot 2026-07-07 at 12 29 33 PM
Screenshot 2026-07-07 at 12 30 56 PM

Screenshot 2026-07-07 at 12 31 05 PM
Screenshot 2026-07-08 at 10 45 04 PM

Screenshot 2026-07-08 at 10 46 56 PM
Screenshot 2026-07-08 at 10 47 02 PM

### DIFF
--- a/apps/app/src/components/chat/message-list.tsx
+++ b/apps/app/src/components/chat/message-list.tsx
@@ -669,11 +669,11 @@ const UserMessage = React.memo(
   ({ message, isStreaming }: UserMessageProps) => {
     const { onRevertToUserMessage, onForkAtMessage, onEditUserMessage, highlightQuery } = useMessageList()
     const messageText = React.useMemo(() => getMessagesText([message]), [message])
-    const inlineParts = React.useMemo(
-      () => message.parts.filter((part) => (part.type === "text" && Boolean(part.text)) || isFileUIPart(part)),
+    const userText = React.useMemo(
+      () => message.parts.map((part) => (part.type === "text" ? part.text : "")).join(""),
       [message.parts],
     )
-    const hasContent = inlineParts.length > 0
+    const hasSkillChips = USER_SKILL_TOKEN_RE.test(userText)
 
     return (
       <Message
@@ -690,32 +690,30 @@ const UserMessage = React.memo(
                 className="group flex w-full flex-col items-end gap-1 !select-text"
                 style={{ userSelect: "text" }}
               >
-                {hasContent ? (
-                  <MessageContent
-                    className="bg-muted text-foreground max-w-[85%] rounded-3xl px-4 py-2.5 leading-6 sm:max-w-[75%] !select-text not-prose"
-                    style={{ userSelect: "text" }}
-                  >
-                    {inlineParts.map((part, index) => {
-                      if (part.type === "text") {
-                        return (
-                          <span key={`text-${index}`} className="whitespace-pre-wrap">
-                            {renderUserTextWithSkillChips(part.text, highlightQuery)}
-                          </span>
-                        )
-                      }
-                      if (isFileUIPart(part)) {
-                        return (
-                          <span
-                            key={`file-${part.url}-${index}`}
-                            className="mx-1 inline-flex align-middle not-prose"
-                          >
-                            <FileMessage part={part} tone="user" />
-                          </span>
-                        )
-                      }
-                      return null
-                    })}
-                  </MessageContent>
+                {message.parts.filter(isFileUIPart).map((part, index) => (
+                  <FileMessage key={`${part.url}-${index}`} part={part} tone="user" />
+                ))}
+                {message.parts.some((part) => part.type === "text" && part.text) ? (
+                  // NOTE: when skill chips are present we fall back to plain-text rendering.
+                  // LaTeX won't render in that path. Known tradeoff — chips + math in the same
+                  // message is rare today, but worth revisiting if math skills are introduced.
+                  hasSkillChips ? (
+                    <MessageContent
+                      className="bg-muted text-foreground max-w-[85%] rounded-3xl px-4 py-2.5 leading-6 whitespace-pre-wrap sm:max-w-[75%] !select-text not-prose"
+                      style={{ userSelect: "text" }}
+                    >
+                      {renderUserTextWithSkillChips(userText, highlightQuery)}
+                    </MessageContent>
+                  ) : (
+                    <MessageContent
+                      className="bg-muted text-foreground max-w-[85%] rounded-3xl px-4 py-2.5 leading-6 whitespace-pre-wrap sm:max-w-[75%] !select-text not-prose"
+                      style={{ userSelect: "text" }}
+                      mathOnly
+                      highlightQuery={highlightQuery}
+                    >
+                      {userText}
+                    </MessageContent>
+                  )
                 ) : null}
                 {!isStreaming && (
                   <MessageActions
--- a/apps/app/src/components/markdown/markdown-primitive.ts
+++ b/apps/app/src/components/markdown/markdown-primitive.ts
@@ -1,5 +1,7 @@
 import DOMPurify from "dompurify";
 import emojiKeywords from "emojilib";
+import katex from "katex";
+import "katex/dist/katex.min.css";
 import { Marked, type MarkedExtension, type Token, type Tokens } from "marked";
 import { markedEmoji } from "marked-emoji";
 import {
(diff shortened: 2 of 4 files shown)


Pablosinyores (commented):
Good to see math rendering in the bubbles. Two things worth weighing before merge, both stemming from user messages moving from inert `whitespace-pre-wrap` text to full markdown:

1. Single-`$` inline math vs currency. `markedKatex({ throwOnError: false })` uses the default delimiters, so `$...$` is live inline math. In user-typed chat that false-triggers on ordinary dollar amounts — e.g. "it costs $5 and $10 more" parses `$5 and $` as a math span, typesetting "5 and " and eating the literal dollar signs. Prices in chat are common enough that this will bite regularly. Options: drop the single-`$` inline delimiter and keep only `$$...$$` for display math, or configure the extension so a `$` immediately followed by a digit is not treated as an opener. Worth trying a couple of price messages.

2. Trust boundary on user text. Routing user messages through the markdown renderer means their text is now interpreted as markdown/HTML rather than shown verbatim. Two sub-points: (a) formatting surprises — a user who types `*`, `_`, `#`, or `> ` now gets it reformatted instead of shown literally, a behavior change from the old bubble; (b) security — `[x](...)` links become active, so please confirm this path runs the same DOMPurify config as assistant markdown and that it rejects `javascript:`/`data:` URL schemes, since this is the first place raw user input gets HTML-rendered. The `aria-hidden` ADD_ATTR and `throwOnError: false` are both the right calls.

The documented skill-chip fallback tradeoff reads reasonable — agree chips plus math in one message is rare today.

PriyeshPandey2000:
> Good to see math rendering in the bubbles. Two things worth weighing before merge, both stemming from user messages moving from inert `whitespace-pre-wrap` text to full markdown:
> 
> 1. Single-`$` inline math vs currency. `markedKatex({ throwOnError: false })` uses the default delimiters, so `$...$` is live inline math. In user-typed chat that false-triggers on ordinary dollar amounts — e.g. "it costs $5 and $10 more" parses `$5 and $` as a math span, typesetting "5 and " and eating the literal dollar signs. Prices in chat are common enough that this will bite regularly. Options: drop the single-`$` inline delimiter and keep only `$$...$$` for display math, or configure the extension so a `$` immediately followed by a digit is not treated as an opener. Worth trying a couple of price messages.
> 2. Trust boundary on user text. Routing user messages through the markdown renderer means their text is now interpreted as markdown/HTML rather than shown verbatim. Two sub-points: (a) formatting surprises — a user who types `*`, `_`, `#`, or `> ` now gets it reformatted instead of shown literally, a behavior change from the old bubble; (b) security — `[x](...)` links become active, so please confirm this path runs the same DOMPurify config as assistant markdown and that it rejects `javascript:`/`data:` URL schemes, since this is the first place raw user input gets HTML-rendered. The `aria-hidden` ADD_ATTR and `throwOnError: false` are both the right calls.
> 
> The documented skill-chip fallback tradeoff reads reasonable — agree chips plus math in one message is rare today.

Thanks for the detailed review , reworked the approach to address both points.

1. **Approach**: replaced full markdown parser with a math-only renderer (`renderUserMath`) that handles only `$$...$$` and `$...$` — everything else is HTML-escaped verbatim.

2. **Currency false-positives**: inline opener requires the char after `$` to be non-digit and non-whitespace; closer `$` must not be followed by a digit."it costs $5 and $10 more" stays plain text; `$E=mc^2$` renders correctly.

3. **Formatting surprises**: eliminated entirely — `#`, `*`, `_`, `>` are now escaped as plain text, same behaviour as the old bubble.

**Security**: renderer never produces links or user-controlled HTML. Plain text goes through `escapeHtml()`, KaTeX output is from a trusted library, and the full output runs through `sanitizeMarkdownHtml()` (same DOMPurify config as assistant messages) before hitting the DOM.

PriyeshPandey2000:
Hi @reachjalil - rebased onto current `dev` and resolved conflicts. Main change during rebase: moved `renderUserMathHtml` + katex import into `markdown-primitive.ts` (where the rest of the markdown pipeline now lives after the refactor) instead of adding them back to `markdown.tsx`. `pnpm typecheck` and `pnpm test` (536 pass, 0 fail) both pass.

PR #2552, opened 2026-07-07
fix(den): pin org-scoped Connections requests — explicit org scope wins over drifted sessions

benjaminshafii:
## Problem

Multi-org accounts hit `organization_not_found` (404) on the cloud Connections surface — e.g. creating the Slack MCP connection — whenever the session's active org was stale or drifted mid-flow (second tab, desktop app, reseeded instance). Worse, a *valid but different* session org silently routed writes into the wrong org.

Root cause chain:
- `sessionMiddleware` eagerly sets `activeOrganizationId` from the session, and `resolveOrganizationContextMiddleware` short-circuited on it — ignoring the explicit org header (which the desktop already sends, and the internal legacy `/v1/orgs/:orgId/*` proxy sets) and skipping the multi-org re-resolution fallback entirely.
- den-web never pinned org-scoped calls at all; the org shown in the UI lived only in client state while requests rode the session-global active org.

## Fix

**den-api**
- Canonical `x-openwork-org-id` header (the legacy `x-openwork-legacy-org-id` stays as an alias; API-key scope still wins over both).
- Explicit scope now takes precedence over the session-derived active org; membership is verified and a pin to a non-member org stays a hard 404.
- A stale session org self-heals by re-resolving (single-org auto-pick + session rehydration) instead of hard-404ing before the fallback ever ran.
- Explicit scopes no longer mutate the session's active org.

**den-web**
- Connections-surface requests (list / create / access / delete / connect-start / native provider client GET+POST) carry `x-openwork-org-id` for the org on screen; the proxy whitelists the header.

## Evidence

- `bun test` (den-api middleware/org/api-key/internal-principal/route-policy suites): **31 pass / 0 fail**, incl. new `test/organization-context.test.ts` covering precedence, alias, non-member 404, stale-session heal, multi-org 404, drift+header, API-key-over-header, and the happy path staying resolution-free.
- `tsc --noEmit` clean for den-api and den-web.
- fraimz `org-scope-pinning` (new, gated by `OPENWORK_EVAL_DEN_MULTI_ORG`; skips on single-org stacks): **PASSED** — session drifts to another org mid-dialog, create still succeeds, the connection lands in the on-screen org and *not* the drifted org (API-witnessed both ways). **Falsification: the same flow fails with `organization_not_found` on unpatched code.**
- fraimz `slack-org-connection` (canonical surface flow): **PASSED** on the patched single-org stack.
- Proof comment with validated screenshots follows below.

Known pre-existing (untouched): `pnpm -C ee/apps/den-web build` fails on clean dev (`useSearchParams()` without Suspense in `org-dashboard-provider.tsx`, introduced by #2503).

### DIFF
--- a/ee/apps/den-api/src/middleware/organization-context.ts
+++ b/ee/apps/den-api/src/middleware/organization-context.ts
@@ -3,7 +3,7 @@ import type { MiddlewareHandler } from "hono"
 import { getApiKeyScopedOrganizationId, isScopedApiKeyForOrganization } from "../api-keys.js"
 import { getOrganizationContextForUser, resolveUserOrganizations, type OrganizationContext } from "../orgs.js"
 import type { AuthContextVariables } from "../session.js"
-import { getLegacyProxyOrganizationId, hydrateSessionActiveOrganization, shouldHydrateSessionActiveOrganization, type UserOrganizationsContext } from "./user-organizations.js"
+import { getRequestScopedOrganizationId, hydrateSessionActiveOrganization, shouldHydrateSessionActiveOrganization, type UserOrganizationsContext } from "./user-organizations.js"
 
 export type OrganizationContextVariables = {
   organizationContext: OrganizationContext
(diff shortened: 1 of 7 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-07T22-19-37-994Z`

Full frame proof with validated screenshots: `evals/results/2026-07-07T22-19-37-994Z/fraimz.html` (re-run: `pnpm fraimz --flow org-scope-pinning`)

### ✅ org-scope-pinning — Connections requests stay pinned to the org on screen when the session's active org drifts
_User-facing flow demo_

1. **The admin is unmistakably operating in the chosen org's Connections screen**
   > 🎙 An admin who belongs to several organizations signs in, picks one in the organization chooser, and lands on that org's Connections screen — the dashboard is now clearly operating in the org they chose.
   - ✅ Visible text includes "Add Custom"
2. **The Connections screen is open in the org the admin chose.**
   > 🎙 An admin who belongs to several organizations signs in, picks one in the organization chooser, and lands on that org's Connections screen — the dashboard is now clearly operating in the org they chose.
   - 📸 `org-scope-pinning-01-connections-in-chosen-org.png` — 8 validations passed

   

3. **Creating a connection survives an active-org drift because requests are pinned to the on-screen org**
   > 🎙 While the add-connection dialog is open, the account's active organization flips to a different org behind the scenes — the same drift a second tab or the desktop app can cause. The admin finishes the dialog anyway, and the connection is created without any organization-not-found error, because every Connections request now carries the org the admin is actually looking at.
   - ✅ Visible text includes "/connect/callback"
   - ✅ Visible text does not include "organization_not_found"
4. **The connection is created and the redirect-URL handoff renders even though the session's active org drifted mid-dialog.**
   > 🎙 While the add-connection dialog is open, the account's active organization flips to a different org behind the scenes — the same drift a second tab or the desktop app can cause. The admin finishes the dialog anyway, and the connection is created without any organization-not-found error, because every Connections request now carries the org the admin is actually looking at.
   - 📸 `org-scope-pinning-02-create-survives-org-drift.png` — 7 validations passed

   

5. **The org on screen owns the connection; the drifted org never saw the write**
   > 🎙 Back on the same org's Connections list, the new connection is right there — and the API confirms it lives in the org the admin saw on screen, not in the org the session had drifted to. The stray write and the confusing 404 are both gone.
6. **After the drift is undone, the org's Connections list shows the probe connection exactly where the admin created it.**
   > 🎙 Back on the same org's Connections list, the new connection is right there — and the API confirms it lives in the org the admin saw on screen, not in the org the session had drifted to. The stray write and the confusing 404 are both gone.
   - 📸 `org-scope-pinning-03-connection-listed-in-correct-org.png` — 6 validations passed

   

   - ✅ Script frame 1 narrated: "An admin who belongs to several organizations signs in, picks one in the organization ch"
   - ✅ Script frame 2 narrated: "While the add-connection dialog is open, the account's active organization flips to a di"
   - ✅ Script frame 3 narrated: "Back on the same org's Connections list, the new connection is right there — and the API"

PR #2578, opened 2026-07-08
fix(den): pin org scope on every dashboard surface and the desktop config fetch

benjaminshafii:
## Why

Multi-org accounts still hit intermittent `organization_not_found` after #2552. That fix pinned `x-openwork-org-id` only on the Connections hooks — every other den-web dashboard request (org settings, invitations, members, roles, teams, integrations, plugins, LLM providers, analytics, workers) and the desktop app's `/v1/me/desktop-config` fetch still rode the session's drift-prone active org. Any second tab / desktop app / reseeded instance could flip or stale the active org and 404 those surfaces (or silently read/write the wrong org). The desktop config fetch runs at boot, on every settings event, and hourly — its provider even has a dedicated `organization_not_found` recovery handler.

## What

- **den-web**: `requestJson` now auto-pins every `/v1/*` request with the on-screen org via a module scope owned by the org dashboard provider. `/v1/me` and `/v1/me/orgs` stay unpinned (a scope header filters the org directory server-side, which would break the switcher). `/v1/org` context loads are pinned to the org picked from the fresh directory, killing the boot race. `organization_not_found` on context load/switch now recovers to the org chooser instead of dead-ending in an error banner. A routine refresh keeps the on-screen org sticky (a drifted session no longer silently flips the tab's org), and a failed switch restores the previous scope.
- **desktop**: `getDesktopConfig(orgId)` pins the persisted active org via the existing legacy header — stale multi-org sessions stop 404ing the desktop-policy fetch.
- **server**: zero changes; the middleware precedence from #2552 already honors explicit scopes.
- **evals**: new `org-scope-dashboard-pinning` fraimz flow (gated by `OPENWORK_EVAL_DEN_MULTI_ORG`) proves a settings rename issued mid-drift lands on the org on screen and leaves the drifted org untouched. Eval lib now routes auth calls through den-web so Node's forbidden-header `Origin` stripping can't break den flows.

## Testing

- `cd ee/apps/den-api && bun test test/organization-context.test.ts` — 8/8 pass (middleware precedence unchanged).
- `pnpm --filter @openwork-ee/den-web exec tsc --noEmit` — pass.
- `pnpm --filter @openwork/app typecheck` — pass.
- `pnpm --filter @openwork/app exec bun test tests/den-desktop-config.test.ts` — pass (desktop config request carries the org header).
- `node org-scope.test.mts && node den-org.selection.test.mts` (den-web `_lib`) — pass.
- Fraimz e2e on a Daytona two-sandbox stack (Den server @ this branch with `DEN_ORG_MODE=multi_org` + seeded demo org; standalone Chrome via CDP): `pnpm fraimz --flow org-scope-dashboard-pinning` — **PASSED** (proof comment below).

Known limitation: the Daytona Electron sandbox's app instance was flaky (SIGTRAP crashes unrelated to this diff — the flow drives den-web through standalone Chrome instead). The desktop-side change is covered by the bun unit test + typecheck.

### DIFF
--- a/apps/app/src/app/lib/den.ts
+++ b/apps/app/src/app/lib/den.ts
@@ -2029,10 +2029,11 @@ export function createDenClient(options: { baseUrl: string; apiBaseUrl?: string 
       return appVersionMetadata;
     },
 
-    async getDesktopConfig(): Promise<DenDesktopConfig> {
+    async getDesktopConfig(orgId?: string | null): Promise<DenDesktopConfig> {
       const payload = await requestJson<unknown>(baseUrls, "/v1/me/desktop-config", {
         method: "GET",
         token,
+        organizationId: orgId,
       });
       return normalizeDenDesktopConfig(payload);
     },
--- a/apps/app/src/react-app/domains/cloud/desktop-config-provider.tsx
+++ b/apps/app/src/react-app/domains/cloud/desktop-config-provider.tsx
@@ -179,9 +179,10 @@ export function DesktopConfigProvider({ children }: DesktopConfigProviderProps) 
     const currentRun = ++refreshRunRef.current;
     const settings = readDenSettings();
     const token = settings.authToken?.trim() ?? "";
+    const activeOrgId = settings.activeOrgId?.trim() ?? "";
     const cacheKey = getDesktopConfigCacheKey();
 
-    if (!isSignedIn || !token || !settings.activeOrgId?.trim()) {
+    if (!isSignedIn || !token || !activeOrgId) {
       applyDesktopConfigActions(DEFAULT_DESKTOP_CONFIG);
       setDesktopConfigState((current) => ({ ...current, loading: false }));
       return;
@@ -201,7 +202,7 @@ export function DesktopConfigProvider({ children }: DesktopConfigProviderProps) 
         baseUrl: settings.baseUrl,
         apiBaseUrl: settings.apiBaseUrl,
         token,
-      }).getDesktopConfig();
+      }).getDesktopConfig(activeOrgId);
 
       if (currentRun !== refreshRunRef.current) return;
 
--- a/apps/app/tests/den-desktop-config.test.ts
+++ b/apps/app/tests/den-desktop-config.test.ts
@@ -1,0 +1,34 @@
+import { afterEach, describe, expect, test } from "bun:test";
+
+import { createDenClient } from "../src/app/lib/den";
+
+const originalFetch = globalThis.fetch;
+
+describe("Den desktop config client", () => {
+  afterEach(() => {
+    Object.defineProperty(globalThis, "fetch", {
+      configurable: true,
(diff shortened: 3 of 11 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-08T07-08-41-522Z`

Full frame proof with validated screenshots: `evals/results/2026-07-08T07-08-41-522Z/fraimz.html` (re-run: `pnpm fraimz --flow org-scope-dashboard-pinning`)

### ✅ org-scope-dashboard-pinning — Dashboard settings writes stay pinned to the org on screen when the session's active org drifts
_User-facing flow demo_

1. **The admin is on org A's Settings screen**
   > 🎙 The admin signs in, chooses the primary workspace, and lands on that workspace's Settings screen. The name field and the page chrome agree on which organization is being edited before anything changes.
2. **The admin is on org A's Settings screen before the drift is armed.**
   > 🎙 The admin signs in, chooses the primary workspace, and lands on that workspace's Settings screen. The name field and the page chrome agree on which organization is being edited before anything changes.
   - 📸 `org-scope-dashboard-pinning-01-org-settings-on-org-a.png` — 7 validations passed

   

3. **Rename survives an active-org drift and lands on the org on screen**
   > 🎙 While the Settings screen stays open, the account's active workspace changes somewhere else. The admin saves a new workspace name anyway, sees a success message, and the renamed workspace is the one still on screen — not the workspace the session drifted to.
   - ✅ Visible text does not include "organization_not_found"
4. **The rename succeeds from the open Settings screen even after the browser session drifts to another org.**
   > 🎙 While the Settings screen stays open, the account's active workspace changes somewhere else. The admin saves a new workspace name anyway, sees a success message, and the renamed workspace is the one still on screen — not the workspace the session drifted to.
   - 📸 `org-scope-dashboard-pinning-02-rename-survives-org-drift.png` — 6 validations passed

   

5. **The dashboard stays on the admin's org and shows the new name**
   > 🎙 After the save settles, the dashboard remains anchored to the same workspace and shows the new name. The other workspace is unchanged, so the admin can trust that Settings edits stay scoped to what they are looking at.
6. **After the settings refresh, the dashboard remains on the org the admin edited and shows the new name.**
   > 🎙 After the save settles, the dashboard remains anchored to the same workspace and shows the new name. The other workspace is unchanged, so the admin can trust that Settings edits stay scoped to what they are looking at.
   - 📸 `org-scope-dashboard-pinning-03-dashboard-stays-on-screen-org.png` — 7 validations passed

   

   - ✅ Script frame 1 narrated: "The admin signs in, chooses the primary workspace, and lands on that workspace's Setting"
   - ✅ Script frame 2 narrated: "While the Settings screen stays open, the account's active workspace changes somewhere e"
   - ✅ Script frame 3 narrated: "After the save settles, the dashboard remains anchored to the same workspace and shows t"

PR #2597, opened 2026-07-08
feat(extensions): connect-aware gating of legacy Google Workspace actions behind connectEnabled

benjaminshafii:
## Why

Enterprise sessions hit a collision between the two Google Workspace integrations: the legacy per-device extensions surface always advertises 14 GW actions, so on cloud-connected devices the agent runs `status`, reads "missing OAuth client secret", and tells users to reconnect in **Settings > Extensions** — even though they are connected via Connect. Seen live on an enterprise demo call.

## What

Adds an org-level `connectEnabled` switch to the openwork-server, with a hard safety rule: **gating only ever hides actions that would fail with `google_workspace_not_connected` anyway.**

| Legacy configured? | connectEnabled | Behavior |
|---|---|---|
| yes | any | untouched (all 14 GW actions) |
| no | off | untouched (today's behavior) |
| no | on | 13 dead GW actions hidden; `status` stays; calls to hidden actions return a structured `use_openwork_cloud` redirect; status gains an additive `connect` guidance block. Image generation and every other extension untouched |

- `apps/server/src/connect-state.ts` — persisted `connect-state.json` next to `runtime.sqlite`, request-time snapshot (flag + openwork-cloud MCP presence + legacy GW credentials).
- `GET/PUT /experimental/connect/state` (client/host scope) in `routes/core.ts`.
- Gating + redirect in `extensions/index.ts`; additive `connect` block in GW status.
- Desktop pushes the org desktop-config `connectEnabled` to the server on change (fire-and-forget, gated on config load; harmless 404 on older servers).

Flag default is **off** → zero behavior change until an org opts in. Effective on the next request — no engine restart.

## Tests

Run on a Daytona sandbox (`openwork-test-20260708-151032`, branch stack incl. this PR):
- `bun test` (apps/server, full suite): **581 pass / 4 skip / 1 fail** — the 1 failure (`ensureWorkspaceFiles > uses shipped extension preview plugin`) **also fails on the base commit 848f65a4c** (pre-existing test-order pollution; passes in isolation on this branch).
- New matrix tests: `apps/server/src/extensions-connect-gating.test.ts` (5 tests — default/off/legacy-configured/gated+redirect+status/PUT validation+roundtrip) — all pass.
- `pnpm --filter openwork-server typecheck`, `pnpm --filter @openwork/app typecheck` — pass.

## Evidence

Fraimz (7 frames, PASSED) driven against the real Electron app + embedded server on Daytona — posted as a comment below (flow `connect-aware-extension-gating`, committed on the stacked steering PR).

## Notes for reviewers

- Stacked follow-up PR: connect-aware system-prompt steering (consumes `GET /experimental/connect/state`).
- Repro: `PUT /experimental/connect/state {"connectEnabled":true}` (host token) then `GET /experimental/extensions/actions`.

### DIFF
--- a/apps/app/src/app/lib/openwork-server.ts
+++ b/apps/app/src/app/lib/openwork-server.ts
@@ -390,6 +390,19 @@ export type GoogleWorkspaceAuthStatus = {
     driveFileName: string | null;
     gmailDraftId: string | null;
   } | null;
+  connect?: {
+    enabled: true;
+    cloudMcpPresent: boolean;
+    guidance: string;
+  };
+};
+
+export type OpenworkConnectState = {
+  ok: true;
+  schemaVersion: 1;
+  connectEnabled: boolean;
+  cloudMcpPresent: boolean;
+  googleWorkspace: { legacyConfigured: boolean };
 };
 
 export type GoogleWorkspaceConnectStart = {
@@ -413,13 +426,19 @@ export type OpenworkExtensionActionCall = {
   context?: Record<string, unknown>;
 };
 
-export type OpenworkExtensionActionResult = {
-  ok: boolean;
-  extensionId: string;
-  action: string;
-  result: unknown;
-  context?: Record<string, unknown>;
-};
+export type OpenworkExtensionActionResult =
+  | {
+    ok: true;
+    extensionId: string;
+    action: string;
+    result: unknown;
+    context?: Record<string, unknown>;
+  }
+  | {
+    ok: false;
+    error: string;
+    message: string;
+  };
 
 export type OpenworkResolvedArtifactTarget = {
   id: string;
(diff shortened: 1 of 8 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-08T22-57-03-705Z`

Full frame proof with validated screenshots: `evals/results/2026-07-08T22-57-03-705Z/fraimz.html` (re-run: `pnpm fraimz --flow connect-aware-extension-gating`)

### ✅ connect-aware-extension-gating — Connect flag gates dead legacy Google Workspace actions and redirects the agent to OpenWork Cloud
_Internal demo_

1. **With connectEnabled off, every legacy extension action is offered exactly as before**
   > 🎙 This is a fresh device: the org has not enabled Connect, and no legacy Google OAuth client is configured. The legacy extension surface offers every action it always has — Google Workspace and image generation side by side — byte-for-byte the behavior shipped today.
2. **PUT /experimental/connect/state {connectEnabled:true} persists and reports the device snapshot**
   > 🎙 Now the organization flips one switch: connect enabled. This is the exact endpoint the desktop app pushes when the org's cloud config arrives — one boolean, stored server-side, effective on the very next request.
3. **Gating hides exactly the 13 dead Google Workspace actions and nothing else**
   > 🎙 The tool surface transforms. The thirteen Google Workspace actions that could only fail on this device are gone; the safe status probe stays; and image generation — which has no cloud equivalent — is untouched. The agent can no longer stumble into a dead end.
4. **A gated call answers use_openwork_cloud with Settings > Connect guidance**
   > 🎙 If an agent still calls a hidden Google Workspace action, it no longer gets an OAuth error to misread. It gets a redirect: use OpenWork Cloud's search and execute capabilities, and if the user needs to connect, send them to Settings > Connect — never Settings > Extensions.
5. **GET google-workspace/status includes an additive connect block with guidance**
   > 🎙 The status probe itself now carries the same guidance, so even an agent that only checks status walks away with the correct next step instead of a missing-client-secret diagnosis.
6. **Disabling connectEnabled restores all legacy actions with no restart**
   > 🎙 And it is a real kill switch: flip the flag off and the full legacy surface is back on the next request — no restart, no reinstall, nothing lost for teams that are not ready.
7. **Settings > Connect renders as the org connections surface**
   > 🎙 Behind it all is the surface users are actually pointed to: Settings > Connect, where Google Workspace and the rest of the org's connections live in the cloud, connected once per member.
   - ✅ URL hash includes "/settings/connect"
   - ✅ Visible text includes "Connect for teams"
   - 📸 `connect-aware-extension-gating-01-settings-connect-surface.png` — 5 validations passed

   

   - ✅ Script frame 1 narrated: "This is a fresh device: the org has not enabled Connect, and no legacy Google OAuth clie"
   - ✅ Script frame 2 narrated: "Now the organization flips one switch: connect enabled. This is the exact endpoint the d"
   - ✅ Script frame 3 narrated: "The tool surface transforms. The thirteen Google Workspace actions that could only fail "
   - ✅ Script frame 4 narrated: "If an agent still calls a hidden Google Workspace action, it no longer gets an OAuth err"
   - ✅ Script frame 5 narrated: "The status probe itself now carries the same guidance, so even an agent that only checks"
   - ✅ Script frame 6 narrated: "And it is a real kill switch: flip the flag off and the full legacy surface is back on t"
   - ✅ Script frame 7 narrated: "Behind it all is the surface users are actually pointed to: Settings > Connect, where Go"

PR #2598, opened 2026-07-08
feat(extensions): connect-aware system-prompt steering toward openwork-cloud capabilities

benjaminshafii:
## Why

Stacked on #2597 (auto-retargets to `dev` when it merges). Gating removes the dead tools; this PR fixes the **recruitment**: the extensions-preview plugin unconditionally injects a system-prompt instruction steering agents to `openwork_extension_list_actions` for anything unusual, with zero equivalent steering toward `openwork-cloud_search_capabilities`. That is why "check my email" walks into the legacy path in the first place.

## What

`apps/server/src/opencode-plugins/openwork-extensions-preview.ts`:
- At system-transform time the plugin fetches `GET /experimental/connect/state` (its existing server phone-home pattern), with a 15s TTL cache and **fail-open** to today's static instruction on any error — steering can never break a session.
- Pure, exported composition `composeOpenWorkExtensionDiscoveryInstruction(state)`:
  - state unavailable / flag off / legacy configured → **byte-identical** static instruction (zero change);
  - Connect active + cloud MCP present → steer Gmail/Calendar/Drive/org services to `openwork-cloud_search_capabilities` → `execute_capability` first; extensions stay for local actions (image generation); never direct users to Settings > Extensions for Google Workspace — use Settings > Connect;
  - Connect active + cloud MCP absent → static instruction + "connect in Settings > Connect (never Settings > Extensions)".

Also carries the fraimz flow + approved voiceover for the whole gating experience (`evals/flows/connect-aware-extension-gating.flow.mjs`).

## Tests

Run on Daytona (same sandbox as #2597):
- `apps/server/src/opencode-plugins/openwork-extensions-preview-connect-steering.test.ts` — 8 tests: byte-identical fallback ×3, both steering variants, bearer-auth fetch + 15s TTL, fail-open on fetch error and on parse error. All pass.
- Full apps/server suite: 581 pass / 1 pre-existing fail (identical on base commit — see #2597).
- `pnpm --filter openwork-server typecheck` — pass.

## Evidence

Fraimz (7 frames, PASSED) against the real app on Daytona — comment below.

## Honest gap

Model-driven e2e (a live agent choosing the cloud tool from a natural prompt) is not covered here; the composition layer is unit-tested and the consumed state endpoint is fraimz-proven. A model-in-the-loop eval matrix is the natural follow-up.

### DIFF
--- a/apps/server/src/opencode-plugins/openwork-extensions-preview-connect-steering.test.ts
+++ b/apps/server/src/opencode-plugins/openwork-extensions-preview-connect-steering.test.ts
@@ -1,0 +1,145 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+
+import {
+  composeOpenWorkExtensionDiscoveryInstruction,
+  OPENWORK_CLOUD_CONNECTION_INSTRUCTION,
+  OPENWORK_CONNECT_GOOGLE_WORKSPACE_DISCONNECTED_INSTRUCTION,
+  OPENWORK_EXTENSION_DISCOVERY_INSTRUCTION,
+  resetOpenWorkExtensionDiscoveryInstructionCacheForTests,
+  resolveOpenWorkExtensionDiscoveryInstruction,
+  type OpenWorkExtensionConnectState,
+} from "./openwork-extensions-preview.js";
+
+const originalServerUrl = process.env.OPENWORK_SERVER_URL;
+const originalServerToken = process.env.OPENWORK_SERVER_TOKEN;
+
+const UNCHANGED_EXTENSION_DISCOVERY_INSTRUCTION =
+  "If the user asks for something you cannot do with obvious built-in tools, check OpenWork extensions before saying the capability is unavailable. Use openwork_extension_list_actions to inspect available extension actions, then call the matching action with openwork_extension_call.";
+
+const CLOUD_CONNECTION_INSTRUCTION =
+  "The OpenWork Cloud connection is active. For email (Gmail), calendar, Google Drive, and org-connected services such as Notion, Linear, Slack, etc., FIRST call openwork-cloud_search_capabilities with 2-4 keyword variants, then call openwork-cloud_execute_capability with an exact returned name. Do not claim these are unavailable without searching. OpenWork extensions (openwork_extension_list_actions / openwork_extension_call) remain available for other local actions such as image generation, but do NOT use them for Google Workspace, and never direct the user to Settings > Extensions for Google Workspace; use Settings > Connect.";
+
+const GOOGLE_WORKSPACE_DISCONNECTED_INSTRUCTION =
+  `${UNCHANGED_EXTENSION_DISCOVERY_INSTRUCTION} Google Workspace is not connected on this device; if the user asks for email, calendar, or Google Drive, tell them to connect their account in Settings > Connect (never Settings > Extensions).`;
+
+beforeEach(() => {
+  resetOpenWorkExtensionDiscoveryInstructionCacheForTests();
+});
+
+afterEach(() => {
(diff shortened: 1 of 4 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-08T22-57-20-011Z`

Full frame proof with validated screenshots: `evals/results/2026-07-08T22-57-20-011Z/fraimz.html` (re-run: `pnpm fraimz --flow connect-aware-extension-gating`)

### ✅ connect-aware-extension-gating — Connect flag gates dead legacy Google Workspace actions and redirects the agent to OpenWork Cloud
_Internal demo_

1. **With connectEnabled off, every legacy extension action is offered exactly as before**
   > 🎙 This is a fresh device: the org has not enabled Connect, and no legacy Google OAuth client is configured. The legacy extension surface offers every action it always has — Google Workspace and image generation side by side — byte-for-byte the behavior shipped today.
2. **PUT /experimental/connect/state {connectEnabled:true} persists and reports the device snapshot**
   > 🎙 Now the organization flips one switch: connect enabled. This is the exact endpoint the desktop app pushes when the org's cloud config arrives — one boolean, stored server-side, effective on the very next request.
3. **Gating hides exactly the 13 dead Google Workspace actions and nothing else**
   > 🎙 The tool surface transforms. The thirteen Google Workspace actions that could only fail on this device are gone; the safe status probe stays; and image generation — which has no cloud equivalent — is untouched. The agent can no longer stumble into a dead end.
4. **A gated call answers use_openwork_cloud with Settings > Connect guidance**
   > 🎙 If an agent still calls a hidden Google Workspace action, it no longer gets an OAuth error to misread. It gets a redirect: use OpenWork Cloud's search and execute capabilities, and if the user needs to connect, send them to Settings > Connect — never Settings > Extensions.
5. **GET google-workspace/status includes an additive connect block with guidance**
   > 🎙 The status probe itself now carries the same guidance, so even an agent that only checks status walks away with the correct next step instead of a missing-client-secret diagnosis.
6. **Disabling connectEnabled restores all legacy actions with no restart**
   > 🎙 And it is a real kill switch: flip the flag off and the full legacy surface is back on the next request — no restart, no reinstall, nothing lost for teams that are not ready.
7. **Settings > Connect renders as the org connections surface**
   > 🎙 Behind it all is the surface users are actually pointed to: Settings > Connect, where Google Workspace and the rest of the org's connections live in the cloud, connected once per member.
   - ✅ URL hash includes "/settings/connect"
   - ✅ Visible text includes "Connect for teams"
   - 📸 `connect-aware-extension-gating-01-settings-connect-surface.png` — 5 validations passed

   

   - ✅ Script frame 1 narrated: "This is a fresh device: the org has not enabled Connect, and no legacy Google OAuth clie"
   - ✅ Script frame 2 narrated: "Now the organization flips one switch: connect enabled. This is the exact endpoint the d"
   - ✅ Script frame 3 narrated: "The tool surface transforms. The thirteen Google Workspace actions that could only fail "
   - ✅ Script frame 4 narrated: "If an agent still calls a hidden Google Workspace action, it no longer gets an OAuth err"
   - ✅ Script frame 5 narrated: "The status probe itself now carries the same guidance, so even an agent that only checks"
   - ✅ Script frame 6 narrated: "And it is a real kill switch: flip the flag off and the full legacy surface is back on t"
   - ✅ Script frame 7 narrated: "Behind it all is the surface users are actually pointed to: Settings > Connect, where Go"

benjaminshafii:
Superseded by a fresh PR against `dev` — base branch was deleted on merge of #2597 and GitHub can't reopen a PR whose base ref no longer exists. Same commits, retargeted below.

Issue #2341, opened 2026-06-22
[Bug]: Many color classes render nothing because they are not in the Tailwind palette


evanklem:
### Summary

`apps/app` replaces Tailwind's default `theme.colors` entirely with the Radix palette from `src/styles/tailwind-colors.ts`: 31 families, steps `1` to `12`.

Any class that uses a color outside that set compiles to no rule, so the element renders with no style: transparent background, no border, or text falling back to the inherited color.

A repo-wide sweep found **38 such classes across 13 files**, from two causes:

1. Families that do not exist in this palette: `emerald`, `zinc`, `neutral`.
   - Radix has no `emerald`; its greens are `green`, `jade`, `grass`, and `mint`.
   - `emerald` is used as the app's success green in about 8 files, and none of it renders.

2. Default Tailwind scales on otherwise valid families.
   - Radix only goes from `1` to `12`.
   - Classes like `amber-300`, `red-300`, `gray-100`, `sky-300`, etc. compile to nothing.

The `tailwindSafelist` only covers:

```txt
(bg|text|border)-(family)-(1 to 12)
```

So it catches neither case.

Found while reviewing #2329, the fix for #2328, which was `bg-zinc-950` in the theme previews. Same bug; this is the full list.

### To Reproduce

Clearest single instance, a `bg-*` miss that goes fully transparent:

1. Open the desktop app.
2. Open a local workspace.
3. In the sidebar, open the workspace `...` menu.
4. Select **Share**.
5. Select **Access workspace remotely**.
6. Look at the **Remote access** toggle when it is off by default.

The toggle track used `bg-zinc-300` and renders invisible, so only the white knob is visible.

Other visible spots:

- Warning and error banners in `shell/architecture-mismatch-gate.tsx` and `components/chat/message-list.tsx` use `amber-*` and `red-*` on the default `50`-`950` scale, so no background or border renders.
- Success and connected states using `emerald-*` lose their green styling, including toasts in `components/ui/sonner.tsx`, `messaging-view.tsx`, and `command-palette.tsx`.

Reproduce the full list with grep:

```bash
# Families that do not exist in the palette
grep -rnE "(bg|text|border|ring|from|via|to|fill|stroke)-(zinc|neutral|stone|emerald|rose|fuchsia)-[0-9]" apps/app/src

# Valid families using the invalid default 50-950 Tailwind scale
# Radix is 1 to 12
[...]

PR #2609, opened 2026-07-09
fix(app): show a check icon for completed workspace-creation steps

magic-peach:
## Summary

In the create-workspace progress list, `stepIcon` renders `XCircle` for the `done` state — the same glyph it uses for `error`. So every successfully completed step shows an ✕ instead of a check. On current `dev` it's compounded by the icon's `text-emerald-10` class, which doesn't exist in the app's Radix-only Tailwind palette (see #2341), so the ✕ also renders with no color of its own.

`apps/app/src/react-app/domains/workspace/create-workspace-local-panel.tsx`:

- `done` now renders `CircleCheck` — the same success glyph the notification center uses — in `text-green-10`, which exists in the palette.
- `error` keeps `XCircle` in red, so the two states are finally distinguishable at a glance.

## Testing

- `pnpm typecheck` — clean
- `bun test tests/` — 142 pass, 0 fail
- Two-line change confined to the `stepIcon` helper; the reviewer can reproduce by creating a local workspace and expanding the progress details while steps complete.

### DIFF
--- a/apps/app/src/react-app/domains/workspace/create-workspace-local-panel.tsx
+++ b/apps/app/src/react-app/domains/workspace/create-workspace-local-panel.tsx
@@ -81,7 +81,7 @@ export type CreateWorkspaceLocalPanelProps = {
 
 function stepIcon(status: CreateWorkspaceProgressStep["status"]) {
   if (status === "done")
-    return <CircleCheck size={16} className="text-green-10" />;
+    return <CircleCheck size={16} className="text-emerald-10" />;
   if (status === "active")
     return <Loader2 size={16} className="animate-spin text-dls-accent" />;
   if (status === "error") return <XCircle size={16} className="text-red-10" />;
--- a/apps/app/src/react-app/domains/workspace/create-workspace-local-panel.tsx
+++ b/apps/app/src/react-app/domains/workspace/create-workspace-local-panel.tsx
@@ -2,6 +2,7 @@
 import {
   ChartNoAxesColumnIncreasing,
   Check,
+  CircleCheck,
   FolderPlus,
   Loader2,
   XCircle,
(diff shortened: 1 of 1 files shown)


magic-peach:
@benjaminshafii could you please review my pr?

Pablosinyores (commented):
Good catch — this is a real bug. `done` was rendering `<XCircle>`, the same ✕ glyph as the `error` branch right below it, just tinted `text-emerald-10`. So a successfully completed step showed a green ✕, which reads as "failed but green" — confusing exactly where you want a clear success signal. Switching to `CircleCheck` (✓) fixes it and keeps the circular style consistent with the `XCircle` used for `error`.

One thing that's slightly out of scope: the tint also changes from `text-emerald-10` to `text-green-10`. The actual bug was the icon, not the color, so it's worth confirming `green-10` is a real token in the design system — the sibling branches use `text-red-10` and `text-dls-accent`, and `emerald-10` was clearly defined, but if `green-10` isn't in the palette the icon would fall back to an unstyled/default color and you'd trade one visual glitch for another. If `green-10` is intentional and defined, great; if you're unsure, keeping `emerald-10` would leave this scoped purely to the icon fix.

Minor: you're adding the `CircleCheck` import while `Check` is still imported at the top — if `Check` was only used for this `done` state it's now a dangling import (the linter will flag it); worth a quick check that it's still referenced elsewhere in the file.

The core fix is right regardless.

Iflaqbhat:
Re-checked after the update to keep `text-emerald-10`.

Verified:
- `pnpm --filter @openwork/app typecheck` — pass
- `git diff --check upstream/dev...HEAD` — pass

The fix is now nicely scoped: completed steps use `CircleCheck`, error steps still use `XCircle`, and the existing success color stays unchanged.

LGTM.

PR #2611, opened 2026-07-09
fix: handle GitHub connector branch refs and resyncs

### LINKED ISSUE #2533: [Bug]: GitHub Connector Cloud MCP — sync stuck on queued, setup fails with github_branch_not_found
## Environment
- OpenWork Cloud (app.openworklabs.com)
- GitHub Connector via Cloud MCP OAuth
- Desktop app v0.17.13
- Org: AlternaR&D

## Steps to reproduce
1. Create GitHub connector account via POST /v1/connectors/github/install/start
2. Install GitHub App on org (All repositories access)
3. Complete installation via POST /v1/connectors/github/install/complete
4. Try POST /v1/connectors/github/setup with installationId, repoId, branch="main", ref="main"

### 1. Setup fails
POST /v1/connectors/github/setup → {"error": "github_branch_not_found"}

### 2. Validate target contradiction
validate-target with ref="main" → branchExists: false (but repo IS accessible)
validate-target with ref="refs/heads/main" → branchExists: true
setup with ref="refs/heads/main" → Internal Server Error

### 3. Sync stays "queued" forever
POST /v1/connector-targets/{id}/resync → queued: true
Never progresses, no errors reported.

### 4. Discovery fails
GET /v1/connector-instances/{id}/discovery
→ {"error": "invalid_github_connector_target", "message": "missing repository, branch, or installation metadata"}

### 5. Account metadata shows repositories: []
Despite "All repositories" access on GitHub.

## Workaround
Creating config objects manually via POST /v1/config-objects and linking to a plugin bypasses the connector entirely. The connector discovery/sync flow appears broken on the backend.

## Key IDs
- Connector Account: cac_01kwx03jdheh4ba3efws7cypp6
- Connector Instance: cin_01kwx167t7eh4bbka0w28xp1j2
- GitHub Installation: 144874662

cyphercodes:
## Summary
- Normalize GitHub connector branch inputs so bare branch refs like `main` validate against `refs/heads/main`.
- Run manual GitHub connector resync/retry work immediately so sync events do not stay queued forever.

## Why
- Fixes the connector setup/resync failure reported in #2533 where `ref=main` fails validation and resync remains queued without progress.

## Issue
- Fixes #2533

## Scope
- GitHub connector setup validation, discovery target hydration, and manual resync/retry handling.
- Adds regression coverage for bare branch refs and repository identity validation.

## Out of scope
- No UI changes.
- No changes to GitHub webhook sync behavior.

### Ran
- `bun test ee/apps/den-api/test/github-connector-app.test.ts`
- `pnpm --filter @openwork-ee/den-api build`
- `git diff --check HEAD~1..HEAD`

### Result
- pass: all 8 targeted connector tests passed.
- pass: den-api build completed.
- pass: diff whitespace check completed.

## CI status
- pass: local targeted checks above.
- code-related failures: N/A locally.
- external/env/auth blockers: N/A.

## Manual verification
1. Verified the issue, comments, timeline, and duplicate PR searches before opening.
2. Verified the branch-ref helper accepts `branch=main` / `ref=main` and stores canonical `refs/heads/main`.
3. Verified manual resync/retry now executes the GitHub import path and records a terminal event status instead of leaving the event queued.

## Evidence
- N/A (API/backend regression tests)

## Risk
- Low/medium: changes are limited to GitHub connector setup and manual sync paths. Sync now runs inline for manual resync, so callers get the same response shape after the event has been processed.

## Rollback
- Revert this PR to restore previous queued-only manual resync behavior.

### DIFF
--- a/ee/apps/den-api/src/routes/org/plugin-system/github-app.ts
+++ b/ee/apps/den-api/src/routes/org/plugin-system/github-app.ts
@@ -79,6 +79,25 @@ export type GithubInstallStatePayload = {
 
 const GITHUB_API_BASE = "https://api.github.com"
 const GITHUB_API_VERSION = "2022-11-28"
+const GITHUB_HEAD_REF_PREFIX = "refs/heads/"
+
+export function normalizeGithubBranchRef(input: { branch: string; ref: string }) {
+  const rawBranch = input.branch.trim()
+  const rawRef = input.ref.trim()
+  const branch = rawBranch.startsWith(GITHUB_HEAD_REF_PREFIX)
+    ? rawBranch.slice(GITHUB_HEAD_REF_PREFIX.length).trim()
+    : rawBranch
+  if (!branch || !rawRef) {
+    return null
+  }
+
+  const ref = `${GITHUB_HEAD_REF_PREFIX}${branch}`
+  if (rawRef === branch || rawRef === ref) {
+    return { branch, ref }
+  }
+
+  return null
+}
 
 function base64UrlEncode(value: unknown) {
   const buffer = typeof value === "string"
@@ -695,8 +714,8 @@ export async function validateGithubInstallationTarget(input: {
     }
   }
 
-  const expectedRef = `refs/heads/${input.branch.trim()}`
-  if (input.ref.trim() !== expectedRef) {
+  const normalizedTarget = normalizeGithubBranchRef(input)
+  if (!normalizedTarget) {
     return {
       branchExists: false,
       defaultBranch,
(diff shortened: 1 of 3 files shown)


Pablosinyores (commented):
This reads well. `normalizeGithubBranchRef` correctly accepts both a bare `main` and a full `refs/heads/main` (stripping the prefix, rebuilding the canonical ref, and validating the caller's `ref` against either form), and threading the normalized branch through the branch-existence path and the `path`/`branchExists` checks keeps them consistent. The immediate-sync path is also careful: `autoImportError` is accumulated, the `catch` records it, and `eventStatus` resolves to `failed`/`completed` with `completedAt` set — so a failing manual sync ends in a terminal `failed` state rather than the "stuck forever" the queued path had. Good.

One design question on running the sync inline (`runGithubConnectorSyncNow` awaited directly in the route): the fix trades "sync event stays queued forever" for "the HTTP request now blocks until the whole import finishes." For a large repository that import could take a while — is there a bound on what a single manual resync pulls, or a request timeout in front of this route? If manual resync is always small/bounded, inline is fine; if it can be large, it might be worth keeping the work off the request path (respond `running`, do the import in the background) so a big repo doesn't time out the retry button. Not a blocker, just want to make sure the immediate path can't reintroduce a different failure mode.

Minor: the `eventStatus` nested ternary is a little hard to scan — a small helper or an explicit if/else would read easier for the next person, but purely cosmetic.

The test using both bare and fully-qualified refs is a good addition and covers the case #2533 hit.

PR #2685, opened 2026-07-12
fix(app): keep remote workspaces visible while reconnecting

reachjalil:

This is the foundation of a focused reliability series. Review and merge in this order:

1. **#2685 — retain remote workspaces through disconnects** (this PR)
2. **#2686 — route Settings task history to each workspace's owning server**
3. **#2687 — route Auto context compaction to the selected workspace endpoint**
4. **#2688 — keep Settings connection stores and reloads on the workspace server**

#2686 has two straightforward overlap conflicts with this PR in the shared route reconciliation code and will be rebased after this foundation lands. #2688 can merge independently and closes the reload-ownership gap required before #2687 is marked ready.

## Proposal

This PR makes workspace visibility resilient to temporary disconnects between the OpenWork client, its local OpenWork server, remote workers, and OpenWork Cloud.

It extends the same principle already used for durable Den authentication: **an unavailable dependency is unknown state, not proof that data was deleted**.

## The issue

Session and Settings each merge two workspace sources:

1. the desktop workspace store, which owns local and desktop-managed remote connection details; and
2. the connected OpenWork server's live workspace list.

On a successful refresh, that merge is correct. On a failed refresh, both routes replaced the last rendered list with only the desktop list. A remote workspace known only through the server therefore disappeared from the sidebar, then reappeared after the next successful refresh.

That made a transport failure look like deletion and could also move the user's selection.

## Before and after

| Event | Before | This PR |
| --- | --- | --- |
| Server or network disconnect | Server-sourced remote rows disappear. | Last successful rows remain visible. |
| Remote workspace during recovery | No stable transitional state. | The main Session sidebar shows `Remote · Reconnecting…`. |
| Cached tasks | Workspace replacement discards the visible grouping. | Cached sessions remain associated with the retained row while the worker is rechecked. |
| Reconnection | A later refresh makes rows reappear. | Online, focus, visibility, settings changes, and a bounded timer retry automatically. |
| Explicit removal during outage | A stale fallback could resurrect the row. | Removal updates persistent stores and the retained in-memory snapshot immediately. |

## Reconciliation contract

```mermaid
flowchart LR
  D["Desktop workspace store"] -->|routing fields| C["Client reconciliation"]
  S["OpenWork server"] -->|successful snapshot| C
  L["Last successful client snapshot"] -->|temporary fallback| C
  C --> U["Stable workspace sidebar"]
  S -. "timeout / disconnect" .-> R["Retain + mark reconnecting"]
  R --> U
  R -->|online, focus, or 30-second retry| S
```

The ordering is deliberate:

1. explicit user removal wins immediately;
2. fresh desktop-owned routing fields win for desktop-managed connections;
3. a successful server response may reconcile confirmed additions and removals; and
4. a transport failure retains the previous successful snapshot and cannot infer deletion.

The full explainer is in [`docs/remote-workspace-reconciliation.md`](https://github.com/reachjalil/openwork/blob/codex/remote-workspace-resilience/docs/remote-workspace-reconciliation.md).

## What changed

- Adds one shared failure-reconciliation helper used by Session and Settings.
- Preserves existing sessions and selection when the server is unavailable.
- Adds a distinct `reconnecting` workspace connection state and Session-sidebar label.
- Rechecks retained remote workspaces directly so an unavailable local server does not automatically make a healthy remote worker unusable.
- Retries failed refreshes on network restoration, app focus, visibility changes, settings changes, and every 30 seconds while failure remains known.
- Keeps the existing in-flight guard intact so the timer does not multiply slow requests within one route.
- Removes explicitly forgotten workspaces from the retained snapshot before another refresh can run.
- Exposes connection state through the existing route inspector for deterministic support diagnostics.

## Best-practice and security notes

- No new browser or disk cache is introduced; worker tokens and paths are not copied to another persistence layer.
- Existing desktop storage remains the authority for desktop-managed connection details.
- Specific worker failures, such as a rejected token, are preserved instead of being overwritten by a generic reconnecting state.
- Successful live responses remain authoritative, so real server-side removals converge normally after reconnection.
- No Den API, OpenWork server, IPC, database, or wire-contract change is required.

## Validation

- `pnpm --filter @openwork/app exec bun test tests/remote-workspace-resilience.test.ts tests/den-auth-provider.test.ts tests/settings-route.test.ts` — **10 passed**
- `pnpm --filter @openwork/app test` — **178 passed across 39 files**
- `pnpm --filter @openwork/app typecheck` — **passed**
- `pnpm --filter @openwork/app build` — **passed**
- `pnpm fraimz --flow remote-workspace-resilience` — **internal deterministic flow passed: 4 chapters plus voice-over coverage**

The current Fraimz flow runs unit tests and source-level assertions with `requiresApp: false`; it does not launch a real remote worker or produce a live disconnect screenshot. A reviewer can reproduce the internal proof with the command above. Live remote-worker evidence remains a follow-up review gate before this draft is marked ready.

## Scope and limitations

This PR protects workspace state already known to the running client. A completely fresh browser launch cannot reconstruct a server-only workspace while every server is unavailable. Desktop-managed remote workspaces remain recoverable on fresh launch because their existing desktop store is loaded before server reconciliation.

The reconnecting label is currently rendered in the main Session sidebar. Settings retains the workspace and cached tasks but does not yet render that same status label.

### DIFF
--- a/apps/app/src/app/types.ts
+++ b/apps/app/src/app/types.ts
@@ -202,7 +202,7 @@ export type SettingsTab = (typeof SETTINGS_TAB_VALUES)[number];
 
 export type WorkspacePreset = "starter" | "automation" | "minimal";
 
-export type WorkspaceConnectionStatus = "idle" | "connecting" | "connected" | "error";
+export type WorkspaceConnectionStatus = "idle" | "connecting" | "reconnecting" | "connected" | "error";
 
 export type WorkspaceConnectionState = {
   status: WorkspaceConnectionStatus;
--- a/apps/app/src/react-app/domains/session/sidebar/app-sidebar.tsx
+++ b/apps/app/src/react-app/domains/session/sidebar/app-sidebar.tsx
@@ -984,6 +984,9 @@ function WorkspaceSidebarGroup({
     if (connectionState.status === "error") return connectionState.message?.trim() || taskLoadError.message;
     if (group.status === "error") return taskLoadError.label;
     if (isConnectionActionBusy) return t("workspace_list.connecting");
+    if (isRemoteWorkspace && connectionState.status === "reconnecting") {
+      return `${t("workspace.remote_badge")} · ${t("settings.reconnecting")}`;
+    }
     if (isRemoteWorkspace && connectionState.status === "connected") return connectionState.message?.trim() || t("workspace_list.connected");
     if (!ctx.developerMode) return "";
     if (isSelected) return t("workspace.selected");
--- a/apps/app/src/react-app/shell/route-workspaces.ts
+++ b/apps/app/src/react-app/shell/route-workspaces.ts
@@ -7,7 +7,7 @@ import type { Session } from "@opencode-ai/sdk/v2/client";
 
 import type { OpenworkWorkspaceInfo } from "@/app/lib/openwork-server";
 import type { WorkspaceInfo } from "@/app/lib/desktop-types";
-import type { WorkspaceSessionGroup } from "@/app/types";
+import type { WorkspaceConnectionState, WorkspaceSessionGroup } from "@/app/types";
 import {
   normalizeDirectoryPath,
   normalizeSessionStatus,
@@ -30,6 +30,8 @@ export type RouteSession = Session & {
   runStatus?: unknown;
   slug?: string | null;
 };
+
+export const WORKSPACE_REFRESH_RETRY_INTERVAL_MS = 30_000;
 
 export function mapDesktopWorkspace(workspace: WorkspaceInfo): RouteWorkspace {
   return {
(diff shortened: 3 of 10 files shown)


reachjalil:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-12T06-04-54-741Z`

Full frame proof with validated screenshots: `evals/results/2026-07-12T06-04-54-741Z/fraimz.html` (re-run: `pnpm fraimz --flow remote-workspace-resilience`)

### ✅ remote-workspace-resilience — Remote workspaces remain stable through server disconnects and reconnect cleanly
_Internal demo_

1. **Local and remote workspace state begins from a verified reconciliation contract**
   > 🎙 A user starts with local and remote workspaces visible in the sidebar, including tasks loaded from a remote OpenWork server discovered through OpenWork Cloud.
   - ✅ The focused workspace reconciliation suite passes
   - ✅ Server-sourced remote retention is exercised
2. **A temporary disconnect retains remote rows and exposes reconnecting state**
   > 🎙 When the client temporarily loses its connection to Den or an OpenWork server, the remote workspace stays in place with its cached tasks and a clear Remote · Reconnecting status instead of disappearing.
   - ✅ Session retains its last successful snapshot on refresh failure
   - ✅ Settings uses the same failure reconciliation policy
   - ✅ The sidebar renders a distinct reconnecting state
3. **Unknown availability cannot delete data, while explicit removal still wins**
   > 🎙 OpenWork treats a failed refresh as unknown state, never as deletion: desktop connection details remain authoritative for routing, the last successful server snapshot remains visible, and only an explicit removal or a successful live response can remove a workspace.
   - ✅ Desktop-owned routing fields remain authoritative during fallback
   - ✅ Session removal updates the retained snapshot immediately
   - ✅ Settings removal updates the retained snapshot immediately
   - ✅ The conflict and deletion policy is documented
4. **OpenWork retries automatically without overlapping an in-flight refresh**
   > 🎙 When connectivity returns, OpenWork retries automatically, reconciles the live workspace list without duplicates or reordering, clears the warning, and restores normal task actions without requiring the user to add the remote workspace again.
   - ✅ Network restoration triggers a retry
   - ✅ Returning to the app triggers a retry
   - ✅ Background retry is bounded to a 30-second interval
   - ✅ Automatic retry respects the in-flight guard
   - ✅ Script frame 1 narrated: "A user starts with local and remote workspaces visible in the sidebar, including tasks l"
   - ✅ Script frame 2 narrated: "When the client temporarily loses its connection to Den or an OpenWork server, the remot"
   - ✅ Script frame 3 narrated: "OpenWork treats a failed refresh as unknown state, never as deletion: desktop connection"
   - ✅ Script frame 4 narrated: "When connectivity returns, OpenWork retries automatically, reconciles the live workspace"

reachjalil:
## Evidence clarification

The earlier automated Fraimz comment uses generic “validated screenshots” wording. This flow is actually an internal `requiresApp: false` proof backed by focused tests and source assertions; it does not launch a live remote worker or capture a disconnect screenshot.

The PR body now states that limitation explicitly. Current code checks are green on Linux and macOS, and a live worker/disconnect journey remains a gate before moving this draft to ready.

PR #2686, opened 2026-07-12
fix(app): keep remote tasks visible in Settings

reachjalil:
## Summary

Settings now loads each workspace's task history from the OpenWork server that owns that workspace, matching the main Session route. A remote workspace therefore keeps the same tasks when the user moves into Settings instead of being queried through (or discarded by) the local server.

This draft has been refreshed onto current `upstream/dev` (`ac1711fe9`) and is now reviewable independently; it no longer depends on the stale series ordering described in the previous body. Current head: `a16ef41d6`.

## What changed

- Removes Settings' local-server membership gate for merged desktop remote workspaces.
- Resolves every Settings workspace through `resolveWorkspaceEndpoint`.
- Binds the resolved client and server-side workspace ID into one shared `loadRouteWorkspaceSessions` request path used by Session and Settings.
- Keeps remote worker responses worker-scoped while preserving local directory filtering.
- Preserves the existing remote-worker diagnostic and recovery path.
- Preserves local client and host authentication for local workspace requests.
- Adds request-level regression coverage for exact local/remote URLs and credential boundaries.

## Behavior

| Scenario | Before | This PR |
| --- | --- | --- |
| Open Settings for a remote workspace | Settings asks the local server and may show no tasks. | Settings asks the owning worker and keeps its task history. |
| Remote workspace is absent from the local server list | A local membership gate silently returns an empty list. | The merged workspace resolves independently. |
| Worker uses a different server-side workspace ID | The desktop ID is sent to the wrong server. | The resolved worker ID is sent to the owning worker. |
| Remote sessions use worker-side directories | Valid sessions can be filtered out by the local path. | Worker-scoped results are retained. |
| Local server requires host authentication | Endpoint routing could drop the host credential. | Local client/host credentials stay on the local request. |

## Security and compatibility

- The saved worker token is sent only to the resolved remote worker.
- The local client token and host token are sent only to the local OpenWork server; the tests explicitly assert that the host token is absent from remote requests.
- No token is copied into a new cache or persistence layer.
- No Den API, OpenWork server, IPC, database, SDK, or wire-contract change is introduced.
- Missing or unreachable worker endpoints still use `diagnoseRemoteWorkspaceTaskLoadFailure`.

The detailed routing note is in [`docs/settings-remote-workspace-routing.md`](https://github.com/reachjalil/openwork/blob/codex/remote-settings-endpoint/docs/settings-remote-workspace-routing.md).

## Validation on refreshed head

[...]

PR #2687, opened 2026-07-12
fix(app): scope auto-compaction to the selected workspace

reachjalil:

This is the third PR in a focused reliability series:

1. **#2685 — retain remote workspaces through disconnects**
2. **#2686 — route Settings task history to each workspace's owning server**
3. **#2687 — route Auto context compaction to the selected workspace endpoint** (this PR)
4. **#2688 — keep Settings connection stores and reloads on the workspace server**

Review this after #2685, #2686, and #2688. The config read/write is independently scoped, while #2688 supplies the endpoint-owned reload coordinator required before this draft should be marked ready.

## What was wrong

Auto context compaction is workspace-owned configuration, but Preferences always read and wrote it through the local OpenWork client. For a remote workspace, the route combined a remote server-side workspace ID with the local server. That could show an unverified default, fail the update, or target unrelated local state.

The existing `autoCompactContextLoaded` flag also became `true` after a failed read and was not used to disable the switch. An unavailable worker therefore looked like a confirmed enabled preference.

## What changed

- resolve one endpoint target for the selected workspace and use its paired client plus server-side workspace ID for both config reads and writes
- keep local workspaces on the existing local endpoint
- enable the switch only after configuration loads for the exact endpoint currently selected
- ignore late reads after a workspace change, and prevent a late failed save from reverting the newly selected workspace's value
- leave the switch disabled when the initial authoritative load fails; there is no local fallback or speculative write
- isolate config parsing and transport behind focused helpers with characterization tests
- document the endpoint ownership, reconnect, credential, and compatibility boundaries

## Why this shape

The client and workspace ID are passed as one target so they cannot drift independently. Loaded state is tied to endpoint identity instead of a global boolean, which makes selection changes fail closed. Async completion guards prevent stale reads and failed saves from crossing workspace boundaries.

No Den, server, IPC, database, generated client, or wire contract changes are introduced. Remote credentials remain attached only to their existing remote client and are not copied, logged, cached, or sent to a fallback endpoint.

## Validation

- `pnpm --filter @openwork/app test` — 177 passed, 0 failed
- focused route and remote auto-compaction tests — 4 passed, 0 failed
- `pnpm --filter @openwork/app build` — passed
- `pnpm fraimz --flow remote-auto-compaction` — internal deterministic flow passed: four behavior steps plus voice-over coverage

The current Fraimz flow uses focused tests and source assertions with `requiresApp: false`; it does not launch a real remote worker or capture the Preferences switch. Live remote-worker evidence remains a follow-up review gate before this draft is marked ready.

## Review guide

Start with `docs/remote-auto-compaction.md` for the read/write ownership model, then review `auto-compact-context.ts` and the Preferences effect/toggle in `settings-route.tsx`.

## Known follow-up before ready

#2688 captures the originating endpoint for pending reloads and prevents remote failures from using the local desktop restart fallback. Rebase this PR after #2688 and route its reload request through that captured-target helper before promoting it from draft status.

### DIFF
--- a/apps/app/src/react-app/domains/settings/auto-compact-context.ts
+++ b/apps/app/src/react-app/domains/settings/auto-compact-context.ts
@@ -1,0 +1,37 @@
+export type AutoCompactContextConfig = {
+  opencode?: {
+    compaction?: unknown;
+  };
+};
+
+export type AutoCompactContextClient = {
+  getConfig: (workspaceId: string) => Promise<AutoCompactContextConfig>;
+  patchConfig: (
+    workspaceId: string,
+    patch: { opencode: { compaction: { auto: boolean } } },
+  ) => Promise<unknown>;
+};
+
+export type AutoCompactContextTarget = {
+  client: AutoCompactContextClient;
+  workspaceId: string;
+};
+
+export function readAutoCompactContextValue(config: AutoCompactContextConfig): boolean {
+  const compaction = config.opencode?.compaction;
+  if (!compaction || typeof compaction !== "object") return true;
+  return Reflect.get(compaction, "auto") !== false;
+}
+
+export async function loadAutoCompactContext(target: AutoCompactContextTarget): Promise<boolean> {
+  return readAutoCompactContextValue(await target.client.getConfig(target.workspaceId));
+}
+
+export async function saveAutoCompactContext(
+  target: AutoCompactContextTarget,
+  auto: boolean,
+): Promise<void> {
+  await target.client.patchConfig(target.workspaceId, {
+    opencode: { compaction: { auto } },
+  });
+}
--- a/apps/app/src/react-app/domains/settings/pages/preferences-view.tsx
+++ b/apps/app/src/react-app/domains/settings/pages/preferences-view.tsx
@@ -34,6 +34,7 @@ export type PreferencesViewProps = {
   onToggleShowThinking: () => void;
   autoCompactContext: boolean;
   autoCompactContextBusy: boolean;
+  autoCompactContextLoaded: boolean;
   onToggleAutoCompactContext: () => void;
   analyticsEnabled: boolean;
   onToggleAnalytics: () => void;
@@ -93,7 +94,7 @@ export function PreferencesView(props: PreferencesViewProps) {
               <Switch
                 aria-label={t("settings.auto_compact")}
                 checked={props.autoCompactContext}
-                disabled={props.busy || props.autoCompactContextBusy}
+                disabled={props.busy || props.autoCompactContextBusy || !props.autoCompactContextLoaded}
                 onCheckedChange={props.onToggleAutoCompactContext}
               />
             </LayoutSectionItemHeaderActions>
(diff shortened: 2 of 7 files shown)


reachjalil:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-12T06-20-27-490Z`

Full frame proof with validated screenshots: `evals/results/2026-07-12T06-20-27-490Z/fraimz.html` (re-run: `pnpm fraimz --flow remote-auto-compaction`)

### ✅ remote-auto-compaction — Auto-compaction follows the selected workspace's owning endpoint
_Internal demo_

1. **A remote workspace starts from its authoritative saved value**
   > 🎙 A user opens Preferences for a remote workspace; Auto context compaction stays disabled while OpenWork loads that worker’s actual configuration, then shows the saved remote value.
   - ✅ The focused auto-compaction suite passes
   - ✅ The default parsing contract is exercised
   - ✅ Endpoint-owned loading is exercised
2. **The selected endpoint owns both sides of the preference update**
   > 🎙 Toggling Auto context compaction writes `opencode.compaction.auto` through the selected workspace endpoint and server-side workspace ID, without reading from or mutating the local workspace configuration.
   - ✅ Settings loads through the resolved endpoint
   - ✅ Settings saves through that same endpoint
   - ✅ The client and server-side workspace ID remain paired
3. **Only the current endpoint can enable or update the switch**
   > 🎙 Switching between local and remote workspaces reloads the preference from each owning server, and a late response from the previous workspace cannot overwrite the newly selected workspace’s state.
   - ✅ Loaded state is tied to exact endpoint identity
   - ✅ Late responses are ignored after selection changes
   - ✅ A late failed save cannot revert another workspace's value
   - ✅ The switch stays disabled until the current endpoint is loaded
4. **An unavailable worker cannot produce an unverified write or local fallback**
   > 🎙 If the remote worker is unavailable, the switch remains disabled instead of presenting an unverified default or attempting a local fallback; reconnecting and refreshing restores the authoritative remote value.
   - ✅ Writes require a loaded current endpoint
   - ✅ Fail-closed routing is documented
   - ✅ The credential boundary is documented
   - ✅ Script frame 1 narrated: "A user opens Preferences for a remote workspace; Auto context compaction stays disabled "
   - ✅ Script frame 2 narrated: "Toggling Auto context compaction writes `opencode.compaction.auto` through the selected "
   - ✅ Script frame 3 narrated: "Switching between local and remote workspaces reloads the preference from each owning se"
   - ✅ Script frame 4 narrated: "If the remote worker is unavailable, the switch remains disabled instead of presenting a"

reachjalil:
## Evidence clarification

The earlier automated Fraimz comment uses generic “validated screenshots” wording. This flow is actually an internal `requiresApp: false` proof backed by focused tests and source assertions; it does not launch a live remote worker or capture the Preferences switch.

The PR body now states that limitation and links #2688 as the endpoint-owned reload prerequisite. Current code checks are green on Linux and macOS; rebase on #2688 plus live remote-worker evidence remain gates before moving this draft to ready.

PR #2688, opened 2026-07-12
fix(app): keep Settings connections on the workspace server

reachjalil:

This is an independent follow-up to #2685, #2686, and #2687. It can merge directly against `dev`, but it also closes #2687's known reload-ownership gap and should land before #2687 is promoted from draft.

## What was wrong

Settings already resolved a remote OpenCode client for the selected workspace, but its MCP and provider stores still received the local host's OpenWork server store. That mixed two authorities: a remote server-side workspace ID and the local server client.

The mismatch could show local connection state inside a remote workspace, write connection/provider configuration to the wrong server, or service a remote reload through the local desktop engine. A delayed reload could also follow a later workspace selection instead of the workspace that originated the mutation.

## What changed

- introduces one narrow selected-workspace OpenWork server projection containing the endpoint client, capabilities, base URL, token presence, and remote/local ownership
- passes that endpoint-backed projection to Settings' MCP and provider stores
- updates the extension-store connection override from the same selected endpoint
- captures the endpoint client and server-side workspace ID together when a config mutation requests a reload
- makes the captured mutation target win over later workspace selection when the reload executes
- prevents remote engine failures from invoking the local desktop restart fallback
- shares the same projection with Session provider auth, replacing its parallel partial adapter
- fails closed when no selected workspace endpoint is available

## Ownership and ordering

```mermaid
flowchart LR
  W["Selected workspace"] --> E["resolveWorkspaceEndpoint"]
  E --> P["Workspace OpenWork snapshot"]
  P --> M["MCP store"]
  P --> A["Provider store"]
  P --> X["Extension store"]
  P --> R["Captured reload target"]
  R -->|local| L["Local server / allowed desktop fallback"]
  R -->|remote| Q["Owning worker / no local fallback"]
```

The client and workspace ID are captured at mutation time. Reload execution therefore cannot drift to whichever workspace happens to be selected later.

## Security and compatibility

- remote credentials remain attached to the existing endpoint client and are never substituted with the local token
- no token value is logged or copied into a new cache
- remote failures cannot restart the local desktop engine
- no Den, server, IPC, database, generated-client, or wire-contract changes are introduced

See `docs/remote-settings-connections.md` for the full design and failure behavior.

## Validation

- `pnpm --filter @openwork/app test` — **177 passed, 0 failed**
- focused Settings/MCP/provider suite — **16 passed, 0 failed**
- `pnpm --filter @openwork/app typecheck` — **passed**
- `pnpm --filter @openwork/app build` — **passed**
- `pnpm fraimz --flow remote-settings-connections` — **internal deterministic flow passed: 4 chapters plus voice-over coverage**

The Fraimz flow uses focused tests and source-level assertions with `requiresApp: false`; it does not claim a live remote-worker screenshot. A live worker/disconnect journey remains a review gate before this draft is marked ready.

### DIFF
--- a/apps/app/src/react-app/domains/connections/provider-auth/store.ts
+++ b/apps/app/src/react-app/domains/connections/provider-auth/store.ts
@@ -34,21 +34,17 @@ import {
 } from "../../../../app/utils/providers";
 import { getReactQueryClient } from "../../../infra/query-client";
 import { ensureProviderListQuery } from "../../../infra/provider-list-query";
-import type { OpenworkServerStoreSnapshot } from "../openwork-server-store";
+import {
+  canFallbackToDesktopEngineRestart,
+  type WorkspaceOpenworkServer,
+} from "../workspace-openwork-server";
 
 /**
- * The slice of the openwork-server store this store actually consumes.
- * The settings route passes the full store; the session route passes a
- * lightweight endpoint-backed adapter (previously forced through `as never`).
+ * The workspace-owned OpenWork connection consumed by provider actions.
+ * Both Settings and Session pass endpoint-backed adapters so a remote
+ * workspace can never inherit the local host client's authority.
  */
-export type ProviderAuthOpenworkServer = {
-  getSnapshot: () => Pick<
-    OpenworkServerStoreSnapshot,
-    "openworkServerStatus" | "openworkServerClient"
-  > & {
-    openworkServerCapabilities: { config?: { read?: boolean; write?: boolean } } | null;
-  };
-};
+export type ProviderAuthOpenworkServer = WorkspaceOpenworkServer;
 import {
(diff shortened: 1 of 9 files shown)


reachjalil:
## Deterministic regression proof — passed

`pnpm fraimz --flow remote-settings-connections`

- selected remote endpoint projection: passed
- MCP/provider stores share the endpoint-backed server: passed
- pending reload retains its originating client and workspace ID: passed
- remote failures cannot invoke the local desktop restart fallback: passed
- voice-over coverage: passed

This is an internal `requiresApp: false` proof backed by focused tests and source assertions. It does **not** claim a live remote-worker screenshot. Re-run locally with the command above; the full app suite also passed 177/177 tests, followed by app typecheck and production build.

reachjalil:

PR #2689, opened 2026-07-12
fix(app): discard stale Den cloud snapshots

reachjalil:
> [!WARNING]
> **Queue audit (2026-07-17): superseded by #2869 — do not merge this revision.** The stale-org snapshot race is real in the current desktop Cloud sync path, but ready upstream PR #2869 deletes that path and `apps/app/src/app/cloud/desktop-cloud-sync.ts` entirely as part of Connect-only Marketplace delivery. Hardening code scheduled for removal would create immediate conflict and no durable value. Reassess identity scoping only if #2869 does not land or an equivalent sync path returns.

## Problem

Desktop cloud sync captured the active Den organization before requesting its resource snapshot, but it did not verify that the same Den session was still active after the request completed. A slow organization A response could therefore be applied to the workspace after the user had already switched to organization B.

A fetched Den snapshot may reach workspace sync only while the exact request identity is still current:

- Den base URL is unchanged
- authentication token identity is unchanged
- active organization is unchanged
- authentication is still present

If any part changes while the request is in flight, the response is stale and is discarded. A later sync for the current identity remains responsible for applying current state.

## Implementation

- Capture a small in-memory identity before loading the snapshot.
- Re-read that identity immediately before applying the result.
- Centralize the equality and identity-bound apply behavior in testable helpers.
- Keep credentials opaque: the token is compared in memory and is never logged or newly persisted.
- Preserve the existing serialized desktop cloud sync queue and missing-auth no-op behavior.

## Regression proof

- Added a delayed-promise test that starts organization A, switches to B before A resolves, and proves apply is never called.
- Added the positive counterpart proving an unchanged identity still applies normally.
- Added coverage for base URL, token, organization, and missing-auth identity changes.
- Added an approved voiceover and an internal fraimz flow for the ordering contract.

## Validation

- pnpm --filter @openwork/app test:desktop-cloud-sync — 11 passed, 0 failed
- pnpm --filter @openwork/app build — passed (existing bundle-size and module-directive warnings only)
- pnpm fraimz --flow den-cloud-sync-identity — Passed, 4 proof frames plus voiceover coverage

## Evidence limitations

The fraimz run is an internal, test-backed ordering proof. It does not claim to be a live UI recording or a real multi-organization Den session. No production Den credentials or tenant data were used.

- comparing the full opaque token in memory is the desired session-identity boundary;
- discarding a stale response, rather than automatically scheduling another sync here, composes correctly with the existing event/interval triggers;
- the narrow check immediately before workspace apply is sufficient for the intended session-switch semantics;
- a manual desktop pass with two real test organizations shows no temporary cross-organization provider/plugin/config state;
- the branch is still appropriate against the latest dev head when reviewed.

## Scope and mergeability

This is an independent fix based on dev. It has no ordering dependency on the other remote-workspace or Den-state reliability drafts and should be reviewed and merged on its own merits.

### DIFF
--- a/apps/app/scripts/desktop-cloud-sync.test.ts
+++ b/apps/app/scripts/desktop-cloud-sync.test.ts
@@ -1,9 +1,12 @@
 import { describe, expect, test } from "bun:test";
 
 import {
+  applyIdentityBoundSnapshot,
   derivePendingCloudPluginChanges,
+  isSameDesktopCloudSyncIdentity,
   readPendingCloudSyncChanges,
 } from "../src/app/cloud/desktop-cloud-sync";
+import type { DesktopCloudSyncIdentity } from "../src/app/cloud/desktop-cloud-sync";
 import type { OpenworkDesktopCloudSyncChange } from "../src/app/lib/openwork-server";
 
 function change(input: Partial<OpenworkDesktopCloudSyncChange> & Pick<OpenworkDesktopCloudSyncChange, "id" | "kind" | "resourceKind">): OpenworkDesktopCloudSyncChange {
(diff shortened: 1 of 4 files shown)


reachjalil:

reachjalil:
Closing during the 2026-07-27 repository-progress reconciliation. Explicitly superseded by #2869 in the PR's own queue audit; the changed desktop Cloud sync path was scheduled for removal. No unmerged branch is being treated as delivered; any residual gap must be reproduced before new implementation.

PR #2690, opened 2026-07-12
fix(app): coalesce Den sync requests

reachjalil:
## Problem

The provider auto-sync hook prevented overlapping network requests with an in-flight boolean. That guard also discarded every Den settings-change event received while a request was running. Switching organizations at the wrong moment could therefore leave providers and cloud MCP state on the previous settings until the periodic interval fired.

- Never run more than one reconciliation at a time.
- While one is running, retain one pending pass for the newest settings state.
- Coalesce a burst into one immediate follow-up, using the latest reason.
- Cancel pending work on sign-out or unmount.
- A failed pass must release the runner and allow already-pending current state to reconcile.

## Implementation

- Replaced the drop-on-in-flight guard with a small serialized sync runner.
- Each signal updates one pending reason; the drain loop consumes it after the active request settles.
- The hook still keeps the latest sync closure without restarting its interval on render.
- Cancellation clears pending work but does not attempt to abort an already-owned request.
- Errors remain owned by the existing stores and do not create an automatic retry loop.

## Regression proof

Tests hold the first request open, send several settings changes, and prove:

- no second request starts concurrently;
- the burst produces exactly one follow-up using the newest reason;
- cancellation removes the pending pass;
- a failed first request still releases the queued current-state pass.

## Validation

- bun test apps/app/scripts/cloud-provider-auto-sync.test.ts — 3 passed, 0 failed
- pnpm fraimz --flow den-sync-coalesce — Passed, 4 proof frames plus voiceover coverage

## Evidence limitations

The fraimz run is an internal concurrency proof backed by deterministic deferred promises. It is not a live UI recording and does not exercise a production Den tenant or real network latency.

- latest-reason coalescing is correct for both provider sync consumers mounted by Settings;
- one follow-up per in-flight window is the desired behavior under sustained settings-event traffic;
- clearing pending work without aborting the active request is correct on sign-out/unmount;
- swallowed promise failures remain correctly surfaced by each owning store;
- a manual organization switch during a deliberately slow sync refreshes current providers and MCP state immediately after the active request settles;

## Scope and mergeability

This is independent from the stale-snapshot identity guard and other remote reliability drafts. It can be reviewed and merged separately.

### DIFF
--- a/apps/app/scripts/cloud-provider-auto-sync.test.ts
+++ b/apps/app/scripts/cloud-provider-auto-sync.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, test } from "bun:test";
+
+import { createCloudProviderSyncRunner } from "../src/react-app/domains/cloud/use-cloud-provider-auto-sync";
+import type { CloudProviderSyncReason } from "../src/react-app/domains/cloud/use-cloud-provider-auto-sync";
+
+function deferred() {
+  let resolve: () => void = () => undefined;
+  const promise = new Promise<void>((done) => {
+    resolve = done;
+  });
+  return { promise, resolve };
+}
+
+describe("createCloudProviderSyncRunner", () => {
+  test("coalesces settings changes into one follow-up without overlap", async () => {
+    const first = deferred();
+    const reasons: CloudProviderSyncReason[] = [];
+    let active = 0;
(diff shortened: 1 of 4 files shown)


reachjalil:

PR #2691, opened 2026-07-12
fix(app): cancel stale onboarding redirects

reachjalil:
## Problem

After a successful Den sign-in, the shell recursively scheduled up to ten organization checks. Cleanup removed only the session event listener; it did not cancel those timers. Repeated success events could create overlapping polling chains, and an older chain could redirect to onboarding after sign-out, navigation, a newer sign-in, or unmount.

- Every session event invalidates any older redirect wait.
- Only a success event with a token starts a new bounded wait.
- The current stored token must still match that success event when the organization appears.
- A matching token plus active organization may navigate once.
- Sign-out, error, newer success, and unmount cancel pending timers.
- The existing ten checks at 500ms remain the maximum wait; expiry is silent.

## Implementation

- Extracted a small cancellable redirect scheduler with a generation boundary.
- Captured the successful sign-in token and compare it with current Den settings before navigation.
- Clear the active timer and invalidate the generation on every session event and effect cleanup.
- Mark the generation complete before navigating so no duplicate callback can redirect twice.

## Regression proof

Deterministic fake-timer tests prove:

- two successful sign-ins leave only the latest eligible to navigate;
- a matching current token and organization navigate exactly once;
- sign-out/unmount cancellation leaves no timer or navigation;
- a different current token cannot redirect;
- the bounded wait expires without late navigation.

## Validation

- bun test apps/app/scripts/den-onboarding-redirect-order.test.ts — 4 passed, 0 failed
- pnpm fraimz --flow den-onboarding-redirect-order — Passed, 4 proof frames plus voiceover coverage

## Evidence limitations

The fraimz run is an internal scheduler proof with deterministic timers. It is not a live UI recording and does not exercise browser handoff timing against a real Den deployment.

- the session event token is the correct identity to bind to the redirect;
- all success producers include the token required to start the wait;
- cancelling on every non-success session event matches intended error and sign-out behavior;
- retaining the existing five-second maximum is still the desired UX;
- a manual desktop test covering success, rapid second sign-in, sign-out during org loading, and route navigation produces no surprise redirect;

## Scope and mergeability

This is an independent shell ordering fix with no merge dependency on the other Den synchronization drafts.

### DIFF
--- a/apps/app/scripts/den-onboarding-redirect-order.test.ts
+++ b/apps/app/scripts/den-onboarding-redirect-order.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, test } from "bun:test";
+
+import { createOnboardingRedirectScheduler } from "../src/react-app/shell/app-root";
+
+function harness(maxAttempts = 10) {
+  let identity = { authToken: "token-a", activeOrgId: "" };
+  let navigations = 0;
+  let nextHandle = 1;
+  const callbacks = new Map<number, () => void>();
+  const scheduler = createOnboardingRedirectScheduler({
+    readIdentity: () => identity,
+    navigate: () => {
+      navigations += 1;
+    },
+    schedule: (callback) => {
+      const handle = nextHandle++;
+      callbacks.set(handle, callback);
+      return handle;
+    },
+    clear: (handle) => callbacks.delete(handle),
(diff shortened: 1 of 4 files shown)


reachjalil:

PR #2692, opened 2026-07-12
fix(app): ignore stale Den organization refreshes

reachjalil:
## Problem

Organization refreshes had no request ownership. If a session, credential, or Den endpoint changed while listOrgs was in flight, an older response could arrive last and overwrite the newer organization list, selected organization, persisted Den metadata, active-organization context, loading state, or error state.

- Each refresh receives a monotonically newer request generation.
- Only the latest generation may mutate organization success state.
- Authentication token, Den endpoint, or client changes invalidate in-flight generations.
- Error and completion state obey the same latest-request rule as successful data.
- Starting a newer refresh keeps loading true when an older request finishes.

## Implementation

- Added a small latest-request gate owned by the hook instance.
- Begin a generation before every refresh, including the signed-out clear path.
- Recheck ownership immediately after listOrgs and after active-organization synchronization.
- Guard catch and finally so stale failure/completion cannot cover current data or clear current loading state.
- Invalidate the gate whenever token, base URL, or Den client identity changes.

## Regression proof

Focused tests prove:

- a second refresh immediately makes the first stale;
- changing session identity invalidates the active refresh;
- stale success, error, and completion mutations are all rejected while the newest success/completion remain allowed.

## Validation

- bun test apps/app/scripts/den-org-refresh-order.test.ts — 3 passed, 0 failed
- pnpm fraimz --flow den-org-refresh-order — Passed, 4 proof frames plus voiceover coverage

## Evidence limitations

The fraimz run is an internal request-ownership proof. It does not claim to be a live UI recording, and it does not inject delayed responses into a real Den endpoint or verify persisted browser storage visually.

- token, base URL, and client are the complete session identity boundary for organization refresh ownership;
- every organization-state mutation in refreshOrgs is behind the latest-generation check;
- allowing already-current response mutations before ensureDenActiveOrganization completes is the desired sequence;
- busy state remains correct across manual refresh, automatic refresh, sign-out, and endpoint changes;
- a manual delayed-response test across two accounts/endpoints confirms the older response cannot replace list, selection, persistence, context, error, or loading state;

## Scope and mergeability

This is an independent latest-request-wins fix. It has no merge-order dependency on the other Den synchronization or remote-workspace drafts.

### DIFF
--- a/apps/app/scripts/den-org-refresh-order.test.ts
+++ b/apps/app/scripts/den-org-refresh-order.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, test } from "bun:test";
+
+import { createLatestRequestGate } from "../src/react-app/domains/settings/cloud/use-den-session";
+
+describe("createLatestRequestGate", () => {
+  test("only the newest organization refresh remains current", () => {
+    const gate = createLatestRequestGate();
+    const older = gate.begin();
+    const newer = gate.begin();
+
+    expect(gate.isCurrent(older)).toBe(false);
+    expect(gate.isCurrent(newer)).toBe(true);
+  });
+
+  test("a session identity change invalidates an in-flight refresh", () => {
+    const gate = createLatestRequestGate();
+    const request = gate.begin();
+    gate.invalidate();
+
+    expect(gate.isCurrent(request)).toBe(false);
+  });
+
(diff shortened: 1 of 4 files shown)


reachjalil:

PR #2734, opened 2026-07-14
fix(server): budget tool image request payloads

benjaminshafii:
## Summary
- add a request-only OpenCode message transform that enforces a 10 MiB aggregate encoded inline-image budget
- keep newest user/tool images and replace omitted or malformed images with compact text placeholders
- protect replayed history without mutating persisted sessions or adding native image dependencies
- register and package the plugin in every OpenWork runtime config

## Root cause
The reported malformed-body/invalid-JSON symptom is caused by aggregate base64 media exceeding provider/proxy request limits. OpenCode already serializes JSON and normalizes each image individually, but had no aggregate request budget.

## Daytona validation
- `pnpm --filter openwork-server exec bun test src/opencode-plugins/openwork-media-budget.test.ts src/openwork-runtime-config.test.ts` — 14 passed
- `pnpm --filter openwork-server build` — passed; plugin bundled to `dist/opencode-plugins/openwork-media-budget.js`
- `pnpm fraimz --flow tool-image-request-budget` inside Daytona sandbox `openwork-test-20260713-225226` — passed, 3 internal proof frames

Fraimz run: `2026-07-14T06-01-21-525Z`

## Validation limitation
The deterministic Fraimz uses the real plugin and runtime config with a fake provider budget probe; it does not call a paid external model API.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -10,7 +10,7 @@
     "dev": "OPENWORK_DEV_MODE=1 bun src/cli.ts",
     "test": "bun test",
     "test:artifacts": "bun test src/artifact-files.e2e.test.ts src/workspace-init.test.ts src/server.normalizeWorkspaceRelativePath.test.ts",
-    "build": "tsc -p tsconfig.json && bun build src/opencode-plugins/openwork-extensions-preview.ts src/opencode-plugins/openwork-capabilities-knowledge.ts src/opencode-plugins/openwork-anthropic-adaptive-thinking.ts src/opencode-plugins/openwork-anthropic-tool-schema.ts --outdir dist/opencode-plugins --target node --format esm",
+    "build": "tsc -p tsconfig.json && bun build src/opencode-plugins/openwork-extensions-preview.ts src/opencode-plugins/openwork-capabilities-knowledge.ts src/opencode-plugins/openwork-anthropic-adaptive-thinking.ts src/opencode-plugins/openwork-anthropic-tool-schema.ts src/opencode-plugins/openwork-media-budget.ts --outdir dist/opencode-plugins --target node --format esm",
     "build:bin": "bun build --compile src/cli.ts --outfile dist/bin/openwork-server",
     "build:bin:all": "bun ./script/build.ts --outdir dist/bin --target bun-darwin-arm64 --target bun-darwin-x64 --target bun-linux-x64 --target bun-linux-arm64 --target bun-windows-x64 --target bun-windows-arm64",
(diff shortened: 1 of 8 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-14T06-07-03-009Z`

Full frame proof with validated screenshots: `evals/results/2026-07-14T06-07-03-009Z/fraimz.html` (re-run: `pnpm fraimz --flow tool-image-request-budget`)

### ✅ tool-image-request-budget — Runtime media budget omits old inline tool images before provider serialization
_Internal demo_

1. **A replayed user image plus tool images can exceed the request image budget before serialization**
   > 🎙 The demo starts with the regression shape: a normal user image plus a tool result that returns multiple inline images. With the eval budget lowered, the unprotected provider request is too large and the fake provider reports the same malformed-body class of failure users saw.
   - ✅ The unprotected request is rejected by the fake provider
   - ✅ Encoded inline image bytes exceed the eval request budget
   - ✅ The simulated failure uses the reported malformed-body class
2. **The transformed provider body is valid JSON under budget with text placeholders for omitted images**
   > 🎙 The real OpenWork runtime config includes the media-budget plugin path, and the package build includes the built plugin. When the hook runs on the request copy, the newest image stays available, older images become clear text notes, and the next provider body is valid JSON under the encoded inline-image budget.
   - ✅ The real OpenWork runtime config registers the media-budget plugin
   - ✅ The plugin path helper resolves a built dist plugin to .js
   - ✅ The server build bundles openwork-media-budget into dist/opencode-plugins
   - ✅ The transformed request is accepted by the fake provider
   - ✅ The transformed provider body serializes as valid JSON
   - ✅ Remaining encoded inline image bytes stay within budget
   - ✅ Only the newest image attachment remains
   - ✅ Omitted images are communicated with compact text placeholders
3. **A second request transform leaves the already-cleaned history unchanged**
   > 🎙 The same transformed history is replayed through the hook again. Nothing changes on the second pass, proving the request cleanup is deterministic and safe for session history replay.
   - ✅ Repeated transforms are deterministic and idempotent
   - ✅ Placeholders are not duplicated or lost on replay
   - ✅ Script frame 1 narrated: "The demo starts with the regression shape: a normal user image plus a tool result that r"
   - ✅ Script frame 2 narrated: "The real OpenWork runtime config includes the media-budget plugin path, and the package "
   - ✅ Script frame 3 narrated: "The same transformed history is replayed through the hook again. Nothing changes on the "

PR #2738, opened 2026-07-14
fix(helm): preserve Node options on upgrade

OmarMcAdam:
## Summary
- render `config.denApiNodeOptions` as a direct `NODE_OPTIONS` value while explicitly clearing the 0.17.24 `valueFrom` field
- preserve existing `denApi.env.NODE_OPTIONS` values without requiring operator config changes
- add regression coverage and document legacy-value precedence

## Tests
- `helm lint packaging/helm/openwork-ee`
- `bash packaging/helm/openwork-ee/tests/observability-env.sh`
- `helm template openwork-ee packaging/helm/openwork-ee --set-string 'config.denApiNodeOptions=--tls-max-v1.2' | kubectl create --dry-run=client --validate=false -f - -o name`\n- `git diff --check`\n- `pnpm fraimz --flow helm-den-api-node-options` (Passed)

### DIFF
--- a/evals/flows/helm-den-api-node-options.flow.mjs
+++ b/evals/flows/helm-den-api-node-options.flow.mjs
@@ -58,14 +58,67 @@ export default {
             );
             witness(
               ctx,
-              rendered.includes("key: DEN_API_NODE_OPTIONS"),
-              "NODE_OPTIONS reads DEN_API_NODE_OPTIONS from the ConfigMap",
+              rendered.includes('- name: NODE_OPTIONS\n              value: "--use-openssl-ca --max-old-space-size=4096"'),
+              "Den API receives the configured value directly as NODE_OPTIONS",
+            );
+            witness(
+              ctx,
+              rendered.includes("valueFrom: null") && !rendered.includes("key: DEN_API_NODE_OPTIONS"),
+              "The chart explicitly clears the previous valueFrom reference during upgrades",
             );
             ctx.output(
               "$ helm template openwork-ee ... --set-string config.denApiNodeOptions=...",
               rendered
                 .split("\n")
                 .filter((line) => line.includes("DEN_API_NODE_OPTIONS") || line.includes("NODE_OPTIONS"))
+                .join("\n"),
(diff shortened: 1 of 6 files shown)


OmarMcAdam:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-14T08-48-21-610Z`

Full frame proof with validated screenshots: `evals/results/2026-07-14T08-48-21-610Z/fraimz.html` (re-run: `pnpm fraimz --flow helm-den-api-node-options`)

### ✅ helm-den-api-node-options — Helm passes configured Node.js options to Den API startup
_Internal demo_

1. **Configured Den API Node.js flags reach the Node process at startup**
   > 🎙 An operator keeps the same Den API Node.js flags in the Helm values. The Den API deployment gives Node that value directly as NODE_OPTIONS, so an upgrade from an existing deployment remains valid without changing operator configuration.
   - ✅ helm template exits successfully
   - ✅ The ConfigMap contains DEN_API_NODE_OPTIONS
   - ✅ Only Den API receives NODE_OPTIONS
   - ✅ Den API receives the configured value directly as NODE_OPTIONS
   - ✅ The chart explicitly clears the previous valueFrom reference during upgrades
2. **Existing denApi.env.NODE_OPTIONS values upgrade without configuration changes**
   > 🎙 Deployments already using denApi.env.NODE_OPTIONS keep that exact setting during the upgrade. The chart renders one direct NODE_OPTIONS value and never combines it with valueFrom.
   - ✅ helm template exits successfully
   - ✅ The rendered deployment contains one NODE_OPTIONS entry
   - ✅ The existing denApi.env.NODE_OPTIONS value remains unchanged
   - ✅ The legacy direct value is not combined with valueFrom
   - ✅ Script frame 1 narrated: "An operator keeps the same Den API Node.js flags in the Helm values. The Den API deploym"
   - ✅ Script frame 2 narrated: "Deployments already using denApi.env.NODE_OPTIONS keep that exact setting during the upg"

PR #2773, opened 2026-07-15
fix(mcp): retry and surface silent search_capabilities injection failures

reachjalil:
This PR prevents a Cloud MCP tool-injection problem from looking like `search_capabilities` simply disappeared after an app restart. If the search tool is not available for the active workspace, provider, and model, OpenWork now treats that as a real failure instead of silently marking maintenance as skipped: it retries automatically, keeps an actionable warning visible, and records the exact failure stage for diagnosis. This means the issue either self-recovers or becomes visible to the user and support team instead of leaving the assistant without search capabilities and no explanation.

## What this changes

- distinguish Cloud MCP readiness failures from intentional `skipped` states
- retry transient session tool-restoration failures and surface persistent task-level notices when recovery does not complete
- validate readiness against the active provider and model, not only generic endpoint health
- preserve structured startup, reload, and reconcile failures in server logs instead of swallowing them
- add regression coverage for a retryable injection failure that recovers on the third attempt

## Why the search tool could disappear silently

Session MCP maintenance previously collapsed any unusable reconcile result into `skipped`, and caught maintenance errors without leaving a user-visible failure or durable diagnostic. The startup and engine-reload paths also discarded thrown synchronization errors. Because the selected provider and model were not included in the session health check, generic endpoint health could look ready while the active model still had no injected Cloud MCP tools.

Together, those behaviors made a registration, authentication, or model-projection failure look like the assistant had silently lost `search_capabilities` after restart.

The captured incident also exposed an important distinction: the live meta-MCP remained connected and returned both `search_capabilities` and `execute_capability`; the observed calls failed farther downstream because of invalid Microsoft provider aliases and a Microsoft Graph 401. This change does not misreport those provider failures as MCP transport loss. It makes actual tool-injection and readiness failures visible, diagnosable, and recoverable.

## User impact

- Transient restoration failures retry after 1 second and 3 seconds.
- The task screen keeps a warning visible while automatic restoration is running and shows a persistent error if it cannot recover.
- The notice links directly to Connect and clears after recovery.
- Focus, online, and periodic maintenance can re-check readiness after the initial retry budget.
- Explicitly disabled, removed, or signed-out Cloud MCP states remain quiet and are not automatically overridden.
- Server logs now identify the workspace, trigger, failure code, and stage for startup/reload synchronization and reconcile failures.

## Security and trust boundaries

- Inspector events, toasts, and server logs do not include access tokens, authorization headers, or connector payloads.
- User-facing diagnostics reuse the existing redacted Cloud MCP health code, stage, message, and recommended action.
- The change does not alter Den credential ownership, tenant boundaries, or OAuth token handling.

## Validation

- `pnpm --filter openwork-server typecheck` — passed
- `pnpm --filter @openwork/app exec bun test --isolate tests/session-mcp-maintenance.test.ts` — 8 passed, 0 failed
- `pnpm --filter openwork-server exec bun test src/cloud-mcp-reconcile.e2e.test.ts src/mcp.engine-sync.e2e.test.ts` — 34 passed, 0 failed

## Remaining verification

The packaged desktop restart journey with a real organization and deliberate network/tool-injection fault has not yet been run. Windows/Linux coverage and visual evidence were also not collected. This PR is therefore opened as a draft until that final real-desktop failure-and-recovery journey is confirmed.

### DIFF
--- a/apps/app/src/react-app/domains/connections/use-session-mcp-maintenance.ts
+++ b/apps/app/src/react-app/domains/connections/use-session-mcp-maintenance.ts
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 
 import {
   mintCloudControlMcpToken,
@@ -6,7 +6,13 @@ import {
   type DenMcpToken,
   type DenSettings,
 } from "../../../app/lib/den";
-import type { OpenworkServerClient } from "../../../app/lib/openwork-server";
+import { recordInspectorEvent } from "../../../app/lib/app-inspector";
+import type {
+  OpenworkCloudMcpFailure,
+  OpenworkCloudMcpHealth,
+  OpenworkCloudMcpProviderModelContext,
+  OpenworkServerClient,
+} from "../../../app/lib/openwork-server";
 import { unwrap } from "../../../app/lib/opencode";
 import type { Client, McpServerEntry, McpStatusMap } from "../../../app/types";
 import { attemptSilentMcpReauth } from "./mcp-silent-reauth";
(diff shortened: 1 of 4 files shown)


reachjalil:
Superseded by #2779, which combines this silent-failure fix with the graceful pre-send Cloud tool readiness gate, focused regression coverage, and validated end-to-end screenshot evidence.

PR #2779, opened 2026-07-15
fix(mcp): gate sends until Cloud search tools are ready

reachjalil:
This is the single combined Cloud MCP fix and supersedes #2773.

This PR prevents a message from starting an agent run while `search_capabilities` or `execute_capability` may be missing from the selected provider/model. It combines the original Cloud MCP silent-failure diagnostics with a graceful pre-send readiness gate: users can keep typing, editing, changing models, and attaching files while OpenWork verifies or restores the tools, but the queued message is not released until the exact Cloud tool projection is proven.

## Why this fixes the silent missing-search failure

The previous session maintenance path could collapse an unusable reconcile result into `skipped`, swallow startup/reload errors, and consider generic endpoint health sufficient. That could leave the assistant without the Cloud search tool after restart while the UI provided no durable explanation.

The first commit makes restoration failures visible and retryable. The second commit closes the send-time race: run-start analytics, telemetry, attachment conversion, commands, and `promptAsync` now remain behind the readiness decision. A failed check therefore creates no agent run and does not clear the draft.

## Readiness contract and engine gap

`connected` transport status is not enough. The gate requires all of the following for the active workspace, provider, and model:

1. The engine reports the exact `openwork-cloud` registration as connected and the overall Cloud health check has no failure.
2. A direct MCP `tools/list` probe returns both `search_capabilities` and `execute_capability`.
3. The provider/model-scoped OpenCode tool listing returns both `openwork-cloud_search_capabilities` and `openwork-cloud_execute_capability`.

The current health endpoint can prove selected-model projection only when OpenCode's provider/model tool-list endpoint enumerates MCP tools. Some engine versions omit MCP tools there and fall back to `/provider`, which proves only that the model supports tool calling in general. This PR explicitly rejects that fallback with `provider_tool_projection_unverified`; generic tool-calling support is never treated as proof that the two Cloud tools were injected.

That is the strongest honest check available without a stronger engine contract. A future OpenCode contract should expose an authoritative provider/model-scoped effective tool manifest on every supported engine version.

## Submission behavior

- Ready tools send immediately.
- Repairable failures keep the exact submitted draft and attachments in place, show **Preparing connected service tools…**, reconcile with bounded retries after 1 second and 3 seconds, then send exactly once after proof succeeds.
- The composer editor, attachment controls, and model picker remain usable during preparation; only the already-queued send action is deduplicated.
- If the user edits while preparation runs, the original snapshot is sent after readiness and the newer composer contents are not cleared.
[...]

PR #2780, opened 2026-07-15
fix(mcp): prevent expired Cloud tokens from silently removing search tools

reachjalil:
> [!WARNING]
> **Review status (2026-07-17): supersede/reslice — do not merge this branch.**
>
> The incident is real, but the broad branch is stale and conflict-heavy. Most of its core protection is now present on `dev` through #2776, #2777, #2778, #2779, #2786, and #2829. The remaining behavior should be rebuilt as focused current-`dev` changes rather than resolving this branch wholesale.

## Independent review findings

- A stale maintenance snapshot can race with a fresh explicit **Disable** or removal and clear that newer intent, re-enabling a connection the user just turned off. Clearing legacy state needs a compare-and-set/revision guard and must only target the raw legacy marker.
- The branch can clear intent after health alone without proving that the repaired Cloud tools are actually usable.
- Cancellation must be rethrown rather than converted into retry/repair failure; otherwise later transport cancellation work will be swallowed here.
- Automatic retired-`openwork-admin` cleanup should only mutate the server-owned runtime source. Static/project/global entries must remain diagnostic-only.
- A failed engine disconnect is currently swallowed while configuration is removed, which can report readiness with a lingering live client.
- Token mint failures are attributed to the wrong diagnostic stage.
- The reported GitHub test failure is an unrelated 5 ms evidence-window flake; the same focused test passed 20/20 locally and does not validate this branch's incident behavior.

## Recommended path

Keep the merged incident protections, then add only the residual legacy-intent compare-and-set repair with a race regression test. Handle retired-admin cleanup separately. This PR remains useful as incident history, not as a merge candidate.

---

Depends on #2779.

This follow-up prevents an expired seven-day OpenWork Cloud MCP token from silently removing `search_capabilities` and `execute_capability`. PR #2779 protects message submission while tools are being prepared; this PR fixes the underlying repair loop so a stale desktop intent marker or the exact Den expiry error cannot permanently suppress token renewal.

## Confirmed incident

Ben desktop retained `openwork.den.mcp.cloudControlUserState = "removed"` while an enabled `openwork-cloud` configuration still existed. When the first-party opaque token expired after seven days, maintenance treated any stored state as a permanent opt-out. The health result then returned `invalid_mcp_token`, but the remint matcher neither recognized that exact code nor inspected `firstFailure.aliases`. OpenCode consequently dropped the MCP and its search and execution tools.

## What changed

- Recognize exact `invalid_mcp_token` failures and all canonical health aliases, while retaining the existing membership, scope, policy, and resource exclusions.
- Compare stored disabled or removed intent with the authoritative workspace MCP configuration before skipping maintenance.
  - Missing or disabled Cloud configuration still preserves explicit user intent.
  - Existing enabled Cloud configuration makes contradictory intent stale; OpenWork clears it, records a credential-free inspector event, and repairs the token.
- Read storage on every maintenance attempt so external changes are observed without a module reload or desktop restart.
- Return structured, actionable failures for configuration reads, token minting, reconciliation, bounded retry exhaustion, and authentication health failures. These flow into the Connect/session status UI added by #2779.
- Remove the retired `openwork-admin` runtime entry before engine registration, disconnect a lingering engine client, and never register it again. Static project or global entries are not silently rewritten; maintenance surfaces a safe removal action instead.
- Keep runtime `/mcp`, strict Cloud health, and direct tool checks authoritative. No `opencode.jsonc` timestamp inference was added.

## Verification

Passed:

- `bun test --isolate tests/cloud-mcp-submit-readiness.test.ts tests/connect-view.test.ts tests/builtin-mcp-visibility.test.ts tests/cloud-mcp-health.test.ts tests/session-mcp-maintenance.test.ts tests/cloud-mcp-user-state.test.ts` — 55 passed
- `bun test src/cloud-mcp-reconcile.e2e.test.ts` — 16 passed

The regressions cover exact and aliased token failures, stale legacy removed state with enabled Cloud config, genuine removal and disablement, live storage changes, actionable mint retry exhaustion, pre-send readiness compatibility, and retired admin cleanup without re-registration.

## Manual verification

Not run — the affected desktop is remote and a real seven-day token-aging cycle was not available locally. Ben incident bundle already confirms that clearing the stale state and restarting allowed a fresh token, an engine-connected `openwork-cloud`, and a successful real `search_capabilities` call.

Reviewer reproduction:

1. Seed an enabled `openwork-cloud` entry with stale legacy `cloudControlUserState = "removed"`.
2. Return `invalid_mcp_token` from strict Cloud health.
3. Trigger session maintenance or focus the app.
4. Confirm intent is cleared, a new token is minted, reconciliation runs once, and both Cloud tools remain available.
5. Remove or disable the Cloud entry explicitly and confirm maintenance does not recreate it.

## Security and risk

- Structured events contain workspace IDs, states, stages, retry flags, and failure codes only; no bearer tokens, Den session tokens, headers, URLs, or credentials are logged.
- The only automatic legacy deletion is the retired server-managed runtime name `openwork-admin`; user-owned static configuration is preserved and surfaced for manual action.
- This PR is intentionally stacked on #2779 and should remain draft until that readiness-gate PR finishes verification.

### DIFF
--- a/apps/app/src/app/lib/app-inspector.ts
+++ b/apps/app/src/app/lib/app-inspector.ts
@@ -120,7 +120,7 @@ export function publishInspectorSlice(
 
 export function recordInspectorEvent(name: string, data?: unknown) {
   installIfNeeded();
-  window.__openwork?.record(name, data);
+  if (typeof window !== "undefined") window.__openwork?.record(name, data);
 }
 
 export function ensureInspectorInstalled() {
--- a/apps/app/src/react-app/domains/connections/cloud-mcp-reconciler.ts
+++ b/apps/app/src/react-app/domains/connections/cloud-mcp-reconciler.ts
@@ -3,6 +3,7 @@ import {
   type DenMcpTokenMintContext,
   resolveCloudMcpResourceUrl,
 } from "../../../app/lib/den";
+import { recordInspectorEvent } from "../../../app/lib/app-inspector";
 import type {
   OpenworkCloudMcpFailure,
   OpenworkCloudMcpHealth,
@@ -82,6 +83,7 @@ type CloudMcpReconcilerInput = {
   force?: boolean;
   refreshMarginMs: number;
   now?: number;
+  configuredPresent?: boolean;
   configuredEnabled?: boolean | null;
 };
 
@@ -189,7 +191,8 @@ export function isCloudMcpAuthTokenFailureCode(code: string | null | undefined):
   ) {
     return false;
   }
-  return normalized === "openwork_cloud_auth_required" ||
+  return normalized === "invalid_mcp_token" ||
+    normalized === "openwork_cloud_auth_required" ||
     normalized === "openwork_cloud_auth_invalid" ||
     normalized === "openwork_cloud_token_expired" ||
     normalized.includes("invalid_token") ||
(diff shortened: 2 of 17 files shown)


reachjalil:
Closing during the 2026-07-27 repository-progress reconciliation. Its own review status says supersede/reslice and notes that most protection already landed. Residual recovery races remain backlog evidence, not a mergeable branch. No unmerged branch is being treated as delivered; any residual gap must be reproduced before new implementation.

PR #2815, opened 2026-07-15
feat(evals): tag local nightly suites (desktop-core, extensions-local, landing)

benjaminshafii:
## What

Part 2 of the nightly-suites series (runner + `nightly-connect`: #2814). Tags the **local-only** flows into three named suites so overnight runs cover whole desktop journeys with one command each:

- **`nightly-desktop-core`** (27 flows) — the solo-user journey in order: boot/smoke → onboarding → chat round-trip with session persistence → session tabs/history/groups → cross-session + in-chat + split-screen find → artifacts (markdown, PDF, tab overflow, narrow header) → built-in browser tabs → composer paste/attachments → model-missing prompt → notifications → Command-K/settings → diagnostics/analytics/voice context.
- **`nightly-extensions-local`** (9 flows) — extensions & MCP without cloud: settings render, marketplace update filters, hidden built-in MCPs, control-pane cleanup, portable export with secret redaction, OAuth silent reauth, cloud MCP force-sync, white-label desktop policies, UI-control tools opt-in.
- **`nightly-landing`** (5 flows) — website: download card, paper shader scope, connect-MCP story, hero prompt analytics, PostHog prod gate (env-gated members skip cleanly without `OPENWORK_EVAL_LANDING_URL`).

Tag-only change (`suite` + `suiteOrder` fields): inert until #2814 merges, so merge order is flexible. Central suite documentation lands with the nightly orchestrator PR.

## Tests run

Validated against the #2814 runner via `OPENWORK_EVAL_FLOWS_DIR`:

- `--list` shows `Suites: nightly-desktop-core, nightly-extensions-local, nightly-landing` ✅
- `--list --suite <each>` prints all 41 flows in the exact journey order above (suiteOrder respected) ✅
- All 170 flow modules still load (no loader validation errors) ✅

No runtime behavior changes; flows themselves are untouched beyond the two metadata fields.

PR #2814, opened 2026-07-15
feat(evals): suite runner support + nightly-connect suite

benjaminshafii:
## What

First step of unifying our ~170 coded eval flows into a few named journeys we can run overnight (analysis in the tracking discussion; suites follow in stacked PRs).

- Adds a **suite** concept to the eval runner: flows opt in with `suite: "<id>"` and optional `suiteOrder` (journey position; falls back to flow id).
- `pnpm evals --suite <id>` runs a suite in journey order; repeatable and deduped; unknown suites error with the available list.
- `pnpm evals --list` shows suite tags plus a suite summary; `--list --suite <id>` prints a suite's journey order.
- `report.md` / `report.json` / `fraimz.html` now carry each flow's suite.
- Replaces the `selectedStackNeedsApp` source-regex sniffing with real flow metadata (selection now happens before `--stack den` boots, so `--list` no longer brings the stack up).

## First suite: `nightly-connect`

Tags the ten OpenWork Connect flows as one journey: `connect-tab-beta` → `connect-cloud-partition` → `connect-delivery-switch` → `connect-aware-extension-gating` → `connect-agent-diagnostics` → `openwork-connect-status` → `docs-openwork-connect` → `keyless-den-connect` → `reliable-openwork-connect` → `connector-health-recovery`. Env-gated members skip cleanly when the nightly environment does not provide their env.

## Tests run

- `node evals/runner/run.mjs --list --suite nightly-connect` — prints the 10 flows in journey order with suite tags and env gates. ✅
- `node evals/runner/run.mjs --list` — all flows plus `Suites: nightly-connect` summary. ✅
- `node evals/runner/run.mjs --suite nope` — errors with `Unknown suite: nope. Available suites: nightly-connect`. ✅
- Dedup: `--list --suite nightly-connect --suite nightly-connect` → 10 (not 20); `--flow voiceover-first-dx --flow voiceover-first-dx` → runs once. ✅
- Full runner loop through the new selection path: `node evals/runner/run.mjs --flow voiceover-first-dx` → PASSED, `fraimz.html` written. ✅
- `node evals/runner/run.mjs --flow connector-health-recovery` — runs (fails on its own DB-backed integration dependency on this machine; unrelated to the selection change). ⚠️

[...]

PR #2816, opened 2026-07-15
feat(evals): den stack starts den-web and exports the full cloud eval env

benjaminshafii:
## What

Part 3 of the nightly-suites series (#2814 runner + connect suite, #2815 local suites). The nightly cloud suites can only run if `--stack den` provides everything the env-gated flows need — today ~25 cloud flows **skip silently** because the harness exports only `OPENWORK_EVAL_DEN_API_URL`/`_TOKEN` while flows also require the den-web origin and MySQL access.

`--stack den` now:

- **Starts den-web** (`pnpm --filter @openwork-ee/den-web dev:local`) on the eval web port, pointed at the local den-api. Idempotent (reuses an already-serving origin), pid-tracked, and stopped by `--stack-down`.
- **Trusts the den-web origin** (localhost + 127.0.0.1 forms) in den-api's `DEN_BETTER_AUTH_TRUSTED_ORIGINS`/`CORS_ORIGINS` so browser sign-in through den-web works.
- **Builds den-api's workspace deps** (`pnpm --filter @openwork-ee/den-api^... build`) before booting. Previously a fresh checkout with an existing schema crashed den-api with `ERR_MODULE_NOT_FOUND` (`@openwork-ee/utils`, `@openwork-ee/den-db`, `@openwork/email` dists missing) because only the schema-push path built anything.
- **Exports** `OPENWORK_EVAL_DEN_WEB_URL`, `OPENWORK_EVAL_DEN_MYSQL_CONTAINER`, `OPENWORK_EVAL_DEN_MYSQL_DATABASE` alongside the API URL + token.

Ports overridable via `OPENWORK_EVAL_DEN_PORT` / `OPENWORK_EVAL_DEN_WEB_URL`.

## Tests run (real stack, this machine)

Fresh worktree, isolated ports (`OPENWORK_EVAL_DEN_PORT=8894`, web `:3106`):

1. `node evals/runner/run.mjs --stack den --flow voiceover-first-dx`
   - MySQL healthy → schema present → **dep build** → den-api + proxy healthy → demo org present → **den-web serving at http://localhost:3106** → env exported → flow PASSED. ✅
2. Auth through the harness-started den-web with the flows' origin:
   `POST http://localhost:3106/api/auth/sign-in/email` (origin `localhost:3106`) → **token minted for alex@acme.test** — proves den-web→den-api proxying and the trusted-origins wiring (a reused den-api without this env still rejects, as before). ✅
3. `--stack-down` → stopped den-proxy, den-api, **den-web**; ports dead afterwards. ✅
4. Repro of the pre-fix failure: without the dep build, den-api crashed with `ERR_MODULE_NOT_FOUND @openwork-ee/utils/dist/observability.js` on this fresh worktree.

Not run here: full env-gated cloud flows against this stack (they need the desktop app too; that's the nightly orchestrator PR's job).

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -93,16 +93,21 @@ A non-zero exit code means at least one flow failed.
 
 ```bash
 pnpm evals --all --stack den     # MySQL + schema + den-api + demo seed +
-                                 # desktop bootstrap + dev app, then runs flows
+                                 # den-web + desktop bootstrap + dev app,
+                                 # then runs flows
 pnpm evals --stack-down          # stop what --stack den started
 ```
 
-`--stack den` is idempotent: each layer (MySQL, schema, den-api, seed, app)
-is skipped when already up. It signs in as the seeded demo owner
-(`alex@acme.test`) and exports `OPENWORK_EVAL_DEN_API_URL` /
-`OPENWORK_EVAL_DEN_TOKEN`, so the env-gated cloud flows run with zero manual
-setup. Requires Docker. The MySQL volume survives `--stack-down`, so
-subsequent runs skip schema push and seeding.
+`--stack den` is idempotent: each layer (MySQL, schema, den-api, seed,
+den-web, app) is skipped when already up. It signs in as the seeded demo
+owner (`alex@acme.test`) and exports `OPENWORK_EVAL_DEN_API_URL` /
+`OPENWORK_EVAL_DEN_TOKEN` / `OPENWORK_EVAL_DEN_WEB_URL` /
+`OPENWORK_EVAL_DEN_MYSQL_CONTAINER` / `OPENWORK_EVAL_DEN_MYSQL_DATABASE`,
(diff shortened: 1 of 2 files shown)


Pallavikumarimdb:
I have been testing the evals automation flow and got the full boot sequence working on Windows > MySQL → schema → den-api → proxy → den-web, all healthy. `DEN_ORG_MODE` fix confirmed present. Trusted-origins/CORS scope looks appropriately local-dev-only.

Needed two small Windows-only patches to get it booting at all:
- `build-assets.mjs`'s `spawnSync` call needs `shell: true` (breaks on Windows' handling of `^` in filter names otherwise)
- `dev:local` in den-web's `package.json` needs a `cross-env`-style fix for its inline env vars

Happy to send these as a quick patch on top if useful. Otherwise the design here is solid and verified end-to-end.

PR #2825, opened 2026-07-15
fix(app): refresh BYO model catalog in active scope

reachjalil:
> [!NOTE]
> **Refreshed-head CI (2026-07-17):** Ubuntu, macOS, and i18n GitHub Actions passed. The remaining Vercel failures are fork-preview authorization requirements, not code/test failures.

## Summary

- keep the invalid-model recovery popup and send guard intact
- refresh provider authentication against the same OpenCode endpoint and workspace cache scope used by the model picker
- prove both same-endpoint cache reuse and cross-endpoint isolation when two OpenCode servers use the same workspace path
- sync the branch with current `dev` at `ac1711fe`

## Confirmed issue found

The defect addressed here is: **the popup can treat a newly connected Bring Your Own model as unavailable**.

Provider authentication refreshed a catalog keyed only by workspace directory, while the session guard and model picker read a catalog keyed by OpenCode base URL plus directory. After saving a Bring Your Own provider credential, connected-provider state could be fresh while the availability check still read the stale pre-connection catalog.

## Relationship to the originally observed popup

We have **not confirmed that this cache defect caused the originally observed behavior**. If the selected model was deliberately removed and no valid replacement was selected, reopening the recovery popup on each application launch is expected: the saved selection remains invalid and OpenWork must continue blocking it.

This PR should therefore be evaluated specifically against the newly connected BYO-provider scenario. It does not attempt to suppress the popup for a genuinely removed model.

## Checks

- `./bin/openwork-hub check desktop byo-model-mismatch` — passed
- `pnpm --filter @openwork/app exec bun test --isolate tests/provider-auth-cache-scope.test.ts` — passed (2 tests, 0 failures)
- substantive GitHub Actions on the prior head were green; current-head Actions are running after the branch update

The added cross-endpoint assertion would fail before this fix: a catalog primed from one OpenCode base URL must not satisfy a refresh against another base URL, even when both use the same workspace directory.

## Manual behavior

Not run — the isolated worktree does not contain a live provider API key. To verify the confirmed defect: add a Bring Your Own API key, select one of that provider’s models, reload the session, and confirm the recovery popup does not reopen. Then disconnect or remove the selected model and confirm the popup still appears and sending remains blocked until a valid replacement is selected.

## Risk

Low and localized to provider catalog cache identity. No credential storage, model identifiers, or unavailable-model guard behavior changed.


reachjalil:
Closing during the 2026-07-27 repository-progress reconciliation. Superseded by merged #3150, which changes the same provider-auth/model-picker path and delivers managed-model refresh, recovery, and import ordering. No unmerged branch is being treated as delivered; any residual gap must be reproduced before new implementation.

PR #2837, opened 2026-07-16
fix(den): restart Daytona workers with sandbox

benjaminshafii:
## Summary

- start Daytona-hosted OpenWork workers from the runtime image ENTRYPOINT so they return after sandbox stop/start
- persist the complete worker runtime contract in sandbox env and retain a marker-gated legacy fallback for older/custom images
- bound startup health probes, preserve useful diagnostics, and add restart-safe tests plus an internal fraimz flow

## Root cause

The provisioner launched `openwork serve` as a one-off async Daytona session. Stopping the sandbox killed that process, and starting it again only ran the image entrypoint; the worker server never returned. On the current `openwork-0.17.30` snapshot, `/health` stayed at HTTP 502 for 67 seconds after stop/start and the internal port refused connections.

## Validation

- `bun test ee/apps/den-api/test/daytona-runtime-env.test.ts` — 6 passed, 0 failed
- `bash -n ee/apps/den-worker-runtime/openwork-daytona-worker-start` — passed
- `node --check evals/flows/daytona-worker-restart-readiness.flow.mjs` — passed
- `DEN_API_LATEST_APP_VERSION=0.17.1 pnpm --filter @openwork-ee/den-api build` — passed
- built and published a fresh amd64 Daytona snapshot from this branch
- real Daytona stop/start: worker `/health` returned HTTP 200 automatically about 6 seconds after start, with no manual relaunch
- `pnpm fraimz --flow daytona-worker-restart-readiness` — PASSED; proof will be posted below

## Rollout note

Hosted Den must point `DAYTONA_SNAPSHOT` at a snapshot built from this change before newly provisioned workers gain restart-safe autostart.

### DIFF
--- a/ee/apps/den-api/src/workers/daytona-runtime-env.ts
+++ b/ee/apps/den-api/src/workers/daytona-runtime-env.ts
@@ -1,0 +1,68 @@
+export const daytonaWorkerStartupScriptPath = "/usr/local/bin/openwork-daytona-worker-start"
+export const daytonaWorkerStartupLogPath = "/tmp/openwork-daytona-worker-start.log"
+export const daytonaWorkerAutostartMarkerPath = "/usr/local/share/openwork/daytona-worker-entrypoint"
+
+export type DaytonaWorkerRuntimeEnv = Record<string, string>
+
+export type DaytonaWorkerRuntimeEnvInput = {
+  workerId: string
+  hostToken: string
+  clientToken: string
+  activityToken: string
+  activityHeartbeatUrl: string
+  workspaceMountPath: string
+  dataMountPath: string
+  runtimeWorkspacePath: string
+  runtimeDataPath: string
+  sidecarDir: string
+  openworkPort: number
+  opencodePort: number
+}
+
+export const daytonaWorkerRuntimeEnvKeys = [
+  "DEN_RUNTIME_PROVIDER",
+  "DEN_WORKER_ID",
+  "DEN_ACTIVITY_HEARTBEAT_ENABLED",
+  "DEN_ACTIVITY_HEARTBEAT_URL",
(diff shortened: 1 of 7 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-16T06-39-46-981Z`

Full frame proof with validated screenshots: `evals/results/2026-07-16T06-39-46-981Z/fraimz.html` (re-run: `pnpm fraimz --flow daytona-worker-restart-readiness`)

### ✅ daytona-worker-restart-readiness — Daytona production worker restarts from container startup
_Internal demo_

1. **The configured Daytona worker is healthy before restart**
   > 🎙 We start with the worker already healthy through its public Daytona health URL. This proves the sandbox is configured and the official runtime has OpenWork serving before we touch the instance.
   - ✅ Public Daytona /health returned 2xx before stop/start
2. **The worker returns automatically after Daytona stop/start**
   > 🎙 We stop and start that same Daytona sandbox, then only poll the public health URL. When health returns without any manual relaunch command, the frame proves the worker came back from the container startup path.
   - ✅ daytona stop exited 0
   - ✅ Daytona reported the sandbox fully stopped
   - ✅ daytona start exited 0
   - ✅ Public Daytona /health returned 2xx after start without a manual openwork relaunch
   - ✅ Script frame 1 narrated: "We start with the worker already healthy through its public Daytona health URL. This pro"
   - ✅ Script frame 2 narrated: "We stop and start that same Daytona sandbox, then only poll the public health URL. When "

benjaminshafii:
Closing this because I scoped the fix to Daytona worker infrastructure instead of using Daytona only to reproduce the production restart issue. I’m restarting the investigation against the actual production startup/readiness path.

PR #2841, opened 2026-07-16
fix(orchestrator): wait through slow startup health

benjaminshafii:
## Summary

- extend shared production startup health waits from 10 seconds to 60 seconds so slow OpenCode and OpenWork cold starts do not terminate the orchestrator
- bound each individual health probe to 2 seconds while preserving explicit short call-site timeouts, OpenCode path fallback, and proxy readiness semantics
- add focused regression tests plus an internal voiceover/fraimz flow for the real production `openwork serve` path

## Root cause

`openwork serve` waited only 10 seconds for OpenCode health. A legitimate cold start beyond that boundary rejected startup before `openwork-server` could spawn, shutting down the full production process.

## Validation

- `pnpm --filter openwork-orchestrator test` — passed (5 tests)
- `pnpm --filter openwork-orchestrator typecheck` — passed
- `pnpm --filter openwork-orchestrator build:bin` — passed
- Daytona Linux bundled build — passed after `pnpm --filter openwork-server build:bin` and `pnpm --filter openwork-orchestrator build:bin:bundled`
- `pnpm fraimz --flow production-server-restart-readiness` — Passed, run `2026-07-16T14-14-45-623Z`

## End-to-end proof

Using the official `openwork-0.17.30` Daytona snapshot as the runtime environment, the current-branch Linux binary stayed alive 11 seconds after a real `opencode serve` delay began, completed the full 12-second delay, and brought the public OpenWork `/health` endpoint to HTTP 200 without image or provisioning changes. The frame-by-frame proof will be posted below.

### DIFF
--- a/apps/orchestrator/package.json
+++ b/apps/orchestrator/package.json
@@ -12,6 +12,7 @@
     "build:bin:bundled": "node scripts/clean-dist.mjs && node ../desktop/scripts/prepare-sidecar.mjs --outdir dist && bun ./script/build.ts --outdir dist --filename openwork && bun scripts/build-bin.ts",
     "build:sidecars": "node scripts/build-sidecars.mjs",
     "typecheck": "tsc -p tsconfig.typecheck.json",
+    "test": "bun test test/startup-health.test.ts",
     "test:files": "pnpm build && node scripts/files-session.mjs",
     "prepublishOnly": "node -e \"console.error('openwork-orchestrator is published via apps/orchestrator/scripts/publish-npm.mjs (meta + platform packages)'); process.exit(1);\""
   },
--- a/apps/orchestrator/src/cli.ts
+++ b/apps/orchestrator/src/cli.ts
@@ -39,6 +39,11 @@ import { createRequire } from "node:module";
 import { once } from "node:events";
 
 import { createOpencodeClient } from "@opencode-ai/sdk/v2/client";
+import {
+  waitForHealthy,
+  waitForHealthyViaProxy,
+  waitForOpencodeHealthy,
+} from "./startup-health.js";
 import type { TuiHandle } from "./tui/app.js";
 
 type ApprovalMode = "manual" | "auto";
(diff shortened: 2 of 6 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-16T14-17-09-639Z`

Full internal frame proof with command outputs and observable assertions (app-less flow; screenshots do not apply): `evals/results/2026-07-16T14-17-09-639Z/fraimz.html` (re-run: `pnpm fraimz --flow production-server-restart-readiness`)

### ✅ production-server-restart-readiness — Production openwork serve waits through slow OpenCode readiness
_Internal demo_

1. **A twelve-second OpenCode cold start no longer exits `openwork serve` at ten seconds**
   > 🎙 In an already running Daytona sandbox, we start the current-branch `openwork serve` binary with a wrapper that delays the real OpenCode server by twelve seconds. The important signal is that OpenWork is still running after the old ten-second boundary instead of exiting early.
   - ✅ The opencode serve delay marker appeared before checking the old boundary
   - ✅ The production openwork process is still alive 11 seconds after the serve delay began
   - ✅ The old-boundary check ran 11 seconds after opencode serve delay began
   - ✅ The OpenCode wrapper delays the real binary for 12 seconds
   - ✅ The OpenCode --version preflight was not delayed by the wrapper
   - ✅ The 12-second delay was applied specifically to opencode serve
   - ✅ The wrapper execs the sandbox's real OpenCode binary after the delay
2. **OpenWork server health reaches 2xx after the delayed OpenCode startup completes**
   > 🎙 Without touching sandbox images or provisioning, we keep polling the shared public `/health` endpoint. Once the delayed OpenCode becomes healthy, OpenWork brings up its own server and the health check turns green automatically.
   - ✅ The public OpenWork /health endpoint returns a 2xx status
   - ✅ The serve-only 12-second delay completed before OpenWork became healthy
   - ✅ Script frame 1 narrated: "In an already running Daytona sandbox, we start the current-branch `openwork serve` bina"
   - ✅ Script frame 2 narrated: "Without touching sandbox images or provisioning, we keep polling the shared public `/hea"

PR #2886, opened 2026-07-17
feat(app): create tasks via internal API + selectable user chat bubbles

benjaminshafii:
## What

Two chat/session fixes (the "Empty message" placeholder removal was **split out to #2913**):

1. **User chat bubbles are selectable.** The shared `ContextMenuTrigger` adds
   `select-none`, and the user bubble is wrapped in one — so drag-select did
   nothing and Select All / Cmd+A silently skipped your own messages
   (assistant text was unaffected). The bubble trigger now opts back into
   `select-text`.
2. **"Create a new task/session" goes through the internal API.** New
   `POST /workspace/:id/sessions` on the OpenWork server + an
   `openwork_session_create` plugin tool (with steering) so the agent creates
   sessions via the backend instead of driving the UI. The sidebar picks up
   API-created sessions live: session-sync listeners now fire for untracked
   `session.updated` **and `session.created`** events, and the route state
   does a debounced background sessions reload for unknown root sessions.

## Validation (fraimz)

Two voiceover-first flows, driven end-to-end on a **Daytona Electron sandbox**
(real desktop build, real OpenAI GPT-5.5 via the eval secrets volume),
re-run on the slimmed branch after the split:

- `chat-user-bubble-selectable` — computed `user-select`, a real CDP mouse
  drag (selection string asserted), and `Select All` including the bubble text.
- `agent-creates-task-via-api` — "Create a new task called Plan the offsite"
  → transcript shows the `openwork_session_create` tool call → the task
  appears in the sidebar with **no reload** → the server sessions API lists it.

Latest combined `fraimz.html` is posted as the most recent proof comment.

> Local note: validated on Daytona rather than local Electron because the work
> session ran over SSH (macOS won't grant WindowServer access to GUI apps
> launched from an SSH context). The sandbox runs the real desktop app.

## Tests

- `pnpm --filter @openwork/app test` — 339 pass, 0 fail (re-run after the split)
- `pnpm --filter openwork-server test` — 477 pass (2 skip), 0 fail
  (includes route tests for `POST /workspace/:id/sessions` and plugin tests
  for `openwork_session_create`; run with a clean `apps/server/dist`)
- `pnpm --filter @openwork/app typecheck` / `pnpm --filter openwork-server typecheck` — clean
- `pnpm fraimz --flow chat-user-bubble-selectable --flow agent-creates-task-via-api --cdp-url <daytona>` — PASSED (latest comment)

PR #2888, opened 2026-07-17
fix(app): sidebar spinner clears task titles; scope Select All to the conversation

benjaminshafii:
## What

Two chat-shell UX fixes:

1. **Sidebar spinner no longer covers the task title.** The session row
   reserves end padding for the absolutely-positioned status spinner, but
   `isSessionStreaming || isSessionActive && "pe-8"` parses as
   `isSessionStreaming || (isSessionActive && "pe-8")` — while **streaming**
   it yields `true`, which `cn()` ignores, so the padding was dropped exactly
   when the spinner is shown and long titles rendered underneath it. Fixed the
   precedence in both row variants (`pe-8` / `pe-12`); titles now truncate
   before the spinner, so nothing needs a masking background.

   | Before (dev) | After (this PR) |
   |---|---|
   | before | after |

   Measured on dev during streaming: `titleRect.right=199` vs `spinnerRect.left=186`
   (13px of title under the spinner), `paddingRight=12`. With the fix:
   no overlap and `paddingRight≥32` (assertions in the fraimz flow).

2. **Select All is scoped to the conversation.** Cmd/Ctrl+A used to sweep the
   entire UI (sidebar labels, tabs, status bar) into the selection. Now, when
   focus is not in an editable element and a transcript is visible, Select All
   selects only the chat messages. Two trigger paths are wired:
   - renderer keydown (capture) — covers web mode;
   - the Electron **Edit ▸ Select All** menu item (which intercepts the
     accelerator before any DOM event): it now calls the renderer's scoped
     handler and falls back to native `webContents.selectAll()` for editable
     focus, the built-in browser panel, devtools, and non-chat surfaces —
     so composer/inputs keep behaving like normal text fields.

   Note: on this base branch user bubbles are still `user-select: none`; once
   #2886 lands, Select All includes your own bubbles too. The two PRs are
   independent and merge in any order.

## Validation (fraimz)

Two voiceover-first flows on a Daytona Electron sandbox (real desktop build,
real GPT-5.5):

- `sidebar-spinner-clear-of-title` — runs a real task with a long
  auto-generated title and measures, mid-stream, that the title box ends
  before the spinner box and the padding reservation is applied; also run
  against dev's sidebar file in the same sandbox to reproduce the failure
  (numbers above).
- `chat-select-all-scoped` — real CDP Ctrl+A: selection includes the
  assistant reply and none of the shell strings ("Add workspace",
  "Ready for new tasks", "Search sessions"); the same scoped call the menu
  executes is asserted directly; with the cursor in the composer, Ctrl+A
  selects only the draft text (anchorNode inside the composer).

Combined `fraimz.html` posted as a comment below.

## Tests

- `pnpm --filter @openwork/app typecheck` — clean
- `pnpm --filter @openwork/app test` — 350 pass, 0 fail
- `node --check apps/desktop/electron/app-menu.mjs` — clean
- `pnpm fraimz --flow sidebar-spinner-clear-of-title --flow chat-select-all-scoped --cdp-url <daytona>` — PASSED (comment below)

Known gap: the OS-level menu accelerator itself can't be triggered via CDP;
the menu's exact renderer call (`window.__openworkChatSelectAll()`) is
asserted instead, and the menu change is a 12-line reviewed diff with native
fallback.

### DIFF
--- a/apps/app/src/components/chat/message-list.tsx
+++ b/apps/app/src/components/chat/message-list.tsx
@@ -38,6 +38,7 @@ import { WebfetchTool } from "@/components/tools/webfetch"
 import { WebsearchTool } from "@/components/tools/websearch"
 import { useMessageList, useSessionErrorMessage } from "@/components/chat/message-list-provider"
 import { ArtifactList } from "@/components/chat/artifact"
+import { useChatSelectAll } from "@/components/chat/use-chat-select-all"
 import { TaskSuggestions } from "@/components/chat/task-suggestions"
 import {
   DescriptiveButtonContent,
@@ -854,6 +855,8 @@ interface MessageListProps {
 }
 
 export function MessageList({ messages, status, retryStatus }: MessageListProps) {
+  useChatSelectAll()
+
   const isStreaming = status === "streaming" || status === "retrying"
   const items = React.useMemo(() => groupMessages(messages, status), [messages, status]);
   const error = useSessionErrorMessage();
@@ -863,7 +866,7 @@ export function MessageList({ messages, status, retryStatus }: MessageListProps)
     : null
 
   return (
-    <div className={cn("flex flex-col gap-2 @container/message-list")}>
+    <div data-chat-transcript className={cn("flex flex-col gap-2 @container/message-list")}>
       {messages.length === 0 && <TaskSuggestions className="mx-auto w-full max-w-3xl shrink-0 px-3 pb-3 md:px-5 md:pb-5 grow" />}
 
       {items.map((item) => {
--- a/apps/app/src/components/chat/use-chat-select-all.ts
+++ b/apps/app/src/components/chat/use-chat-select-all.ts
@@ -1,0 +1,99 @@
+import { useEffect } from "react"
+
+declare global {
+  interface Window {
+    __openworkChatSelectAll?: () => boolean
+  }
+}
+
+let mountedTranscriptCount = 0
+
+function isEditableElement(element: Element | null) {
+  if (!element) return false
+
+  return element instanceof HTMLInputElement
+    || element instanceof HTMLTextAreaElement
+    || Boolean(element.closest('[contenteditable="true"]'))
+}
+
+function isVisible(element: Element) {
+  return element.getBoundingClientRect().width > 0
+}
+
+function findTranscriptRoot(activeElement: Element | null) {
+  const containingRoot = activeElement?.closest("[data-chat-transcript]")
+
+  if (containingRoot) {
+    return containingRoot
+  }
+
+  for (const root of document.querySelectorAll("[data-chat-transcript]")) {
+    if (isVisible(root)) {
+      return root
+    }
+  }
+
+  return null
+}
+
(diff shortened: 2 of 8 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

2 passed · 0 failed · 0 skipped — run `2026-07-17T07-37-53-180Z`

Full frame proof with validated screenshots: `evals/results/2026-07-17T07-37-53-180Z/fraimz.html` (re-run: `pnpm fraimz --flow chat-select-all-scoped --flow sidebar-spinner-clear-of-title`)

### ✅ chat-select-all-scoped — Select All is scoped to chat transcripts and editable composer drafts
_User-facing flow demo_

1. **A fresh chat contains the exact prompt and assistant reply**
   > 🎙 Alex opens a chat that already has a question and an answer in the transcript.
   - 📸 `chat-select-all-scoped-01-select-scope-transcript.png` — 5 validations passed

   

2. **Plain Ctrl+A outside editable fields selects the visible chat transcript and excludes shell chrome**
   > 🎙 Alex presses Select All from the conversation. The highlight covers the chat messages and only the chat messages — the sidebar, tabs, and status bar stay out of the copied text.
   - 📸 `chat-select-all-scoped-02-chat-transcript-selected.png` — 5 validations passed

   

3. **Ctrl+A in the composer selects only the editable draft**
   > 🎙 With the cursor in the composer, Select All behaves like a text field and selects only the draft text.
   - 📸 `chat-select-all-scoped-03-composer-draft-selected.png` — 5 validations passed

   

   - ✅ Script frame 1 narrated: "Alex opens a chat that already has a question and an answer in the transcript."
   - ✅ Script frame 2 narrated: "Alex presses Select All from the conversation. The highlight covers the chat messages an"
   - ✅ Script frame 3 narrated: "With the cursor in the composer, Select All behaves like a text field and selects only t"

### ✅ sidebar-spinner-clear-of-title — Sidebar streaming spinner stays clear of long task titles
_User-facing flow demo_

1. **A fresh task starts running and renders the user's prompt**
   > 🎙 Alex kicks off a task and the sidebar row shows a spinner while the agent works.
   - 📸 `sidebar-spinner-clear-of-title-01-streaming-task-started.png` — 4 validations passed

   

2. **The active sidebar row reserves enough end padding for the streaming spinner**
   > 🎙 Even with a long task name, the title now truncates before the spinner instead of running underneath it — the spinner never covers the letters.
   - 📸 `sidebar-spinner-clear-of-title-02-spinner-clear-of-title.png` — 4 validations passed

   

3. **The same sidebar row removes the spinner after the task finishes**
   > 🎙 When the run finishes, the spinner clears and the row reads normally again.
   - 📸 `sidebar-spinner-clear-of-title-03-spinner-cleared.png` — 4 validations passed

   

   - ✅ Script frame 1 narrated: "Alex kicks off a task and the sidebar row shows a spinner while the agent works."
   - ✅ Script frame 2 narrated: "Even with a long task name, the title now truncates before the spinner instead of runnin"
   - ✅ Script frame 3 narrated: "When the run finishes, the spinner clears and the row reads normally again."

PR #2892, opened 2026-07-17
fix(server): kill child process on spawn failure to prevent orphaned processes

magic-peach:
## Summary

When `createManagedOpencodeServer` spawns a child process, three failure paths reject the promise but never kill the child:

1. **Timeout** (line 103): The setTimeout fires, rejects the promise, but the child process continues running
2. **Error event** (line 125): The child emits an error, `fail()` is called, but the child is not killed
3. **Premature exit** (line 126): The child exits early, `fail()` is called, but no cleanup happens

In all three cases, the child process handle is lost — the caller never receives the `ManagedOpencodeServer` object with its `close()` method. The orphaned `opencode serve` process continues holding a port and consuming memory until manually killed.

**Fix**: Added `child.kill("SIGTERM")` to both the timeout handler and the `fail()` function to ensure the spawned process is cleaned up on all failure paths.

### DIFF
--- a/apps/server/src/managed-opencode.ts
+++ b/apps/server/src/managed-opencode.ts
@@ -101,7 +101,10 @@ export async function createManagedOpencodeServer(options: {
   });
 
   const url = await new Promise<string>((resolve, reject) => {
-    const timeout = setTimeout(() => reject(new Error(`Timeout waiting for OpenCode server after ${options.timeoutMs ?? 15000}ms`)), options.timeoutMs ?? 15000);
+    const timeout = setTimeout(() => {
+      try { child.kill("SIGTERM"); } catch { /* already exited */ }
+      reject(new Error(`Timeout waiting for OpenCode server after ${options.timeoutMs ?? 15000}ms`));
+    }, options.timeoutMs ?? 15000);
     let output = "";
     const done = (value: string) => {
       clearTimeout(timeout);
@@ -109,6 +112,7 @@ export async function createManagedOpencodeServer(options: {
     };
     const fail = (error: Error) => {
       clearTimeout(timeout);
+      try { child.kill("SIGTERM"); } catch { /* already exited */ }
       reject(error);
     };
     child.stdout?.on("data", (chunk) => {


Iflaqbhat:
@magic-peach Verified the startup failure cleanup path locally on this PR branch.

I tested two failure cases with a fake OpenCode binary:

- child never prints the ready URL -> `createManagedOpencodeServer` times out and the fake child is no longer running
- child prints a malformed ready line -> startup rejects and the fake child is no longer running

Also ran:

`pnpm --filter openwork-server typecheck`
`git diff --check`

Both passed.

Non-blocking thought: there does not seem to be a focused regression test for `managed-opencode.ts` yet, so adding one for timeout/fail cleanup would make this harder to regress. But the implementation looks correct to me.

Pablosinyores (commented):
The leak is real and worth fixing. All three paths reject without touching the child, and since the caller never receives the `ManagedOpencodeServer` object, its `close()` is unreachable — so nothing can ever reap that process.

One thing I would change before merge: the new kills stop at `SIGTERM` with no escalation, and the process most likely to leak is exactly the one that will not honour it.

`close()` in the same file already handles this properly:

```typescript
if (!child.killed) child.kill("SIGTERM");
await Promise.race([exited, timeout /* 1000ms */]);
if (child.exitCode === null) {
  child.kill("SIGKILL");
  await Promise.race([exited, /* 500ms */]);
}
```

The timeout path is the strongest case. Reaching it means the child has been unresponsive for 15s — it never printed its listening line — so a single `SIGTERM` to a wedged process is quite likely to leave the same orphan the PR is trying to prevent. Reusing the `SIGTERM` then `SIGKILL` sequence (or factoring it into a small helper both call) would close it for good.

Two smaller notes:

`fail` is also wired to `child.once("exit", ...)`, where the child is already gone, and to the `Failed to parse OpenCode server URL` branch, where it is alive and healthy but being abandoned. Killing is right in both cases, just worth being aware the same helper covers three quite different states. The `try/catch` handles the already-exited case fine.

Since `spawn` is not `detached`, the child gets no process group of its own, so if the OpenCode server spawns children of its own they will not be signalled by these kills. Probably out of scope here, but it is the remaining path to an orphan after this lands.

The `exited` promise defined just above is already the clean way to await the result, so the helper has everything it needs to do the escalation without much restructuring.

Issue #1918, opened 2026-05-23
SPA cannot reply to permission requests: 'Only owner tokens can reply' (client token vs host token gate)


MariusWilsch:
## Summary

The SPA cannot respond to permission requests because it authenticates with the client token (`OPENWORK_TOKEN`), but permission replies are gated to the host token (`OPENWORK_HOST_TOKEN`). Clicking "Deny" / "Allow once" / "Allow for session" returns `403 forbidden — Only owner tokens can reply to permission requests` as toasts, and the underlying tool call stays in `running` limbo.

This is reproducible with the default `external_directory` permission category (which fires by default when opencode native file tools touch paths outside `authorizedRoots` / `workspace.path`).

## Reproduction

1. Vanilla `packaging/docker/docker-compose.dev.yml` on v0.13.12 — auto-generated tokens (or `.env`-pinned).
2. Workspace `opencode.json` — no explicit `permission` block.
3. In the SPA, prompt the agent to read a file outside `/workspace`, e.g. `Run bash 'cat /etc/passwd | head -3'` (or any opencode native file op outside workspace).
4. The "Access an external folder?" / "Run a shell command?" dialog appears.
5. Click any of the three buttons.

### Result

- Toast: `{"code":"forbidden","message":"Only owner tokens can reply to permission requests"}` (×2)
- Tool call stays `status: "running"`
- Dialog auto-dismisses ~10s later (per [#1916](https://github.com/different-ai/openwork/issues/1916))
- Session stuck "Responding"

## Expected

Either the SPA should authenticate the permission-reply endpoint with the host token (so user clicks work), OR the gate should accept client-token replies for SPA-originated UIs. As shipped, the SPA is **fundamentally unable** to clear any permission request in dev-stack default config.

## Workaround

None known short of: (a) patching the SPA to send host token on permission replies, (b) shipping host token to the SPA at boot, or (c) avoiding any operation that triggers a permission gate (which includes any opencode native file op outside `workspace.path`, fired silently by default).

## Environment

OpenWork v0.13.12 (`ef5ae9cc`) · opencode 1.14.38 · vanilla `packaging/docker/docker-compose.dev.yml` · related issues: [#1916](https://github.com/different-ai/openwork/issues/1916) (dialog auto-dismiss), [#1917](https://github.com/different-ai/openwork/issues/1917) (auth.json wipe on restart)

🤖

PR #2893, opened 2026-07-17
fix(server): fix unreachable permission check in OpenCode proxy

magic-peach:
## Summary

In `assertOpencodeProxyAllowed` (`apps/server/src/server.ts:772`), the second `if` block was unreachable dead code:

```typescript
// First block - throws for ALL non-GET/HEAD viewer requests
if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
    throw new ApiError(403, "forbidden", "Viewer tokens are read-only");
}

// Second block - UNREACHABLE (same condition as first block)
if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
    const normalized = normalizeOpencodeProxyPath(proxyPath);
    if (/\/permission\/[^/]+\/reply$/.test(normalized)) {
        throw new ApiError(403, "forbidden", "Viewer tokens cannot reply to permission requests");
    }
}
```

The first block throws for the **exact same condition** as the second block checks. This means:
- The permission reply check is dead code
- All non-GET/HEAD viewer requests are blocked, including permission replies
- Collaborators cannot approve/deny OpenCode permission requests via the proxy

The extensive comment (lines 780-786) explains the original intent was to allow collaborators to reply to permission requests while blocking other viewer writes.

**Fix**: Refactored to check the permission reply path first and `return` early for collaborators, then throw for other viewer write operations.

### DIFF
--- a/apps/server/src/server.ts
+++ b/apps/server/src/server.ts
@@ -774,21 +774,15 @@ export function assertOpencodeProxyAllowed(actor: Actor, method: string, proxyPa
   const scope = actor.scope ?? "viewer";
 
   if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
-    throw new ApiError(403, "forbidden", "Viewer tokens are read-only");
-  }
-
-  // Prevent viewers from self-approving OpenCode permission requests via the
-  // proxy. OpenCode uses /permission/:requestId/reply (and historically also
-  // a session-scoped variant). Collaborators must be allowed: the SPA's only
-  // credential is the collaborator-scoped client token (OPENWORK_TOKEN), so
-  // an owner-only gate made every interactive permission dialog un-answerable
-  // (403 "Only owner tokens can reply") and left tool calls stuck in
-  // "running" forever (#1918).
-  if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
+    // Collaborators must be allowed to reply to permission requests: the SPA's
+    // only credential is the collaborator-scoped client token (OPENWORK_TOKEN),
+    // so an owner-only gate makes every interactive permission dialog
+    // un-answerable and leaves tool calls stuck in "running" forever (#1918).
     const normalized = normalizeOpencodeProxyPath(proxyPath);
     if (/\/permission\/[^/]+\/reply$/.test(normalized)) {
-      throw new ApiError(403, "forbidden", "Viewer tokens cannot reply to permission requests");
+      return;
     }
+    throw new ApiError(403, "forbidden", "Viewer tokens are read-only");
   }
 }
 


Iflaqbhat:
I think this changes the permission boundary in the wrong direction.

The new `return` is inside `scope === "viewer"`, so viewer tokens can now POST to `/permission/:requestId/reply`. Collaborator tokens were already allowed by this function because the write block only applies to viewers.

I verified this on the PR branch with:

`pnpm --filter openwork-server exec bun test src/opencode-proxy-gate.test.ts`

The existing proxy gate tests fail:

- `viewers cannot send any mutating request`
- `missing scope defaults to viewer (read-only)`

So the dead-code cleanup makes sense, but I think the fix should keep viewers read-only and only remove/restructure the unreachable duplicate check.

Pablosinyores (commented):
The dead-code observation is correct: the second `if` has the identical condition to the first, so it can never run. That part is a real cleanup.

But the resolution changes behaviour in the permissive direction, and I think it opens a privilege escalation rather than closing dead code.

Before, a viewer sending any non-GET/HEAD request hit the first `throw` and was refused, permission-reply included. After this change the branch becomes:

```typescript
if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
  const normalized = normalizeOpencodeProxyPath(proxyPath);
  if (/\/permission\/[^/]+\/reply$/.test(normalized)) {
    return;                      // now allowed
  }
  throw new ApiError(403, "forbidden", "Viewer tokens are read-only");
}
```

So a viewer token can now `POST /opencode/permission/:requestId/reply`. That endpoint is how OpenCode tool-permission prompts get answered, so a read-only credential can approve tool calls that write files and run commands. The old dead block was dead in the safe direction, since both branches threw.

The comment is what I think caused this. "Collaborators must be allowed" is a historical note explaining why the gate keys on `viewer` instead of `owner` — it is not saying collaborators reach this branch. They do not:

- `server.ts:1244` declares the scopes as `["owner", "collaborator", "viewer"]`, three distinct values
- the branch only runs when `scope === "viewer"`, so a collaborator never enters it and is already allowed through untouched
- the #1918 regression was an owner-only gate, which was removed by narrowing to `viewer`; that fix is already in place

Worth noting the proxy mount at `server.ts:910` calls `requireClient`, not `requireClientScope(ctx, "collaborator")`, so `assertOpencodeProxyAllowed` is the only thing standing between a viewer token and this route. There is no second gate behind it.

If the goal is just removing the unreachable code, deleting the second block outright preserves current behaviour:

```typescript
if (scope === "viewer" && m !== "GET" && m !== "HEAD") {
  throw new ApiError(403, "forbidden", "Viewer tokens are read-only");
}
```

The specific `"Viewer tokens cannot reply to permission requests"` message is worth keeping if you want the clearer error, by testing the path first and then falling through to the generic throw — same refusals, better message.

Happy to be corrected if viewers are in fact meant to answer permission prompts, but the existing message and the fact that both old branches threw both read as deliberate. Either way this seems worth an explicit decision before merge, since the diff reads as a no-op cleanup.

PR #2898, opened 2026-07-17
fix(mcp): keep OAuth failures visible with diagnostics

reachjalil:
## Summary

- keep MCP OAuth popups open when connection startup fails across both admin-managed and member account flows
- polish the loading state with an intentional OpenWork Connect card, secure-connection progress treatment, and reduced-motion support
- replace the raw failure page with a clearer error layout that explains the window will remain open
- add a collapsed **Technical details** section with the exact redirect URI, HTTP status, error code, client metadata URL, diagnostic reference and phase fields when available, and a sanitized response payload
- add focused popup coverage plus a deterministic real-browser regression flow and voiceover

## Root cause

Both Den Web OAuth launch paths explicitly closed the authorization window in their error handlers. The Den API already returned useful troubleshooting context, including the public `callbackUrl` for `mcp_oauth_configuration_required`, but the client reduced the response to a message and discarded the remaining fields.

## Validation

- `pnpm --filter @openwork-ee/den-web typecheck` — passed
- `pnpm --filter @openwork-ee/den-web exec bun test tests/mcp-authorization-url.test.ts` — 10 passed
- `pnpm --filter @openwork-ee/den-web test` — 57 passed
- `pnpm fraimz --flow mcp-oauth-popup-error-details --cdp-url http://127.0.0.1:33329` against isolated Den Web, Den API, MySQL, and a DCR-less OAuth/MCP provider — passed; the real popup stayed open, expanded Technical details, and showed HTTP 409, `mcp_oauth_configuration_required`, the exact callback URI, client metadata URI, and safe response payload

Additional check from the initial implementation: `bun test tests/mcp-authorization-url.test.ts tests/mcp-oauth-bootstrap.test.ts tests/mcp-oauth-callback-migration.test.ts` produced 17 passes and 1 unrelated failure. The failing callback-migration assertion expects a `canConnectOAuth` source string that is already absent at this branch's `upstream/dev` base; this PR does not modify that row predicate.

## Security and risk

- only allowlisted fields from known Den API error envelopes are copied into the debug payload; arbitrary response data is not exposed
- API messages, URLs, diagnostic values, and JSON are HTML-escaped before being written into the popup
- no credentials, tokens, secrets, or tenant-private request data are added to the UI
- successful already-connected paths still close the temporary popup
- technical details are collapsed by default, keeping the primary error approachable for non-technical users
- if browser isolation prevents rewriting the popup document, it remains open so the browser/provider error is still inspectable

## Manual behavior

The automated browser journey was observed end to end. The loading document is covered by the focused rendering test; it transitions too quickly in the live DCR-less journey to capture as a stable frame. No separate user-confirmed manual run was performed.

### DIFF
--- a/ee/apps/den-web/app/(den)/dashboard/_components/mcp-authorization-url.ts
+++ b/ee/apps/den-web/app/(den)/dashboard/_components/mcp-authorization-url.ts
@@ -31,13 +31,35 @@ export function safeMcpAuthorizationUrl(rawUrl: string): string {
   return url.toString()
 }
 
+export type McpAuthorizationDebugDetails = {
+  httpStatus: number
+  errorCode?: string
+  redirectUri?: string
+  clientMetadataUrl?: string
+  diagnosticReference?: string
+  phase?: string
+  category?: string
+  highestPassed?: string
+  retryable?: boolean
+  actionOwner?: string
+  operatorAction?: string
+  providerStatus?: number
+  providerRequestId?: string
+  providerCode?: string
+  responseJson: string
+}
+
 const authorizationDocumentStyles = `
       :root { color-scheme: light; font-family: Inter, ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif; }
       * { box-sizing: border-box; }
-      body { margin: 0; min-height: 100vh; display: grid; place-items: center; padding: 32px; color: #182033; background: #f7f7f4; }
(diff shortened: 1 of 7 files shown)


reachjalil:
## OAuth popup screenshots

### Loading state

OpenWork OAuth popup loading state

### Error state

OpenWork OAuth popup error state with collapsed technical details

### Expanded diagnostics

OpenWork OAuth popup expanded diagnostics with redirect URI and safe response payload

The loading and collapsed-error captures use the popup’s 600×760 viewport. The expanded capture is taller so the HTTP status, error code, redirect URI, client metadata URL, and sanitized response payload remain visible.

PR #2899, opened 2026-07-18
fix(mcp): make connected-tool schema checks advisory

reachjalil:
## Summary

This PR adds **optional schema guidance** for connected MCP tools without turning OpenWork's local validation into a new execution gate.

Once the existing authentication, access, connection, and tool-existence checks pass, OpenWork **always attempts the downstream provider call**. The provider remains the source of truth: if it accepts a call that does not match its advertised schema, OpenWork returns that success. If it rejects the call, OpenWork returns the real provider error together with guidance that can help the agent correct and retry.

## How this avoids additional failures

The provider call is started **before** advisory validation. Local schema checking cannot reject, delay, or replace that call.

| Situation | Behavior |
| --- | --- |
| No schema digest supplied | Call proceeds normally |
| Provider schema changed after discovery | Call proceeds; optional drift guidance is attached |
| Arguments do not match the advertised schema | Call proceeds; optional issue guidance is attached |
| Provider advertises a malformed or unsupported schema | Call proceeds; OpenWork explains that local guidance was unavailable |
| Provider accepts the call | Provider success is returned; guidance does not trigger a retry |
| Provider rejects the call | Real provider failure is returned with guidance for a possible correction |
| Authentication, access, connection, or tool lookup fails | Existing hard failure remains unchanged |

`schemaDigest` is optional, and `schemaGuidance` is only present when OpenWork has something useful to report. This allows imperfect or unusually implemented MCP servers to keep working.

## Evidence that this does not introduce regressions

In the matched benchmark, there were **32 cases where only the PR succeeded and zero cases where only the base succeeded**. The PR therefore introduced no new failed matched outcome in this test matrix.

- Exact successful tasks: **13/48 without the PR → 45/48 with it**.
- Crowded 229-tool context: **2/9 → 9/9**.
- Deliberately incomplete first-call recovery: **0/9 → 9/9**.
- The nine invalid PR-side provider calls were intentional—one in each recovery test—and all nine corrected successfully. The other 39 PR sessions produced **zero unintended invalid provider calls**.
- Average completion time: **46.4 seconds → 21.1 seconds**.

The comparison used **96 fresh OpenWork sessions / 48 matched A/B pairs** with real GPT-5.4-mini, GPT-5.4, and GPT-5.5. Prompts, models, OpenWork agent configuration, OpenCode `1.17.11`, Den connection, provider state, expected answers, timeouts, grading, and tool context were identical. Only the OpenWork commit changed: base `b22399a` versus PR `8052abb`.

## Verification

- Exact-product benchmark: **15/15 integrity checks passed**; all 96 sessions completed and were independently regraded from provider events.
- Den API focused tests: **111 passed, 0 failed**.
- Enterprise MCP client tests: **46 passed, 0 failed**.
- Den API production build: passed.

The focused tests specifically cover provider success despite a local mismatch, provider rejection with guidance, schema drift, unavailable schema validation, authorization boundaries, and timeouts.

## Remaining limits

Keep this PR in draft. The PR still missed three of 48 benchmark tasks. Unusual JSON Schema dialects and very large schemas can make the optional guidance incomplete or expensive, but they no longer prevent a valid provider call. Electron UI/OAuth onboarding and remote-network provider behavior were not part of this benchmark.

### DIFF
--- a/ee/apps/den-api/src/capability-sources/external-mcp-diagnostics.ts
+++ b/ee/apps/den-api/src/capability-sources/external-mcp-diagnostics.ts
@@ -356,6 +356,12 @@ function providerToolContentEvidence(result: unknown): ProviderToolContentEviden
   }
 }
 
+function isProviderInputValidationExcerpt(value: string | undefined): boolean {
+  if (!value) return false
+  return /^Input validation error:\s*Invalid arguments for tool\b/i.test(value)
+    || /^Invalid (?:tool )?(?:arguments|params)\b/i.test(value)
+}
+
 function errorName(value: unknown): string {
   const name = value instanceof Error ? value.name : stringProperty(value, "name")
   return name && SAFE_ERROR_NAMES.has(name) ? name : "Error"
@@ -868,6 +874,16 @@ function classifyError(error: unknown, fallbackPhase: ExternalMcpDiagnosticPhase
       operatorAction: "Check provider latency and retry after the current MCP request can complete within its bounded timeout.",
     }
   }
+  if (numericErrorCode(error) === -32602 && fallbackPhase === "MCP_TOOL_EXECUTION") {
+    return {
+      phase: "MCP_TOOL_EXECUTION",
+      category: "mcp_tool_input_invalid",
+      code: "MCP_INVALID_PARAMS",
+      retryable: false,
+      actionOwner: "openwork",
+      operatorAction: "Correct the tool arguments using the latest advertised input schema; do not retry the same arguments unchanged.",
+    }
+  }
 
   const name = errorName(error)
   if (name === "InvalidClientError" || name === "UnauthorizedClientError" || name === "InvalidClientMetadataError") {
@@ -1136,10 +1152,24 @@ export class ExternalMcpDiagnosticTracker {
     const requestId = safeProviderRequestId(structuredContent?.requestId) ?? evidence.providerRequestId
     if (requestId) this.providerRequestId = requestId
 
+    const providerInvalidArguments = [
+      "invalid_arguments",
+      "invalid_params",
+      "validation_error",
+    ].includes(providerCategory ?? "") || isProviderInputValidationExcerpt(evidence.excerpt)
     const providerPolicyDenied = structuredProviderStatus === 403
       ? providerCategory === "provider_policy"
       : evidence.providerStatus === 403
-    const classificationBase: Classification = providerPolicyDenied
+    const classificationBase: Classification = providerInvalidArguments
+      ? {
+          phase: "MCP_TOOL_EXECUTION",
+          category: "mcp_tool_input_invalid",
+          code: "MCP_PROVIDER_INVALID_PARAMS",
+          retryable: false,
+          actionOwner: "openwork",
+          operatorAction: "Correct the tool arguments using the latest advertised input schema; do not retry the same arguments unchanged.",
+        }
+      : providerPolicyDenied
       ? {
           phase: "PROVIDER_AUTHORIZATION",
           category: "provider_policy_denied",
--- a/ee/apps/den-api/src/mcp/agent.ts
+++ b/ee/apps/den-api/src/mcp/agent.ts
@@ -109,8 +109,9 @@ export const AGENT_MCP_INSTRUCTIONS = [
   "After importing, retrieve the resolved marketplace detail and report each plugin's cloudReadiness. An import or plugin binding is not proof that an MCP connection is usable. Relay needs_admin_setup or needs_signin as the next human action instead of claiming the connection is ready.",
   "Do not invent OAuth-client, credential, or local-extension setup. Organization connections are managed in the OpenWork Cloud dashboard / Settings > Connect. When a returned connection or marketplace readiness state requires administrator setup or member sign-in, relay that exact action.",
   "A successful search_capabilities call proves this OpenWork Cloud MCP connection is authorized. Never tell the user to reconnect OpenWork Cloud because a downstream connector failed.",
-  "External MCP matches include argumentsSchema, schemaDigest, and invocation.argumentsField. Put an object matching argumentsSchema in execute_capability.body and copy schemaDigest into execute_capability.schemaDigest.",
-  "If execute_capability returns invalid_capability_arguments, correct the listed issues and retry once with changed arguments; never retry the same arguments unchanged. If it returns capability_schema_changed or unknown_capability, call search_capabilities again before retrying.",
+  "External MCP matches include the provider-advertised argumentsSchema, schemaDigest, and invocation.argumentsField. Put an object matching argumentsSchema in execute_capability.body and copy schemaDigest into execute_capability.schemaDigest.",
+  "OpenWork always attempts the downstream provider call when local schema checks find a mismatch. schemaGuidance is advisory and appears alongside the provider result: if the provider succeeded, accept that result and do not retry solely because of the warning; if it failed, use the warning to correct the arguments or search again.",
+  "If the provider returns invalid_capability_arguments, correct the listed issues and retry once with changed arguments; never retry the same arguments unchanged. If it returns unknown_capability, call search_capabilities again before retrying.",
   "When a match has kind connection_status, name connectionStatus.connectionName and relay connectionStatus.action exactly. Distinguish the member's Your Connections page, the organization Connections dashboard, and the provider's own admin console.",
   "Connection probes are live. After the requested human fixes that connector, search again in the same task; otherwise do not retry unchanged or improvise workarounds through other tools.",
 ].join("\n")
@@ -143,6 +144,7 @@ export function externalCapabilityErrorToolResult(
       ...(result.schemaDigest ? { schemaDigest: result.schemaDigest } : {}),
       ...(result.sameArgumentsRetryable === false ? { sameArgumentsRetryable: false } : {}),
       ...(result.retry ? { retry: result.retry } : {}),
+      ...(result.schemaGuidance ? { schemaGuidance: result.schemaGuidance } : {}),
     })),
   }
 }
@@ -180,6 +182,19 @@ function externalToolContent(result: unknown): { type: "text"; text: string }[] 
     return result.content
   }
   return textContent(JSON.stringify(result))
+}
+
+export function externalCapabilitySuccessToolResult(
+  result: Extract<ExternalCapabilityExecuteResult, { ok: true }>,
+): ExecuteCapabilityToolResult {
+  const content = externalToolContent(result.result)
+  if (!result.schemaGuidance) return { content }
+  return {
+    content: [
+      ...content,
+      ...textContent(JSON.stringify({ schemaGuidance: result.schemaGuidance })),
+    ],
+  }
 }
 
 function capabilityTimeoutResult(capability: string): ExecuteCapabilityToolResult {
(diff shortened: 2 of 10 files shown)


reachjalil:
## Extended real-GPT benchmark + code review update

Benchmarked PR head `3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a` against base `b22399a19259fca6f5410704f9bf30b1d394b089`.

### Verdict

The proposed discovery contract is **materially more reliable** for complex argument construction under large tool context in this controlled benchmark. It is **not merge-ready yet**: direct testing of the PR validator found correctness, isolation, retained-memory, and no-argument regressions.

### Real-GPT benchmark

I ran **156 new OpenCode sessions** with real `openai/gpt-5.4-mini`, `openai/gpt-5.4`, and `openai/gpt-5.5` inference. The qualified A/B matrix below contains 138 sessions:

- 90 construction sessions: 2, 50, and 200 exposed tools; five scenarios per model and arm.
- 18 forced-correction sessions: the prompt required one deliberately malformed call before correction.
- 30 forced-drift sessions: the lab guaranteed `search V1 → execute stale digest → capability_schema_changed → search V2 → execute`.

An additional 18 exploratory drift sessions are retained but excluded from the forced-drift claim: three mini-model runs searched twice before executing, so they did not hit the stale-digest response. Strict grading required one provider success, full JSON Schema 2020-12 validity, every requested value exactly correct, exit 0, no timeout/signal, unique sessions, and consistent raw/server metrics.

| Qualified scenario | PR contract | Baseline contract | Paired exact McNemar |
| --- | ---: | ---: | ---: |
| Construction, all tool contexts | **45/45 exact** | **0/45** | `p=5.68e-14` |
| Construction, 200 tools | **15/15 exact** | **0/15** | — |
| Forced invalid-argument correction | **9/9 exact** | **0/9** | `p=0.003906` |
| Forced stale-schema recovery | **14/15 exact** | **0/15** | `p=0.000122` |

Important detail: forced drift was schema-valid 15/15, but only semantically exact 14/15. GPT-5.4-mini used the ordinary change ticket in the new `changeControl.ticket` field once, so I counted that as failure. Forced drift by model was mini 4/5, GPT-5.4 5/5, GPT-5.5 5/5. Construction was 15/15 for every model.

Confidence remains bounded by sample size: the 95% Wilson interval is 92.1–100% for 45/45 construction, 79.6–100% for 15/15 at 200 tools, and 70.2–98.8% for 14/15 forced drift. This is a +100 percentage-point construction/correction improvement and +93.3 points for forced drift in this lab—not a production reliability guarantee.

Downstream boundary counts:

- Construction: baseline forwarded 108 malformed invocations; PR forwarded 45 valid and 0 malformed.
- Forced correction: baseline forwarded 35 malformed invocations; PR blocked all 9 deliberate malformed calls locally, then forwarded 9 valid and 0 malformed.
- Forced drift: baseline forwarded 52 malformed invocations; PR emitted 15 schema-change responses, then forwarded 15 valid and 0 malformed.
- The PR contract copied the digest on all 93 measured execution inputs (45 construction + 18 correction + 30 drift); none were omitted.

At 200 tools, `tools/list` serialized to **220,582 bytes**. Mean wall time was 34.8s to PR success versus 70.2s to baseline terminal failure; mean cumulative processed input was 111.9k versus 232.5k tokens. These are different terminal outcomes, so this is not a pure speedup. “Cumulative processed input” means input + cache-read tokens summed across steps, not one context-window size.

Methodology caveat: this is **contract-level real-GPT evidence**, not two exact-commit Den deployments. One OpenCode binary called a local MCP lab that emulated the baseline contract (no argument schema in search) and PR contract (schema + digest + local validation). Search returned one exact match; repetitive unrelated decoys primarily stress context length, not realistic discovery ambiguity.

### Focused checks

- 93 focused Den API tests passed.
- 52 enterprise MCP client tests passed.
- Both benchmark result sets passed integrity verification: complete raw files, unique run/session IDs, expected cell counts, exit 0, no timeouts, no metric mismatches.
- Core GitHub Actions checks are green at the reviewed head.

### Review blockers reproduced against the actual PR code

1. **[P1] Wrong JSON Schema dialect.** [`AjvJsonSchemaValidator` is instantiated once with its default behavior](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/external-mcp-tool-arguments.ts#L16), then used for every live schema. [MCP 2025-11-25 requires JSON Schema 2020-12 by default and declared-dialect handling](https://modelcontextprotocol.io/specification/2025-11-25/basic#schema). Actual repro: a missing `dependentRequired` dependency was accepted, while a valid `prefixItems` tuple was rejected. This can both leak invalid calls and block valid tools. Use a 2020-12 validator by default, honor explicit `$schema`, and add dialect conformance tests.

2. **[P1] Process-global schema cache causes `$id` poisoning and unbounded retained memory.** The [module-global provider](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/external-mcp-tool-arguments.ts#L16) is reused at [every validation](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/external-mcp-tool-arguments.ts#L61-L63). After priming `$id=https://provider.example/schemas/tool-input` with a schema requiring `query`, a different schema with the same `$id` requiring `incidentId` incorrectly accepted `{query:"outage"}` and rejected `{incidentId:"INC-1"}`. Because this cache is process-global, different connections/organizations can interfere. In a fresh `bun --smol` process with forced GC and a fresh tiny schema per call, retained heap grew **0.7 → 103.5 MiB** and RSS **38.5 → 427.0 MiB** after 10,000 validations. Isolate/bound compilation and prevent `$id` reuse across trust domains.

3. **[P1] No-argument external tools regress.** [`body` is optional](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/agent.ts#L383-L388), and execution now passes [`normalizeToolBody(body)` directly](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/agent.ts#L411-L417). Omitted body remains `undefined`, so an empty-object no-argument schema fails the new root-object guard; explicit `{}` succeeds. Base normalized non-record external args to `{}`. Preserve that behavior for omitted bodies.

4. **[P2] Digest omission bypasses drift detection.** [`schemaDigest` remains optional](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/agent.ts#L383-L388), and execution compares only [when a digest is present](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/external-capabilities.ts#L932-L946). A client can omit it and bypass the stated stale-schema guarantee. Require it for external names or return `searchRequired: true` when missing.

5. **[P2] Search can inject multi-megabyte schema output into model context.** Every match returns the [entire live `inputSchema`](https://github.com/different-ai/openwork/blob/3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a/ee/apps/den-api/src/mcp/external-capabilities.ts#L689-L700); search defaults to 5 and allows 20 matches, while the catalog accepts 512 KiB per schema. Boundary serialization measured about **2.502 MiB** for five maximum schemas and **10.006 MiB** for twenty. Add a response byte/token budget or a narrow follow-up schema retrieval path.

Also, validation currently collapses Ajv details into one issue at `path: "/"`, so the PR description’s “machine-readable issues” claim is only partially true.

I checked the apparent 45s + 150s versus 180s timeout concern separately. The active enterprise path uses cancellable per-operation deadlines within the facade budget, so I am **not** reporting that as a blocker.

reachjalil:
## Exact-product, apples-to-apples OpenWork A/B

I reran the benchmark as a controlled comparison of the exact base and PR commits:

- **Base:** `b22399a19259fca6f5410704f9bf30b1d394b089`
- **PR:** `3dfc6cb66e1d07115a24c55bd4fd6a2daf8efe5a`

Both arms used the same desktop-equivalent OpenWork embedded server, canonical OpenWork prompt/plugins, authenticated Den `/mcp/agent` and Connect reconciliation, and the same OpenCode `1.17.11` binary (`sha256:472624fbe35fcf2c1b2c878f2ebb0312c80d1501304094f33f6e20c9f61161b7`). Real `gpt-5.4-mini`, `gpt-5.4`, and `gpt-5.5` inference was used.

The visual Electron renderer and interactive OAuth onboarding were omitted, but this uses the same embedded-server seam launched by the desktop app. OpenWork, Den, managed OpenCode, prompts, plugins, sessions, retries, MCP transport, and GPT inference were real. Only the target business MCP was deterministic and local so that correctness could be graded reproducibly.

The run contained **96 fresh OpenWork sessions: 48 matched A/B pairs**. Each pair used the same model, case, prompt, provider schema, and tool context; only the OpenWork commit changed. Arm order alternated.

| Result | Base | PR |
| --- | ---: | ---: |
| Exact successful result | 13/48 (27.1%; 95% CI 16.6–41.0%) | **44/48 (91.7%; 95% CI 80.4–96.7%)** |
| Clean provider execution | 0/48 | **40/48 (83.3%)** |
| Invalid calls reaching provider | **111** | **0** |
| Mean latency | 38.0 s | **19.2 s** |
| Mean processed input | 143,280 | **75,403** |

This is a **64.6 percentage-point** exact-success increase. Mean latency fell 49.5%, and processed input fell 47.4%.

Matched-pair outcomes:

- PR-only success: **31**
- Base-only success: **0**
- Both succeeded: 13
- Neither succeeded: 4

All 31 disagreements favored the PR (exact McNemar `p = 9.31e-10`).

### Scenario breakdown

| Scenario | Base | PR |
| --- | ---: | ---: |
| Normal complex construction | 4/15 | **14/15** |
| Deliberately invalid first call, then correction | 1/9 | **9/9** |
| Evolving provider schema | 5/15 | **12/15** |
| Crowded tool context | 3/9 | **9/9** |

The crowded-context arm exposed **229 callable tools in both versions**: 29 built-in/plugin tools, 2 OpenWork Cloud MCP tools, and 198 decoys. A separate real OpenWork/GPT visibility probe successfully called a decoy tool, confirming that this was actual model tool context rather than a reported configuration count.

“Exact” means the downstream provider observed and accepted a payload matching every requested value, and the assistant confirmed the provider result. “Clean” means one exact provider call and no malformed call reached the provider. The stronger base model sometimes recovered by trial and error, so eventual success and clean execution are intentionally reported separately.

### What “18 forced-correction sessions” means

A session is one fresh OpenWork agent conversation running one task. The 18 correction sessions are **9 matched comparisons**: 3 models × 3 cases × base/PR. The first payload is deliberately incomplete. This checks whether OpenWork blocks it locally, gives GPT a useful correction, and permits a correct retry. On the PR arm, all 9 deliberately malformed attempts were blocked locally, all 9 retries completed correctly, and none of those malformed attempts reached the provider.

A tool status such as `error` describes one tool attempt, not the benchmark outcome. The base often guessed arguments without a schema, sent a malformed request to the provider, and correctly received an error. Conversely, a session can say `completed` while its final values are wrong. The provider-observed exact payload is therefore the primary grade.

### Evolving-schema caveat

This section should not be read as 15 guaranteed stale-digest trials. A live GPT session can search again after the provider changes and receive the current schema before executing. Only one run produced an explicit `capability_schema_changed` response. The 12/15 result shows better real-path adaptation and zero malformed provider calls, but it does not prove every stale invocation is detected and does not resolve the optional-digest bypass concern.

### Relationship to the earlier benchmark

The [earlier 138-session evidence](https://github.com/different-ai/openwork/pull/2899#issuecomment-5012033274) used real GPT and OpenCode but emulated the two Den contract behaviors. This exact-commit, real OpenWork/Den A/B is the stronger primary result. The lower base success here is not forced to zero: GPT-5.5 is allowed to recover through real provider errors, and those retries remain visible in the invalid-call, latency, and token metrics.

### Merge posture

This is strong evidence that the PR materially improves reliable complex tool use in the real post-connection OpenWork path, especially with a crowded tool context. It does **not** make the PR merge-ready. Keep it in draft until the existing blockers are resolved and retested:

1. JSON Schema dialect handling.
2. Process-global `$id`/validator-cache isolation and retained-memory growth.
3. No-argument omitted-body regression.
4. Optional schema digest allows drift checking to be bypassed.
5. Multi-match searches can place several MiB of schemas into model context.

PR #2943, opened 2026-07-21
fix(app): never restore the side panel on app load

benjaminshafii:
## Problem

On every app launch/reload, the right side panel came back in whatever state you last left it — anyone who once opened the **Extensions** panel got it splashed over the app on every reload, instead of landing on the chat.

## Fix

`sidePanelState` (the per-session `panel | extensions | voice` choice) is now **in-memory only**: it is no longer written to or restored from the persisted UI state (`openwork:ui-state:v1`). Every app load starts with just the chat. In-app toggling behaves exactly as before, and legacy `sidePanelState` entries already in localStorage are ignored and dropped on the next persist. All other persisted layout state (sidebar widths, right-sidebar expansion, menu visibility) is untouched.

- `apps/app/src/react-app/shell/ui-state-store.ts`: stop persisting/restoring `sidePanelState`; delete the now-dead legacy normalization helpers.
- `apps/app/tests/ui-state-store.test.ts`: regression tests — persisted JSON contains no `sidePanelState`; a seeded legacy `sidePanelState` in localStorage is ignored at startup while other fields still restore; in-memory toggling still works.
- `evals/`: voiceover + coded fraimz flow `side-panel-not-restored-on-load`.

## Tests

- `bun test tests/` (apps/app) — 369 pass, 0 fail
- `pnpm fraimz --flow side-panel-not-restored-on-load` against the real Electron app in a Daytona sandbox — **PASSED** (proof posted below): open Extensions → reload → same task returns with chat only, and the persisted UI state no longer contains any side panel entry.

### DIFF
--- a/apps/app/src/react-app/shell/ui-state-store.ts
+++ b/apps/app/src/react-app/shell/ui-state-store.ts
@@ -18,27 +18,7 @@ export const SIDE_PANEL_ITEMS = ["panel", "extensions", "voice"] as const;
 export type SidePanelItem = (typeof SIDE_PANEL_ITEMS)[number];
 export type SidePanelState = Record<string, SidePanelItem | null>;
 
-const LEGACY_UNIFIED_PANEL_ITEMS = ["browser", "artifacts"] as const;
-type LegacyUnifiedPanelItem = (typeof LEGACY_UNIFIED_PANEL_ITEMS)[number];
-
-function normalizeSidePanelItem(value: unknown): SidePanelItem | null {
-  if (value === null) {
-    return null;
-  }
-
-  if (isSidePanelItem(value)) {
-    return value;
-  }
-
-  if (LEGACY_UNIFIED_PANEL_ITEMS.includes(value as LegacyUnifiedPanelItem)) {
-    return "panel";
-  }
-
-  return null;
-}
-
 export type PersistedUiState = {
-  sidePanelState?: SidePanelState;
   applicationMenuVisible?: boolean;
   workspaceLeftSidebarWidth?: number;
   workspaceRightSidebarExpanded?: boolean;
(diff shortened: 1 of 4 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-21T08-03-06-073Z`

Full frame proof with validated screenshots: `evals/results/2026-07-21T08-03-06-073Z/fraimz.html` (re-run: `pnpm fraimz --flow side-panel-not-restored-on-load`)

### ✅ side-panel-not-restored-on-load — Reloading the app shows only the chat — the side panel is never restored on launch
_User-facing flow demo_

1. **A fresh task opens with the chat only and the Extensions side panel closed**
   > 🎙 Alex is working in a task in OpenWork. It is just the conversation: the composer at the bottom, the tool rail on the right edge, and no side panel taking up half the window.
   - ✅ Visible text does not include "Extensions (Legacy)"
   - 📸 `side-panel-not-restored-on-load-01-chat-only.png` — 4 validations passed

   

2. **Clicking the Extensions rail button opens the panel next to the chat**
   > 🎙 Alex clicks the Extensions button on the rail, and the Extensions panel opens next to the chat — handy for a quick look at apps and plugins.
   - ✅ Visible text includes "Extensions (Legacy)"
   - 📸 `side-panel-not-restored-on-load-02-extensions-open.png` — 5 validations passed

   

3. **Reloading the app returns to the same task without restoring the side panel**
   > 🎙 The app reloads, just like quitting and reopening OpenWork. The same task comes back — but the Extensions panel does not. It is just the chat again, and the app's saved layout no longer remembers any side panel at all.
   - ✅ Visible text does not include "Extensions (Legacy)"
   - 📸 `side-panel-not-restored-on-load-03-chat-after-reload.png` — 4 validations passed

   

   - ✅ Script frame 1 narrated: "Alex is working in a task in OpenWork. It is just the conversation: the composer at the "
   - ✅ Script frame 2 narrated: "Alex clicks the Extensions button on the rail, and the Extensions panel opens next to th"
   - ✅ Script frame 3 narrated: "The app reloads, just like quitting and reopening OpenWork. The same task comes back — b"

PR #2958, opened 2026-07-21
fix(app): keep model controls available during generation

reachjalil:
## Summary

- keep the composer model and thinking-mode selectors available during ordinary assistant generation
- track steered runs separately and disable both selectors from steer submission until the run returns idle
- close/guard the compact model picker while steering
- add regression coverage for the normal-generation versus steering distinction

## Why

The composer previously disabled model and thinking selection for every busy session. That prevented users from preparing the next run while an ordinary response was loading. Steering remains protected because changing execution settings during an active steer can cause inconsistent behavior.

## Checks

- `pnpm --filter @openwork/app exec bun test tests/composer-model-controls.test.ts` — passed (2 tests)
- `pnpm --filter @openwork/app test` — passed (372 tests)
- isolated Electron `app-smoke` eval on the Hub-assigned CDP port — passed

## Manual verification

Not run — the isolated merge-readiness profile was fresh and had no onboarded workspace or usable model, so a real generation/steer journey was unavailable.

Reviewer steps:

1. Start a task with a configured model.
2. While the first response is generating, confirm model and thinking selectors remain usable.
3. Submit a steering follow-up while that response is active.
4. Confirm both selectors disable for the steered run and re-enable when the run returns idle.

## Risk

Low and UI-local. No API, IPC, persistence, or model request shape changes.

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/composer/composer.tsx
+++ b/apps/app/src/react-app/domains/session/surface/composer/composer.tsx
@@ -61,6 +61,7 @@ type ComposerProps = {
   onQueue: () => void | Promise<void>;
   onStop: () => void | Promise<void>;
   busy: boolean;
+  steering: boolean;
   submissionPreparing: boolean;
   queuedCount: number;
   disabled: boolean;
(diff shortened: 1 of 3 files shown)


reachjalil:
@benjaminshafii This is ready and being merged into `dev`. The fix keeps model/thinking selection available during ordinary generation while preserving the steering lock.

PR #2978, opened 2026-07-22
feat(app): make session pinning global

reachjalil:
## Summary

- surface pinned sessions from every workspace in one global **Pinned** section at the top of the sidebar
- remove pinned roots from their workspace lists so each pinned session appears once without consuming workspace preview slots
- preserve pin order, expanded descendants, workspace identity, and accessible labels
- add focused root-filter tests and a coded two-workspace pin/unpin eval

## Why

Pin persistence was already global, but the sidebar only promoted a pinned session inside its workspace. A collapsed workspace could therefore hide a supposedly pinned session. This change makes the presentation match the global persistence model.

## Checks

- `pnpm --filter @openwork/app typecheck` — passed after rebasing onto current `upstream/dev`
- `pnpm --filter @openwork/app exec bun test --isolate tests/session-sidebar-utils.test.ts` — passed, 2 tests
- `global-session-pinning` coded Electron eval — passed once against the feature commit: two sessions in separate workspaces appeared exactly once in the global section, then unpinning returned one to its workspace

## Verification gap

Two post-rebase eval reruns did not complete in the isolated development runtime: one Electron/Vite process exited with `SIGINT` and produced a CDP timeout; a final retry timed out waiting for a newly created session. The two rebased upstream commits only affect markdown styling and MCP diagnostics, and the post-rebase typecheck and focused tests pass, but the final live rerun remains unconfirmed.

## User confirmation

Not yet manually confirmed by the requester.

## Risk / remaining behavior

- Pin state remains local client state in the existing persisted store.
- The global section promotes active root sessions, matching the existing root-pinning semantics; archived sessions stay in their workspace archive section.
- No video or screenshot evidence was requested.

### DIFF
--- a/apps/app/src/react-app/domains/session/control/session-control-actions.ts
+++ b/apps/app/src/react-app/domains/session/control/session-control-actions.ts
@@ -231,7 +231,7 @@ export function useSessionControlActions(input: UseSessionControlActionsInput) {
   const pinControlAction = useMemo<OpenworkControlAction>(() => ({
     id: "session.pin",
     label: "Pin or unpin a session",
-    description: "Toggle pin on a session. Pinned sessions float to the top of the sidebar.",
+    description: "Toggle pin on a session. Pinned sessions appear in a global section at the top of the sidebar.",
     sideEffect: "mutation",
     requiresArgs: true,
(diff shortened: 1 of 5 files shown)


reachjalil:
@benjaminshafii — PIN-ON-ME

PR #3072, opened 2026-07-24
feat(diagnostics): add Connect incident investigation portal

reachjalil:
## Value

Today the diagnostics service can test paths, but it cannot retain and investigate intermittent OpenWork Connect incidents after they happen. Support therefore sees a generic “disconnected” report without a customer timeline, failure cluster, or way to distinguish DNS/TLS/HTTP/MCP lifecycle symptoms.

This PR gives the diagnostics app one focused incident-investigation surface:

- bearer-authenticated metadata intake at `/api/connections/incidents`
- Redis-backed seven-day rolling retention, bounded to 10,000 incidents, with retry deduplication
- authenticated `/connections` portal with organization, client, source, phase, outcome, error, network, HTTP, and time filters
- failure clusters and a correlated chronological timeline
- authenticated POST-only raw organization/client lookup that immediately HMACs values so identities never enter filter URLs

## Privacy boundary

The shared schema accepts only connection metadata: timestamps, lifecycle phase/outcome, allowlisted DNS/TCP/TLS class, HTTP status, retryability, duration, failure count, categorical platform, versions, and opaque correlation IDs.

It rejects unknown fields and cannot carry emails, member identity, prompts/messages, tokens/cookies, URLs, tool names/arguments, or request/response bodies.

## Scope

This is the diagnostics-app slice only. It does not instrument Den, the desktop client, plugins, or OpenCode. The small `packages/types` incident contract is intentionally shared with the two sibling PRs so every slice independently builds from `dev`; after the first slice merges, the identical contract patch drops out of the others.

## Checks

- shared types build — passed
- diagnostics incident tests — 2 passed
- diagnostics production build — passed
- `git diff --check` — passed

## Rollout

Uses the existing diagnostics bearer and Redis configuration. Production should rate-limit `/api/connections/incidents` at the firewall as documented. No production deployment or live customer verification is included.

Split 1 of 3. Merge this first, followed by the Den relay and desktop client slices.


reachjalil:
PR chain: **#3072 Diagnostics portal → #3073 Den relay → #3074 desktop client**. Each PR targets `dev` and carries the identical small shared incident contract so it can build independently; runtime files do not overlap. After #3072 merges, update/rebase #3073 so the contract-only overlap disappears, then do the same for #3074 after #3073.

PR #3073, opened 2026-07-24
feat(den): correlate and relay Connect incidents

reachjalil:
## Value

When OpenWork Connect fails today, Den has no durable support signal that distinguishes “the request never reached Den” from “Den received the MCP request and rejected or lost it during initialize/tools discovery.” The desktop also needs an authenticated, organization-scoped route for submitting its own sanitized observation.

This PR makes Den the privacy and correlation boundary between opted-in clients and the diagnostics service:

- authenticated organization-member intake for bounded desktop incident batches
- organization identity derived from Den auth context rather than trusted from the client
- purpose-separated HMAC-SHA-256 pseudonyms for organization and random client IDs
- best-effort forwarding to the diagnostics intake with a three-second timeout
- Den-side observations for initialize, initialized notification, tools/list, HTTP failures, preflight failures, and transport exceptions
- stable client correlation accepted only from the UUID-shaped managed header

The result tells support whether an opted-in affected client reached Den, which MCP phase failed, and whether Den and desktop observations belong to the same pseudonymous client—without storing raw member identity.

## Opt-in guarantee — disabled by default

Diagnostics forwarding requires both independent gates:

1. **Operator opt-in:** Den makes no outbound diagnostics request unless `DEN_DIAGNOSTICS_BEARER_TOKEN` is explicitly configured. The default origin alone does not enable reporting.
2. **User opt-in:** Den constructs and forwards MCP lifecycle observations only when the request carries a valid `x-openwork-connect-client` UUID. The official desktop adds that header only after the user explicitly enables **Share desktop Connect diagnostics**.

Without a valid opted-in client header, Den does not construct or forward a lifecycle incident. The authenticated desktop intake is only called by the opted-in client reporter. No prompt content, credentials, URLs, tool payloads, email address, or raw member identity are collected.

## Scope

This is the Den integration slice only. It does not add the diagnostics portal, desktop UI/reporting, plugins, OpenCode changes, or a Den database/outbox. Den forwarding is deliberately best-effort so diagnostics never adds request latency or blocks Connect.

The small `packages/types` incident contract is identical to the contract in the diagnostics and client sibling PRs so this branch independently builds from `dev`. The diagnostics portal has now merged, so its identical contract patch drops out here.

## Checks on the current opt-in revision

- Den diagnostics tests — 3 passed
- Den typecheck — passed
- `git diff --check` — passed

Earlier feature-branch checks, before the opt-in follow-up commit:

- shared types build — passed
- Den supporting utils build — passed
- Den production build — passed

## Rollout

Uses `DEN_DIAGNOSTICS_ORIGIN` and `DEN_DIAGNOSTICS_BEARER_TOKEN`. If the bearer token is absent, all forwarding is disabled without affecting MCP traffic. Even when the operator enables forwarding, Den-side lifecycle observations still require the opted-in client header.

No database migration, deployment, or live on-prem verification is included.

Split 2 of 3. Merge after the diagnostics portal and before the desktop client.


reachjalil:
PR chain: **#3072 Diagnostics portal → #3073 Den relay → #3074 desktop client**. Each PR targets `dev` and carries the identical small shared incident contract so it can build independently; runtime files do not overlap. After #3072 merges, update/rebase this PR so the contract-only overlap disappears before review/merge.

PR #3076, opened 2026-07-24
fix(enterprise-mcp): require invalid_token proof before destroying credentials

benjaminshafii:
## Problem

Prod DB forensics (Den, PlanetScale) showed **every runtime credential invalidation** was written by `invalidateTerminallyRejectedCredential`: any unresolved HTTP 401 on an MCP resource request wiped the member's stored access+refresh tokens and forced a reconnect. Confirmed victims: Notion (ben, 2 days after connect), Stripe (ben, 62 min after connect — token-TTL boundary), Salesforce (jalil, ~2h40m after connect), Granola (jalil). Background `search_capabilities` probes could trigger it on connections the user never touched, and invalidations were completely silent (the probe-failure log is gated behind query-match score).

This is the root of the 'users get prompted to reconnect way too often' issue reported by Blue Yonder (Rashmi) and reproduced on our own first-party connectors.

## Fix

- Credential invalidation only during **user-initiated tool execution** — never during tool-discovery probes
- 401 is terminal **only with an explicit `WWW-Authenticate: Bearer error="invalid_token"` challenge** (mirrors the existing 403 bearer/scope gate); a bare 401 (gateway redeploy, VPN flap, session flush) no longer kills the credential
- 403 policy unchanged
- Every invalidation now emits a `credential-invalidation` diagnostic; Den logs `external_mcp_credential_invalidated` (referenceId, connectionId, phases, httpStatus, invalidToken) unconditionally

## Tests run

- `pnpm --filter @openwork/enterprise-mcp-client typecheck` — pass
- `pnpm --filter @openwork/enterprise-mcp-client test` — **55 pass / 0 fail** (4 new policy tests: bare-401 preserved, invalid_token invalidates + diagnostic, probe safety, 403 regression guard)
- `pnpm exec tsc --noEmit -p ee/apps/den-api/tsconfig.json` — pass
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/external-mcp-diagnostics.test.ts test/mcp-oauth-refresh-lifecycle.test.ts` — **79 pass / 4 skip / 0 fail**
- `pnpm fraimz --flow enterprise-mcp-terminal-401-guard` — **PASSED** (internal demo driving the real client through all three scenarios; proof comment below)

Note: this package has no runtime path into the local Electron app (consumed by den-api cloud), hence the internal-kind fraimz.

Follow-ups tracked separately: per-identity refresh single-flight (M3 race), Slack/github providers issuing no refresh token (M4).

### DIFF
--- a/ee/apps/den-api/src/capability-sources/enterprise-mcp-client-adapter.ts
+++ b/ee/apps/den-api/src/capability-sources/enterprise-mcp-client-adapter.ts
@@ -98,6 +98,17 @@ function diagnosticSink(tracker: ExternalMcpDiagnosticTracker) {
     // still available to package consumers, but must not overwrite that richer
     // Den evidence after a response settles.
     if (event.kind === "request") return
+    if (event.kind === "credential-invalidation") {
+      console.error("external_mcp_credential_invalidated", {
+        referenceId: tracker.referenceId,
+        connectionId: event.connectionId,
+        operationPhase: event.operationPhase,
+        requestPhase: event.requestPhase,
+        httpStatus: event.httpStatus,
+        invalidToken: event.invalidToken,
+      })
+      return
+    }
     const phase = diagnosticPhase(event)
     if (event.outcome === "started") {
       tracker.begin(phase)
--- a/evals/flows/enterprise-mcp-terminal-401-guard.flow.mjs
+++ b/evals/flows/enterprise-mcp-terminal-401-guard.flow.mjs
@@ -1,0 +1,151 @@
+/**
+ * Internal demo: the terminal-401 credential guard.
+ *
+ * Production forensics showed the enterprise MCP client destroyed stored
+ * connection credentials on any unresolved 401 — including transient provider
+ * blips and background search probes — forcing users to reconnect (Notion,
+ * Stripe, Salesforce, Granola cases confirmed in the Den DB). This flow proves
+ * the new policy on the real client (requiresApp: false): evidence is claims +
+ * assertions + real command output from driving createEnterpriseMcpClient
+ * through scripted provider scenarios.
+ */
+import { spawnSync } from "node:child_process";
+import { readFile } from "node:fs/promises";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+import { loadVoiceoverParagraphs } from "../runner/voiceover.mjs";
+
+const FLOW_ID = "enterprise-mcp-terminal-401-guard";
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), "..", "..");
+const CLIENT_SOURCE = join(ROOT, "packages", "enterprise-mcp-client", "src", "enterprise-mcp-client.ts");
+const ADAPTER_SOURCE = join(ROOT, "ee", "apps", "den-api", "src", "capability-sources", "enterprise-mcp-client-adapter.ts");
+
+const vo = await loadVoiceoverParagraphs(FLOW_ID);
+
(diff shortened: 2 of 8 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-24T10-02-17-596Z`

Full frame proof with validated screenshots: `evals/results/2026-07-24T10-02-17-596Z/fraimz.html` (re-run: `pnpm fraimz --flow enterprise-mcp-terminal-401-guard`)

### ✅ enterprise-mcp-terminal-401-guard — A transient 401 no longer destroys stored connection credentials
_Internal demo_

1. **Credential invalidation is gated to tool execution with an invalid_token challenge**
   > 🎙 Here is the new rule in the client itself: a connection credential may only be destroyed during a real tool call the user asked for, and only when the provider proves the token itself is bad with an explicit invalid_token challenge. A plain 401 blip is no longer a death sentence.
   - ✅ Only user-initiated tool execution may invalidate (background probes are excluded)
   - ✅ A 401 is terminal only with an explicit invalid_token bearer challenge
   - ✅ Every invalidation emits a credential-invalidation diagnostic
   - ✅ The guard function is present in the client source
2. **The stored credential survives a bare 401 on a user-initiated tool call**
   > 🎙 First scenario: the user runs a tool and the provider hiccups with a bare 401, the kind a gateway redeploy or a VPN change produces. The call fails, but the stored credential survives untouched — no reconnect prompt will greet the user afterwards.
   - ✅ The demo drive of the real client exits cleanly
   - ✅ The demo prints machine-readable act evidence
   - ✅ The demo run produced parseable act results
   - ✅ The demo run includes act "transient-401-on-execute"
   - ✅ The act is a user-initiated tool call
   - ✅ The stored credential survived the bare 401
   - ✅ No invalidation was written to persistence
3. **A tool-discovery probe cannot invalidate, even on an invalid_token challenge**
   > 🎙 Second scenario: a background capability search probes the same connection and the provider even claims invalid_token. Background probes are never allowed to destroy credentials, so the stored tokens survive this too.
   - ✅ The demo run produced parseable act results
   - ✅ The demo run includes act "invalid-token-on-discovery-probe"
   - ✅ The act is a background discovery probe
   - ✅ The provider claimed invalid_token
   - ✅ The stored credential survived the probe
   - ✅ No invalidation was written to persistence
4. **Provider-proven token rejection still invalidates and is now observable**
   > 🎙 Third scenario: the user runs a tool and the provider genuinely rejects the token with an invalid_token challenge. Now the client does invalidate the credential, exactly once, and for the first time it emits a diagnostic that operators can see in the logs — and the full test suite confirms nothing else regressed.
   - ✅ The demo run produced parseable act results
   - ✅ The demo run includes act "invalid-token-on-execute"
   - ✅ The credential was invalidated
   - ✅ Exactly one invalidation was written
   - ✅ Exactly one credential-invalidation diagnostic was emitted
   - ✅ The diagnostic records the user-initiated phase
   - ✅ The diagnostic records the provider's proof
   - ✅ Den logs external_mcp_credential_invalidated unconditionally (no more silent kills)
   - ✅ The full enterprise MCP client test suite passes
   - ✅ Script frame 1 narrated: "Here is the new rule in the client itself: a connection credential may only be destroyed"
   - ✅ Script frame 2 narrated: "First scenario: the user runs a tool and the provider hiccups with a bare 401, the kind "
   - ✅ Script frame 3 narrated: "Second scenario: a background capability search probes the same connection and the provi"
   - ✅ Script frame 4 narrated: "Third scenario: the user runs a tool and the provider genuinely rejects the token with a"

PR #3087, opened 2026-07-24
fix(installer): trust every OS certificate source on inspected networks

benjaminshafii:
## The bug

On the Blue Yonder POC the installer could not resolve the install link:

> Reached your workspace, but the secure connection isn't trusted on this computer yet.

Their network re-signs HTTPS, so `openwork-poc.blueyonder.com` presents a corporate-CA
chain. The installer already tries to work around Bun's bundled-roots-only `fetch` by
loading OS certificates — but it preferred one source over another:

```ts
const nodeCerts = loadNodeSystemCertificates()
if (nodeCerts.length > 0) return nodeCerts   // never enumerates the platform stores
```

Any non-empty runtime list short-circuits the thorough platform enumeration. A runtime
that reports *some* roots (the public ones) but not the corporate one therefore wins,
and the corporate CA is never loaded. macOS additionally read only `System.keychain`.

## The fix

- **Union all sources** — runtime + platform stores + `NODE_EXTRA_CA_CERTS`, deduped. A
  partial source can no longer mask a complete one.
- **`NODE_EXTRA_CA_CERTS`** — the deterministic override for locked-down fleets, matching
  the variable the desktop shell already honors.
- **macOS also reads `SystemRootCertificates.keychain`.** The user-writable login keychain
  is deliberately *excluded*: `security find-certificate` ignores trust settings, so
  enumerating it would let any local process widen what the installer trusts. User-domain
  corporate roots go through `NODE_EXTRA_CA_CERTS`.
- **Per-source diagnostics** at startup and in the TLS rejection body, so next time an
  untrusted certificate is distinguishable from a trust store that could not be read.

## Fail repro → valid repro

Built a Blue-Yonder-shaped chain (root → intermediate → leaf), none of it in Bun's
bundled roots, serving the Den `install-config` endpoint. Drove the real compiled binary
through its actual `/api/resolve-link` route.

**Fail repro** — corporate root not available to the installer:

```
REJECTED [install_link_tls_untrusted]
message: Reached your workspace, but the secure connection isn't trusted on this
         computer yet. This usually means your company inspects secure traffic...
diagnostics: OS trust store: runtime=0 macos-keychains=165 NODE_EXTRA_CA_CERTS=0
```

This reproduces the reported message byte-for-byte, and it still fails after the fix —
the change does not blanket-trust anything.

**Valid repro** — same binary, same host, root provided the way IT would:

```
NODE_EXTRA_CA_CERTS=/…/corporate-root.pem
RESOLVED ok=True source='install link'
diagnostics: OS trust store: runtime=0 macos-keychains=165 NODE_EXTRA_CA_CERTS=1
```

Public HTTPS is unaffected with the corporate root loaded (`api.github.com -> 200`);
Bun's `tls.ca` augments the default roots rather than replacing them.

## Tests

`pnpm --dir apps/installer test` → **44 pass, 0 fail** (8 new).

Each fix was reverted in isolation to prove its test is load-bearing:

| Reverted | Result |
| --- | --- |
| Restore the short-circuit | union test fails (3 fail) |
| Drop `NODE_EXTRA_CA_CERTS` | both escape-hatch tests fail |
| Re-add the login keychain | security guard fails |

## Not covered

The short-circuit only bites where the runtime returns a partial non-empty list, which is
the **Windows** case — Bun returns 0 on macOS, so the union's effect is not observable
end-to-end on this machine and is covered by unit tests instead. Windows Daytona
validation is blocked (org credits depleted). Worth a run on a real Windows box with the
corporate root in `LocalMachine\Root` before relying on it at Blue Yonder.

### DIFF
--- a/apps/installer/README.md
+++ b/apps/installer/README.md
@@ -27,6 +27,26 @@ a DMG containing the generic `OpenWork Installer.app`; Windows ships as the bare
 generic `OpenWork-Installer-win-x64.exe`. A generic UI build gates on the paste
 screen until the user provides their OpenWork install link.
 
+## Networks that inspect TLS
+
+Bun's `fetch` trusts only its bundled roots, so on a network that re-signs HTTPS the
+installer additionally trusts the OS stores (Windows `Root`/`CA` for both machine and
+user, macOS `System.keychain` and `SystemRootCertificates.keychain`) plus whatever the
+runtime reports. The user-writable macOS login keychain is deliberately excluded, since
+`security find-certificate` ignores trust settings.
+
+When a corporate root lives somewhere the installer cannot enumerate, IT can point it at
+a PEM bundle — the same variable the desktop shell honors:
+
+```bash
+NODE_EXTRA_CA_CERTS=/path/to/corporate-root.pem
+```
+
+On startup the installer logs what each source contributed
+(`OS trust store: runtime=0 windows-cert-stores=214 NODE_EXTRA_CA_CERTS=1`), and a TLS
+rejection repeats that summary in `trustSources` so support can tell an untrusted
+certificate apart from a trust store that could not be read.
+
 ## Local development
 
 ```bash
--- a/apps/installer/src/server.ts
+++ b/apps/installer/src/server.ts
@@ -2,7 +2,7 @@ import { randomBytes } from "node:crypto"
 
 import { installerConfigSourceLabel, parseInstallLinkInput, resolveInstallLinkConfig, type InstallerConfigResolution } from "./config"
 import { installStatus, launchInstalledApp, runInstall } from "./install"
-import { loadSystemCaCertificates } from "./system-ca"
+import { loadSystemCaBundle, summarizeSystemCaSources } from "./system-ca"
 import { renderInstallerHtml } from "./ui-html"
 
 export type InstallerServer = {
@@ -39,7 +39,11 @@ export function startInstallerServer(initialResolution: InstallerConfigResolutio
   let resolution = initialResolution
   // Warm the OS trust-store CA cache now so the first resolve-link fetch does
   // not spend its 10s abort budget waiting on PowerShell/security exports.
-  void loadSystemCaCertificates()
+  // Record what each source produced: a TLS failure is otherwise indistinguishable
+  // from silent enumeration failure when supporting a locked-down fleet.
+  void loadSystemCaBundle().then((bundle) => {
+    console.log(`[openwork-installer] OS trust store: ${summarizeSystemCaSources(bundle.sources)}`)
+  })
 
   const server = Bun.serve({
     hostname: "127.0.0.1",
@@ -74,7 +78,12 @@ export function startInstallerServer(initialResolution: InstallerConfigResolutio
             }
             if (result.status === "unreachable") {
               if (result.reason === "tls") {
-                return Response.json({ error: "install_link_tls_untrusted", message: tlsUntrustedMessage(installLink) }, { status: 400 })
+                const bundle = await loadSystemCaBundle()
+                return Response.json({
+                  error: "install_link_tls_untrusted",
+                  message: tlsUntrustedMessage(installLink),
+                  trustSources: summarizeSystemCaSources(bundle.sources),
+                }, { status: 400 })
               }
               return Response.json({ error: "install_link_unreachable", message: "Could not reach your workspace. Check your internet or VPN connection and try again." }, { status: 400 })
             }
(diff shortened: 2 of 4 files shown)


benjaminshafii:
### Windows validation: added, not yet run

To be explicit about the gap in the description: **everything above was validated on macOS only.** That is the weaker half of the evidence, because Bun reports 0 system certificates on macOS, so the short-circuit this PR removes never triggered there. The macOS runs exercised the `NODE_EXTRA_CA_CERTS` path and the unit-tested union; they never exercised the real failure, which needs a runtime returning a *partial* CA list alongside a corporate root in `LocalMachine\Root`.

Pushed two commits that close it on real Windows (`d58e602`, plus a CI nudge):

- `apps/installer/scripts/make-corporate-ca.mjs` — cross-platform corporate chain (root → re-signing intermediate → leaf), none of it in Bun's bundled roots.
- `apps/installer/scripts/trust-repro.mjs` — drives the compiled binary's real `/api/resolve-link` route and self-asserts via `--expect resolved|tls-untrusted`.
- `.github/workflows/ci-installer-windows-trust.yml` — `windows-2022` job that runs the repro twice: once with the root untrusted (expects `tls-untrusted`), then again with **nothing changed except `Import-Certificate … -CertStoreLocation Cert:\LocalMachine\Root`** (expects `resolved`). It also prints what each CA source contributed and re-checks public HTTPS.

That second run is the assertion that actually matters: the only variable is the Windows machine trust store, so it passes only if the installer enumerates it.

**Status: the job has not run yet.** GitHub is not picking the pushes up — the branch head on the remote is `d58e602`, but this PR still reports `8baf9cae` and no checks have been queued for the newer commits (GitHub Status reports all systems operational). So the Windows result is still unproven, and this PR should not be treated as Windows-validated until that job goes green. Re-running CI once GitHub syncs should be enough.

benjaminshafii:
## Correction: the root cause claimed here is disproven

Windows CI now exists (#3089) and it contradicts this PR's headline diagnosis. **The
short-circuit removed here was not the cause of the field failure.**

The claim was that `if (nodeCerts.length > 0) return nodeCerts` skipped the platform
enumeration, so the corporate root in `LocalMachine\Root` was never loaded. That requires
Bun's `tls.getCACertificates("system")` to return a *partial, non-empty* list. On
`windows-2022` with Bun 1.3.9 it returns **zero**:

```
sources: runtime=0 windows-cert-stores=566 NODE_EXTRA_CA_CERTS=0
```

With `runtime=0` the short-circuit never fires, so the pre-fix code fell through to
PowerShell enumeration exactly as intended. I ran the same repro against the **pre-fix**
sources on `windows-2022` to confirm ([run](https://github.com/different-ai/openwork/actions/runs/30133524360)):

```
PROBE pre-fix loadSystemCaCertificates returned: 566
fail repro:  outcome: tls-untrusted   OK
(Import-Certificate into Cert:\LocalMachine\Root)
valid repro: outcome: resolved        OK
```

The old code picks the corporate root up from the Windows machine trust store on its own.
Bun returns 0 on macOS *and* Windows, which means the short-circuit is effectively dead
code on both platforms this ships to — removing it is tidier, but it fixed nothing.

### What this PR is actually worth

- **`NODE_EXTRA_CA_CERTS`** — a real escape hatch that did not exist before.
- **Per-source diagnostics** — the thing that will identify the true cause. A field report
  showing `windows-cert-stores=0` means enumeration failed; a non-zero count means the
  root is not in the stores we read, or the failure is not trust at all.
- Union + macOS `SystemRootCertificates` — defensible hardening, not a cure.

### Where the real cause probably lives

Still open. The strongest remaining candidates, in order:

1. **PowerShell enumeration failing or timing out.** `COMMAND_TIMEOUT_MS` is 10s and a
   locked-down fleet may block `powershell.exe` (execution policy, AppLocker, Constrained
   Language Mode) or be slow enough to blow the budget. Either way the result is 0 certs
   and this exact error — and unlike the short-circuit, this is entirely plausible on a
   managed machine.
2. **The root is not in a store we enumerate.** We read `Root` and `CA` for both
   `LocalMachine` and `CurrentUser`; a GPO/CTL-deployed or Enterprise Trust placement may
   sit elsewhere.
3. **A proxy in the path.** If the machine uses an authenticated proxy, `tls.ca` may not
   apply to the CONNECT tunnel, and the resulting error would still be classified as TLS.
4. **Misclassification.** `isTlsError()` matches on message text
   (`/certificat|CERT_|TLS|SSL|handshake|.../i`), so a non-trust failure whose message
   merely mentions TLS lands on the certificate copy and sends everyone down this path.

**Blue Yonder should not be told this is fixed.** The next step is a build with these
diagnostics in front of their machine, and reading `trustSources` off the failure.

PR #3097, opened 2026-07-25
refactor: collapse path, base-URL and workspace-store duplication into three primitives

benjaminshafii:
## Why

A scan for fragmentation across the monorepo turned up three clusters where the
same logic had been re-implemented many times over, and where the copies had
already drifted into real defects. This collapses each cluster onto one
primitive and fixes the drift.

I deliberately did **not** try to migrate every call site of every cluster — the
trailing-slash pattern alone appears ~145 times. Each primitive lands with the
provably-divergent consumers migrated, so the diff stays reviewable.

## 1. Path resolution — `packages/paths`

~35 resolvers across the server, desktop shell, installer and bootstrap CLI.
Three of them carried a comment saying they "must agree byte-for-byte" with each
other, which is what a shared module is for. Four had already drifted:

| # | Defect | Effect |
|---|---|---|
| D1 | `server.ts` global-opencode-config route honored neither `OPENCODE_CONFIG_DIR` nor `XDG_CONFIG_HOME`; `mcp.ts` honored both | The desktop runtime sets `OPENCODE_CONFIG_DIR` under `OPENWORK_DEV_MODE=1`, so editing global config in Settings read and wrote the developer's **real** config instead of the dev sandbox |
| D2 | `config.ts` + `extensions/google-workspace.ts` ignored `APPDATA` / `XDG_CONFIG_HOME` for `server.json` | Server and desktop shell disagreed on the path on Windows |
| D3 | `runtimeDbPath` repeated the same wrong `~/.config/openwork` default | Same class of divergence |
| D4 | Desktop checked only root-level `opencode.json(c)`; the server also checked `.opencode/` | The runtime created a **shadowing** root config for workspaces that already had a hidden one |

D1 reproduced against `dev`'s exact logic:

```
expected (honors OPENCODE_CONFIG_DIR): /tmp/d1check/dev-opencode-config/opencode.json
origin/dev server.ts returns:          /Users/…/.config/opencode/opencode.json
D1 regression test on origin/dev => FAIL (bug reproduced)
```

The package ships raw ESM + hand-written declarations so the Electron main
process can `import` it directly (verified), and every resolver takes injectable
`env` / `homeDir` / `platform` so it is testable without touching the real
filesystem. The hardened `OPENCODE_CONFIG_DIR` validation from `mcp.ts` is kept
as the canonical behaviour.

## 2. Base URLs — `@openwork/types/url`

`embedded.ts` inlined its own dev/prod models-URL branch, so **`OPENCODE_MODELS_URL`
was silently ignored** on that path and a dev session without the local inference
stack got a dead catalog URL and an empty model list. It now uses the resolver
that honors the override and probes the local stack before falling back.
`cli.ts` set neither the models URL nor swept legacy config; both boot paths now
agree.

Also collapses the byte-identical copies: `DEN_API_BASE` had **four** independent
readers in den-web (one a bare inline expression in a layout), and the app had two
identical server-URL normalizers.

## 3. Workspace stores — `workspace-kv-store.ts`

Four stores each carried their own copy of the same ~50 lines. The real hazard
was not the line count but **four independent connection caches pointing at the
same SQLite file**.

| | before | after |
|---|---|---|
| store implementations | 4 | 1 factory |
| `CREATE TABLE` DDL copies | 8 | 1 builder |
| `dbByPath` caches | 4 | 1 |
| local `isRecord` | 4 | 1 |

`session_group_states` keeps its extra `schema_version` column via the factory's
`extraColumns` rather than being forced into a lossy shape. All four keep their
exported names and signatures, so callers are untouched and there is no migration.

## Tests

| Suite | Result |
|---|---|
| `apps/server` | **520 pass**, 10 skip, 0 fail |
| `apps/app` | 423 pass, 0 fail |
| `apps/desktop` | 109 pass, 1 skip |
| `apps/installer` | 46 pass |
| `packages/paths` | 16 pass |
| typecheck | server, app, types, paths, den-web all clean |

Regression tests added for D1 and for the models-URL override; both fail on `dev`
and pass here.

**One pre-existing failure fixed, not caused by this branch.** The runtime-config
store test merges the global MCP layer, so it observed whatever MCP servers the
developer happened to have in `~/.config/opencode`. It fails identically on
`origin/dev` on any machine with a global MCP entry — verified. It is now
hermetic.

## Proof

Ran the canonical core flow on Daytona against this branch (`kind: internal` —
these changes should be inert to the user, so the job is proving the core
experience is unchanged): open the app, send a message, get a response, reload,
session intact. Frames attached in the follow-up comment.

While doing that I found the core flow itself could pass **without the agent ever
responding**: the prompt is `"Reply with exactly: core-flow ok"`, so asserting the
page contains `"core-flow ok"` is satisfied by the echoed user bubble. The step
passed in 1.4s with a frame showing only the user message. Scoping the assertion
to an assistant message makes the same step wait 5.9s for the real reply.
Verified the new predicate fails a user-only transcript and passes a real one.

## Deliberately left as follow-ups

- `apps/orchestrator` (18 sites) and `packages/openwork-bootstrap` (5) strip a
  **single** trailing slash (`/\/$/`) where the rest of the repo strips all of them
  (`/\/+$/`), so `http://host//` normalizes differently across process boundaries
  and breaks origin equality and cache keys. Orchestrator is out of scope here and
  the bootstrap CLI ships as raw `.mjs` that cannot import this package's source.
- The remaining ~110 inline trailing-slash sites and the orchestrator's own copies
  of the env-store loader and path resolvers.
- Untouched `ee/apps/den-api` MCP URL normalization, to avoid conflicting with the
  open #2759.


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-25T10-37-12-566Z`

Full frame proof with validated screenshots: `evals/results/2026-07-25T10-37-12-566Z/fraimz.html` (re-run: `pnpm fraimz --flow core-flow`)

### ✅ core-flow — Open app, send a message, get a response, reopen with session intact

1. **App boots clean to a known route**
   - 📸 `core-flow-01-booted.png` — 4 validations passed

   

2. **A new session is created and becomes active**
   - 📸 `core-flow-02-session-created.png` — 3 validations passed

   

3. **The composer accepts a message and the agent responds**
   - 📸 `core-flow-03-task-response.png` — 4 validations passed

   

4. **Reopening restores the session and its message history**
   - 📸 `core-flow-04-reopened-session.png` — 4 validations passed

PR #3099, opened 2026-07-25
feat(den-web): rework "Continue on your computer" install step

benjaminshafii:
## What

Reworks step 2 of the organization install guide (`/install`) around the fresh-install case, matching the approved Paper design (`02B — Desktop Expanded`, macOS + Windows artboards).

Left column, top to bottom: eyebrow "NEXT, ON YOUR COMPUTER" → "Open the file you just downloaded" → file chip with the real installer filename → two numbered actions → calm amber OS trust note → blue handoff note → OR divider → demoted already-installed row with the single dark primary button → quiet "Nothing happening on your computer?" disclosure with the manual paste link.

Right column: existing installer preview, with the fake CTA flattened (transparent, hairline border, muted label, no shadow), the badge relabelled `PREVIEW · WHAT YOU'LL SEE ON YOUR COMPUTER` with a neutral grey dot, and the column stretched so both columns end level.

## Two things worth reviewing carefully

**Real filename, no invented version.** The Paper design showed `OpenWork-1.4.2-arm64.dmg`, which is fictional. The chip now shows the actual published asset name from `genericInstallerArtifactName()` (`OpenWork-Installer-mac-arm64.dmg`, `…mac-x64.dmg`, `…win-x64.exe`). It prefers the platform the user actually downloaded and otherwise uses browser platform detection. Linux has no stable name (its setup script embeds the org slug, which this page never receives), and macOS with an undetectable arch is also ambiguous — in both cases the chip is omitted and the copy does not name a file. No version number is displayed anywhere.

**Auto-advance: chose option (b), softened copy.** The design promised "this page moves to step 3 on its own". That is not implementable safely today, so the copy was changed rather than faked. The installer resolves its **own** activation grant from the same install-link token, so the web page's code is a different code and `/v1/install-connect/status` can never observe the installer's grant. A token-scoped "was a grant minted recently" endpoint would be cheap to add, but it cannot distinguish the installer from a second browser tab or a colleague opening the same link, and step 3 polls a code the installer never uses — so it would auto-advance the user to a step that cannot confirm anything. The blue row now states what actually happens: `<App> continues from here — The moment it opens, it brings up a sign-in page in your browser so you can approve this computer.`

No den-api changes.

## Files

- `ee/apps/den-web/app/(den)/_components/install-screen.tsx` — new step 2 body, per-OS guidance copy, platform/filename resolution, platform-by-href map, step 2 subtitle.
- `ee/apps/den-web/app/(den)/_components/installer-preview.tsx` — flattened fake CTA, relabelled neutral badge, `h-full` stretch, badge sits at the top on Windows.
- `ee/apps/den-web/app/(den)/_lib/install-download.ts` — `installerFileName()` mirroring den-api's `genericInstallerArtifactName()`.
- `evals/flows/org-install-activation.flow.ts`, `evals/flows/org-download-handoff.flow.mjs` — updated for the new copy and for opening the manual-paste disclosure before reading its text.

## Validation

- `cd ee/apps/den-web && pnpm typecheck` → pass
- `pnpm evals:typecheck` (repo root) → pass
- `pnpm lint` skipped: `next lint` was removed in Next 16 and the script is broken on `dev`.
- Drove the real page against a mock den-api (`/v1/install-config` + `/v1/install-connect/status`) with `pnpm dev:den:web`, at 1440×1000 and 900px wide, capturing macOS, Windows, Linux-fallback and clicked-platform variants. One dark primary button, both columns end level, nothing clips or overflows.

Known gaps: the preview illustration still renders macOS chrome for Linux visitors (pre-existing — it only models macOS and Windows), and the "Already have … on this computer?" line wraps to two lines at the narrower real column width where the design showed one.

### DIFF
--- a/ee/apps/den-web/app/(den)/_components/install-screen.tsx
+++ b/ee/apps/den-web/app/(den)/_components/install-screen.tsx
@@ -1,12 +1,12 @@
 "use client";
 
-import { DownloadPlatformGrid, type DownloadPlatformGroup, type DownloadPlatformOption } from "@openwork/ui/react";
-import { ChevronDown } from "lucide-react";
+import { detectPlatform, DownloadPlatformGrid, type DetectedPlatform, type DownloadPlatformGroup, type DownloadPlatformOption } from "@openwork/ui/react";
+import { ChevronDown, Download, ShieldCheck } from "lucide-react";
 import { useEffect, useMemo, useRef, useState, type ReactNode } from "react";
 import { useSearchParams } from "next/navigation";
 import { requestJson } from "../_lib/den-flow";
 import { getInstallConfigErrorMessage } from "../_lib/install-errors";
-import { buildInstallDownloadHref, type InstallPlatform } from "../_lib/install-download";
+import { buildInstallDownloadHref, installerFileName, type InstallPlatform } from "../_lib/install-download";
 import { isMobileUserAgent } from "../_lib/platform";
 import { InstallerPreview } from "./installer-preview";
 import { OnboardingShell } from "./onboarding-shell";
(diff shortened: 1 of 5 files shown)


benjaminshafii:
### Independent verification

Reviewed and re-ran this branch locally rather than relying on the implementing agent's report.

**Commands**

- `pnpm typecheck` (`ee/apps/den-web`) — passes
- `pnpm evals:typecheck` — passes
- `pnpm test` (`ee/apps/den-web`) — 45 pass / 3 fail / 2 errors. All pre-existing on `dev`: a stale expected string in `install-errors.test.ts` and an unbuilt `@openwork-ee/utils` module. Nothing in the touched files.

**Driven in a real browser** against a mocked `/v1/install-config`, at 1440 / 1024 / 768 / 390 px:

- No horizontal overflow at any width (`document.documentElement.scrollWidth === clientWidth`), including with the "Nothing happening on your computer?" disclosure open, where the tokenised URL truncates inside its row.
- Columns stack at narrow widths with the illustration dropping below rather than squashing.
- Step 1's expanded download card also checked at 390 px: the platform grid stacks cleanly and still marks the detected build "For your device".

**Both OS branches asserted from the DOM**

- Windows client hints → chip `OpenWork-Installer-win-x64.exe`, SmartScreen trust note, `data-preview-os="windows"`.
- Forced macOS hints → chip `OpenWork-Installer-mac-arm64.dmg`, Gatekeeper trust note, `data-preview-os="macos"`.

Worth noting for reviewers: a headless Chromium can advertise `userAgentData.platform: "Windows"` while carrying a Mac UA string. The page correctly trusts client hints over the UA string, so a Windows render on a Mac test host is the browser lying, not a detection bug.

**Not verified:** the eval flows were type-checked but not executed end to end.

PR #3110, opened 2026-07-25
chore: retire phantom capability surfaces and dead settings screens

benjaminshafii:
## Problem
The app advertises capabilities and screens that do not exist. During an enterprise walkthrough this cost real credibility — we were asked about a feature the agent still claims to have, and pointed at settings screens that silently redirect.

Verified phantoms:

1. **`help.capabilities` advertises "Automations — Schedule recurring tasks and background agents"** (`control-provider.tsx:814`). The scheduler was deleted in `53e0e1d11`; there is no scheduler, cron, job table, job route, or automations UI anywhere in the tree. We were advertising it **to the agent**.
2. **`skills-view.tsx`** (508 lines: view/edit/create/import/reveal) is imported by **nothing** — superseded by the unified inventory in `820d00ad3`.
3. **`cloud-marketplaces-view.tsx`** (989 lines) is referenced only by its own test and one stale doc line.
4. **`SettingsTab` declares `skills` and `cloud-marketplaces`** with labels, icons and descriptions, but neither is in any nav array and neither has a render `case` — the type was lying.
5. **`background-agents-screen.tsx`** promised scheduling with "Coming soon" while containing none, reachable by URL only.

## Change
Delete the dead code, remove the phantom capability, and make the copy honest. `-2026 / +5` lines.

Orphaned store methods removed after per-symbol proof of zero callers: `saveSkill`, `importLocalSkill`, `installSkillCreator`, `revealSkillsFolder` (+ the `skill-creator.md` template only that path consumed, + their i18n keys across all 10 locales). **`readSkill` and `uninstallSkill` were kept** — `mcp-view.tsx` uses them.

Deliberately **kept**: the legacy `/settings/skills` and `/settings/cloud-marketplaces` redirects (real deep-link back-compat), and the background-agents page itself (it lists real workers and mints connect tokens) — only its misleading "Coming soon" framing was removed.

Also corrected two docs click-paths that pointed at removed sub-tabs (`get-started.mdx`, `team-templates.mdx`), and the agent-facing skills path in `openwork-capabilities-knowledge.ts`.

## Verification
```
pnpm typecheck                            clean
pnpm --filter @openwork/app test          420 pass, 0 fail
pnpm --filter @openwork-ee/den-web build   succeeded
```
Final `rg` sweep for every deleted symbol/file returns **empty**. No behavior changes beyond removing false advertising.


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

benjaminshafii:

PR #3111, opened 2026-07-25
fix(settings): finish extension readiness and section routing wiring

benjaminshafii:
## Problem
Three pieces of **unfinished wiring** (not missing features) meant members could not tell which organization capabilities were ready versus needed setup. This was the single most confusing thing in an enterprise walkthrough: the app already knows the answer and could not say it.

## Fix 1 — organization readiness rows could never render
`connect-view.tsx` declares `marketplaceItems?: ExtensionItem[]`, defaults it to `[]`, and uses it to build the `Needs your sign-in` / `Needs admin setup` / `Ready to use` groups for marketplace plugins. **`settings-route.tsx` never passed it.** Those rows were unreachable in every build.

Now passes the already-derived `extensionItems.cloudPluginItems` — no new fetch, no duplicated derivation. This is the direct answer to "does a teammate installing this skill have to connect the MCP themselves?"

## Fix 2 — `ExtensionsView` ignored its own section routing
It accepted `initialSection` and `setSectionRoute` and used **neither**, so `/settings/extensions`, `/extensions/mcp`, `/extensions/skills` and `/extensions/plugins` all rendered an identical page — including the composer's own `Configure` deep link. The inventory now opens pre-filtered and keeps the URL in sync; `/plugins` opens the plugins block expanded.

## Fix 3 — "View in Marketplace" led nowhere
The notification label promised a Marketplace screen that does not exist while navigating to Extensions. Label now matches the real destination, via i18n rather than a hardcoded string.

## Verification
```
pnpm typecheck                     clean
77 focused tests across 24 connect/extensions/mcp suites   all pass
```
New regression test proves Fix 1: routed cloud-marketplace items produce plugin readiness rows. New route test covers the extensions deep links.

## Note for the reviewer
Trivially conflicts with #3113 in `mcp-view.tsx` — two adjacent additions, both wanted. Verified resolved and green in an integration branch merging all of this work together.

### DIFF
--- a/apps/app/src/i18n/locales/en.ts
+++ b/apps/app/src/i18n/locales/en.ts
@@ -629,6 +629,7 @@ export default {
   "extensions.update_available": "Update available",
   "extensions.update_button": "Update",
   "extensions.updating": "Updating...",
+  "extensions.view_in_extensions": "View in Extensions",
   "mcp.disable_app": "Disable",
   "mcp.enable_app": "Enable",
   "mcp.reloading_status": "Reloading MCP servers…",
--- a/apps/app/src/react-app/domains/settings/pages/connect-view.tsx
+++ b/apps/app/src/react-app/domains/settings/pages/connect-view.tsx
@@ -596,7 +596,7 @@ function ConnectSignInPanel(props: ConnectViewProps) {
   );
 }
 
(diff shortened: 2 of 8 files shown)


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

PR #3112, opened 2026-07-25
fix(composer): make capability origin, status, and triggers legible

benjaminshafii:
## Problem
Three defects in how the composer's capability panel (the `Plug` icon — "Commands, skills, and MCPs") describes what a member has.

## Fix 1 — status with no way out
MCP rows rendered `Sign in needed` / `Offline` / `Issue` as plain non-interactive `<div>`s. A member saw the problem and had **no next step**. Actionable rows are now buttons that route to where it can be fixed, reusing the panel's existing `onOpenSettingsSection` callback. `Ready` rows stay inert — nothing to do.

## Fix 2 — the origin badge was backwards
Personal skills got a `Local` pill; **skills that came from the organization got no badge at all** — exactly backwards from what a member needs to know. Organization skills now carry a badge naming the marketplace they came from, so origin is legible for both cases without adding a row.

## Fix 3 — `trigger` was meaningless for organization skills
Local skills derive `trigger` from frontmatter `trigger:`/`when:` or the first bullet under `## When to use`. Organization skills derived it from the **folder name**, so the `Trigger` card in the detail modal showed noise for precisely the skills an enterprise member consumes.

Both paths now share one trigger rule (`@openwork/types/skill-markdown`), and the `Trigger` card hides when no trigger can be derived rather than showing something misleading.

## Reviewer note on the shared helper
First attempt put a full YAML parser in `packages/types` to share the rule — which would have pulled a parser into the **renderer bundle** for one string field (`apps/app` has no `yaml` dependency today). Caught in review and reworked: the shared helper is now **dependency-free**, reads only `trigger`/`when` scalars, ignores indented keys and block scalars, and yields *no trigger* rather than a wrong one when it cannot read confidently. The server keeps using its own real YAML `parseFrontmatter` and passes the parsed data in. `pnpm-lock.yaml` diff is empty.

## Verification
```
pnpm typecheck                        clean
pnpm --filter @openwork/app test      427 pass, 0 fail
pnpm --filter openwork-server test    520 pass, 10 skip, 0 fail
```
New tests cover the trigger reader (quoted values, `when:`, ignored nested keys, unreadable block scalar → undefined, `## When to use` fallback) and the origin badge for both cases.


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

benjaminshafii:

PR #3113, opened 2026-07-25
fix(app): enforce allowManageExtensions and allowControlSettings

benjaminshafii:
## Problem
Two desktop policies are presented to organization administrators in the Cloud dashboard and are **completely unenforced** by the desktop app. An admin can toggle them and nothing happens.

Verified: `allowManageExtensions` and `allowControlSettings` are declared in `packages/types/src/den/desktop-policies.ts` (both `defaultValue: true`, both with `userNotice` copy already written) and had **zero reads** anywhere in `apps/app/src`. Only `allowBuiltInExtensions` was enforced.

This is a trust and compliance defect: we tell administrators they have a control they do not have. It came up directly in an enterprise review, where the ask was "we don't want to expose everything to them" — and we said the policy existed.

## Change
Enforce both, reusing the **existing** `allowBuiltInExtensions` idiom (`useDesktopRestriction`, explanatory banner, per-card `disabledReason`) and each policy's own `userNotice` string rather than new copy.

**Gated:** Add Custom App · From GitHub · quick-connect install · local extension config slots · MCP Remove/Enable/Disable · skill Uninstall · plugin Add/Remove · plugin package input · sidebar Settings item · command-palette Settings/Extensions/Appearance/Updates · typed `/settings/*` routes.

**Deliberately left open** — this is the important half:
- MCP OAuth **sign-in/sign-out** (that is authentication, not local extension management)
- Signing in to an **organization-assigned** connection (members must still use what the org gave them)
- Cloud Account, AI Providers, Cloud Providers, Connect, Recovery routes — needed for sign-in/out, org switching, provider auth, and recovery

An over-broad implementation here would trap a user in an unusable app with settings locked, which is worse than the bug being fixed. Blocked routes render a notice with escape paths instead of dead-ending.

## Both policies default to `true`
**Default behavior is unchanged.** An organization with no policy data sees zero difference — covered by test.

## Verification
```
pnpm typecheck                       clean
pnpm --filter @openwork/app test     428 pass, 0 fail
```
New `apps/app/tests/desktop-policy-gates.test.ts` proves, per policy: absent policy leaves the affordance enabled; `false` disables it and surfaces the `userNotice`; and the non-blockable routes stay reachable.

## Note for the reviewer
Trivially conflicts with #3111 in `mcp-view.tsx` — two adjacent additions, both wanted. Verified resolved and green in an integration branch merging all of this work together.


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

benjaminshafii:

benjaminshafii:
Merged `dev` to resolve the `mcp-view.tsx` conflict flagged above (it appeared once #3111 landed). Both sides were wanted and kept: #3111's `setInventoryFilter`/`initialFilter` plumbing alongside this PR's `manageExtensionsDisabledReason`. Re-verified after the merge: `pnpm typecheck` clean, `pnpm --filter @openwork/app test` 430 pass / 0 fail.

PR #3114, opened 2026-07-25
feat(den-web): preview organization prompt cards while authoring them

benjaminshafii:
## Problem
Administrators author the organization prompt cards that members see on a new session **completely blind**. There is no preview, and the field mapping is genuinely counter-intuitive:

- the admin's field labelled **"Description"** becomes the member's card **title**
- the **"Prompt"** text becomes the card **body** and the composer text

The only way to check your work today is to save and ask a member to reload. An admin about to author department-specific prompt sets is guessing.

## Change
Dashboard-only. In `desktop-policy-editor-screen.tsx`:

1. **Live preview** of the member's cards as the admin types — same title/body mapping, same Sparkles treatment, same "Try one of your organization's prompts:" header, and the same real fallback (`Organization prompt 1|2|3`) when a description is blank.
2. **Fields relabelled for what they become**, so the inversion stops being a trap. Existing char counters kept.
3. **States that clicking does not send** — admins assume the card auto-runs; it fills the composer.
4. **Surfaces the winner-takes-all priority rule**, which differs from how the boolean policies OR-union and is a real footgun.

Follows the existing preview precedent in `brand-appearance-screen.tsx` (same side-by-side grid, bordered panel, uppercase "Preview" label) and uses the existing `Den*` primitives. No new component library.

## Constraints honored
- **No schema, API, or desktop rendering change** — nothing any member currently sees is altered. `git diff --stat origin/dev...HEAD -- apps/` is **empty**.
- No import from `apps/app` into `ee/apps/den-web`; the card presentation is reimplemented locally with a comment pointing at `task-suggestions.tsx` as the source of truth so drift is detectable.
- Used an inline SVG matching the Sparkles path because den-web does not depend on `@heroicons/react`.

## Verification
```
pnpm typecheck                              clean
pnpm --filter @openwork-ee/den-web build     succeeded
pnpm --filter @openwork-ee/den-web test      73 pass, 0 fail
```
New unit test covers the resolver: description becomes title, blank description falls back to `Organization prompt N`, prompt becomes body.

### DIFF
--- a/ee/apps/den-web/app/(den)/dashboard/_components/desktop-policy-editor-screen.tsx
+++ b/ee/apps/den-web/app/(den)/dashboard/_components/desktop-policy-editor-screen.tsx
@@ -23,6 +23,7 @@ import {
   type DenDesktopPolicy,
   type DesktopPolicyPayload,
 } from "./desktop-policy-data";
+import { resolveOrganizationPromptPreviewCardContent } from "./desktop-policy-onboarding-preview";
 import { EnterprisePlanNotice } from "./enterprise-plan-notice";
 
 type PolicyDraft = {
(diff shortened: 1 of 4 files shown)


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

PR #3115, opened 2026-07-25
feat(desktop-policies): let organization prompts launch a named skill

benjaminshafii:
> **Stacked on #3114** (`feat/desktop-policy-prompt-preview`). Review that first; this extends the preview it adds.

## The problem
We tell administrators that the organization prompt cards are how members discover what the platform can do. That promise is **unbacked**.

A prompt is free text only. There is no `skillId`, `pluginId`, or `capabilityName` on the schema, and `normalizeDesktopPolicyDocument` strips unknown keys. So a card reading "Find what needs my attention" just drops that sentence into the composer and *hopes* the agent selects the right skill.

Meanwhile a binding already works **by accident**: `editor.tsx` `setPrompt()` parses a leading `/slug` or `[skill x]` into a pill, unvalidated and never mentioned in the Cloud UI.

## The change: formalize the accident
A prompt card can now name the skill it launches, and tells the member the truth about readiness **before** they click.

1. **Schema** — prompts accept an optional bound skill reference (local by slug, or organization skill with the marketplace + capability identifiers needed to actually resolve it). **No DB migration**: prompts ride in the existing `policy` JSON column.
2. **Desktop** — a bound card invokes its skill through the **existing** token/expansion mechanism rather than a second invocation path; the guarantee becomes explicit instead of accidental. Applied to **both** render surfaces (`task-suggestions.tsx` and `session-empty-hero.tsx`) so they cannot diverge.
3. **Readiness precheck** — if the bound skill needs a connection the member has not made, the card says so up front instead of failing mid-run, reusing the existing `needs_signin` / `needs_admin_setup` / `ready` vocabulary and copy. Needs-sign-in offers the route to fix it; needs-admin-setup says so plainly and does **not** send a member somewhere they have no permission.
4. **Cloud admin** — the admin picks from skills that actually exist in the organization's marketplaces (reusing the dashboard's existing marketplace data, no raw identifier typing), and the live preview shows the bound skill.

## Compatibility is the risk, so it is tested in both directions
- Existing policies are `string[]` and keep working **unchanged**; cards with no bound skill behave exactly as today (fill the composer, do not auto-send).
- An **old desktop** reading a **new** policy degrades safely; a **new desktop** reading an **old** policy behaves as today.

## Why this shape
Per-team and per-person scoping already ships and is tested (`desktop_policy_member.team_id` / `org_member_id`), so this adds a payload field to working infrastructure rather than building a system. It is also the substrate for scheduled work later: a bound, readiness-checked prompt is exactly the thing you can put on a schedule — without building a dashboard.

## Verification
```
pnpm --filter @openwork/app test                      all pass
pnpm --filter @openwork-ee/den-web build               succeeded
den-api desktop-policy suites (real DB)                9 pass, 0 fail
```
OpenAPI snapshot regenerated via the generator, not hand-edited. New tests cover legacy-string resolution, bound resolution, old-shape-reads-new-shape, each readiness state mapping to the right affordance, and unbound cards still only filling the composer.


benjaminshafii:

|---|---|---|

```
```

```
```

### Honest gaps

benjaminshafii:

PR #3125, opened 2026-07-25
feat(cloud): OpenWork Cloud alpha — a full instance in the browser, no orchestrator

benjaminshafii:
## What this is

**OpenWork Cloud** — an alpha, per-organization capability. When an OpenWork admin enables it for an org, members get a **Cloud** entry in the dashboard that opens a full OpenWork instance in the browser: a Daytona sandbox running `openwork-server` plus the web UI, from our own snapshot. Nothing to install.

It also removes `openwork-orchestrator` from that path entirely.

Spec is the approved voice-over at `evals/voiceovers/cloud-instance.md` (8 frames). This PR implements frames 1-6.

## Two decisions worth reviewing

**1. No credential injection.** The sandbox never receives org credentials. `openwork-server` already has cloud-MCP machinery (`routes/cloud-mcp.ts`, `cloud-mcp-health.ts`), so org connections are brokered server-side by Den and the instance needs only a scoped token. The browser receives **no token at all** — the instance injects its own collaborator-scoped token into the `index.html` it serves. The host token is never injected.

**2. The browser gets the Daytona signed preview URL, not `workerProxyUrl()`.** I verified this empirically against a real sandbox (`public = false`):

| Request | Result |
|---|---|
| `/` top-level nav | 200 `text/html` |
| `/assets/app.js` sub-resource | 200 `text/javascript` |
| `/api/data.json` XHR | 200 `application/json` |
| wrong sandbox id | **401** |
| wrong port | **401** |

So sub-resources work with no cookie/header/shim, and the URL itself is the credential. `den-worker-proxy` cannot serve a browser app — it 401s any request without a token (a `<script src>` has none) and its `/{workerId}` prefix breaks the SPA's `base: "/"`. `provisionWorkerOnDaytona`'s return value is unchanged because `worker_instance.url` is consumed by the Telegram path.

## Self-hosted is untouched

This was a hard requirement and it is enforced structurally:

- **No new required env var**, and **nothing added** to the `superRefine` in `ee/apps/den-api/src/env.ts:173-183`.
- **No DB migration** — the capability rides the existing `organization.metadata` JSON.
- The `multi_org` gate lives in **exactly one helper**, `organizationCloudEnabled()` in `ee/apps/den-api/src/capability-sources/cloud-rollout.ts` — so offering Cloud to other deployments later is a one-line change.
- `startEmbeddedServer` is **untouched**, so the desktop cannot regress. The heartbeat is started only from `apps/server/src/cli.ts`.
- Static UI serving is a **complete no-op** when `OPENWORK_WEB_ROOT` is unset, and the heartbeat stays disabled unless all five `DEN_*` conditions hold.
- When Cloud is off it returns **404, not 403**, so the feature never advertises itself.

## Tests run

| Command | Result |
|---|---|
| `pnpm --filter openwork-server exec bun test` | **540 pass / 0 fail** (550 across 77 files) |
| `pnpm --filter @openwork-ee/den-api exec bun test test/cloud-instance-route.test.ts test/cloud-rollout.test.ts test/organization-capabilities.test.ts test/admin-organization-capabilities.test.ts test/route-access-policy.test.ts` | **18 pass / 0 fail** |
| `pnpm --filter openwork-server exec tsc --noEmit` | PASS |
| `pnpm --filter @openwork-ee/den-api exec tsc --noEmit` | PASS |
| `pnpm --filter @openwork-ee/den-web exec tsc --noEmit` | PASS |
| `pnpm --filter @openwork/app exec tsc --noEmit` | PASS |

New coverage includes the gate returning 404 when the capability is off **and** in `single_org` even with a literal `cloud: true`, plus static-UI traversal rejection, MIME/cache headers, SPA fallback, bootstrap escaping, and the heartbeat config gate.

Live `curl` proof of the server serving the SPA (captured locally):

```
GET /                        -> 200 text/html   (body contains __OPENWORK_BOOTSTRAP__ + token)
GET /assets/<hashed>.js      -> 200 + Cache-Control: public, max-age=31536000, immutable
GET /settings/general        -> 200 text/html   (SPA history fallback)
GET /assets/app-DEADBEEF.js  -> 404 JSON        (missing hashed asset is a hard 404, not HTML)
GET /health                  -> 200 JSON        (API unaffected)
POST /not-a-route            -> 404 JSON
GET / without OPENWORK_WEB_ROOT -> 404 JSON     (proves the no-op)
encoded ../ traversal        -> 400 path_escape (no leak)
```

## NOT verified — please read

Per the repo's validation standard I am **not** claiming this passes.

1. **No fraimz.** `evals/flows/cloud-instance.flow.ts` is scaffolded but its frames are still `ctx.assert(false, ...)` stubs. No `fraimz.html` exists, so nothing here is backed by frame-by-frame proof yet. Frames 1-3 are provable today; frames 4-6 need the two items below.
2. **The snapshot image was never built.** `docker build` was aborted locally (the machine hit 100% disk). So `Dockerfile.daytona-snapshot` is **unproven end to end** — in particular whether `better-sqlite3` (imported by `apps/server/src/opencode-db.ts`) loads inside the `bun --compile` binary. The image asserts `openwork-server --version && opencode --version`, which is where that would surface.
3. **Frames 7-8 (sleep/wake) are not implemented.** `deprovisionWorkerOnDaytona` deletes the sandbox *and* erases `workers/<id>/`; there is still no stop-without-cleanup or wake path. Tracked as follow-up.

## Follow-ups

- **JIT lifecycle**: stop-without-cleanup, wake, and signed-preview-URL re-resolution (frames 7-8).
- **Delete Render worker provisioning.** `ee/apps/den-api/src/workers/provisioner.ts` still runs `npm install -g openwork-orchestrator` at an **unpinned** version (prod `RENDER_WORKER_OPENWORK_VERSION` is unset, so worker builds float to npm `latest`) off branch `dev`. Once Cloud replaces it, that path and the orchestrator sidecar can go, and the npm package can be deprecated.
- Before Cloud leaves alpha, replace the inline bootstrap token with a Den-authenticated one-time-code handoff so the token never sits in static HTML.

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -8,7 +8,7 @@
     "dev": "OPENWORK_DEV_MODE=1 vite",
     "build": "vite build",
     "dev:web": "OPENWORK_DEV_MODE=1 vite",
-    "build:web": "VITE_OPENWORK_DEPLOYMENT=web vite build",
+    "build:web": "VITE_OPENWORK_DEPLOYMENT=web VITE_DEN_REQUIRE_SIGNIN=1 vite build",
     "preview": "vite preview",
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "test:health": "node scripts/health.mjs",
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -8,7 +8,7 @@
     "dev": "OPENWORK_DEV_MODE=1 vite",
     "build": "vite build",
     "dev:web": "OPENWORK_DEV_MODE=1 vite",
-    "build:web": "vite build",
+    "build:web": "VITE_OPENWORK_DEPLOYMENT=web vite build",
     "preview": "vite preview",
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "test:health": "node scripts/health.mjs",
--- a/apps/app/src/app/constants.test.ts
+++ b/apps/app/src/app/constants.test.ts
@@ -1,0 +1,34 @@
+declare const describe: (name: string, fn: () => void) => void;
+declare const test: (name: string, fn: () => void) => void;
+declare const expect: (value: unknown) => {
+  toEqual: (expected: unknown) => void;
+};
+
+import {
+  OPENWORK_EXTENSION_CATALOG,
+  filterOpenWorkExtensionCatalogForPlatform,
+  resolveOpenWorkExtensionCatalogPlatform,
+} from "./constants";
+
+function filteredIds(platform: "darwin" | "linux" | "windows" | "web") {
+  return filterOpenWorkExtensionCatalogForPlatform(OPENWORK_EXTENSION_CATALOG, platform)
+    .flatMap((entry) => entry.id ? [entry.id] : []);
+}
+
+describe("OpenWork extension catalog platform filter", () => {
+  test("resolves browser runtime to web and desktop runtime to OS", () => {
+    expect(resolveOpenWorkExtensionCatalogPlatform("web", "macos")).toEqual("web");
+    expect(resolveOpenWorkExtensionCatalogPlatform("desktop", "macos")).toEqual("darwin");
+    expect(resolveOpenWorkExtensionCatalogPlatform("desktop", "windows")).toEqual("windows");
+    expect(resolveOpenWorkExtensionCatalogPlatform("desktop", "linux")).toEqual("linux");
+  });
+
+  test("hides desktop-only extensions in web", () => {
+    expect(filteredIds("web")).toEqual(["openwork-voice", "google-workspace", "ollama"]);
+  });
+
+  test("keeps OpenWork Browser desktop-only and Computer Use mac-only", () => {
+    expect(filteredIds("darwin")).toEqual(["openwork-browser", "computer-use", "openwork-voice", "google-workspace", "ollama"]);
+    expect(filteredIds("linux")).toEqual(["openwork-browser", "openwork-voice", "google-workspace", "ollama"]);
+  });
+});
--- a/apps/app/src/app/constants.ts
+++ b/apps/app/src/app/constants.ts
@@ -7,6 +7,7 @@ import {
   extensionResource,
   isTrustedBuiltInExtension,
   type OpenWorkExtensionManifest,
+  type OpenWorkExtensionPlatform,
 } from "./extensions";
 
 export const MODEL_PREF_KEY = "openwork.defaultModel";
@@ -193,3 +194,23 @@ export const MCP_QUICK_CONNECT: McpDirectoryInfo[] = [
 ];
 
 export const OPENWORK_EXTENSION_CATALOG = MCP_QUICK_CONNECT.filter((entry) => entry.kind === "extension");
+
+export function resolveOpenWorkExtensionCatalogPlatform(
+  platform: "web" | "desktop",
+  os?: "macos" | "windows" | "linux",
+): OpenWorkExtensionPlatform {
+  if (platform === "web") return "web";
+  if (os === "macos") return "darwin";
+  if (os === "windows") return "windows";
+  return "linux";
+}
+
+export function filterOpenWorkExtensionCatalogForPlatform<TEntry extends Pick<McpDirectoryInfo, "extensionManifest">>(
+  entries: TEntry[],
+  platform: OpenWorkExtensionPlatform,
+): TEntry[] {
+  return entries.filter((entry) => {
+    const platforms = entry.extensionManifest?.platform;
+    return !platforms || platforms.includes(platform);
+  });
+}
--- a/apps/app/src/app/extensions.ts
+++ b/apps/app/src/app/extensions.ts
@@ -131,6 +131,8 @@ export type OpenWorkExtensionManifest = {
   platform?: Array<"darwin" | "linux" | "windows" | "web">;
 };
 
+export type OpenWorkExtensionPlatform = NonNullable<OpenWorkExtensionManifest["platform"]>[number];
+
 export function extensionContribution(
   manifest: OpenWorkExtensionManifest | undefined,
   type: OpenWorkExtensionContributionType,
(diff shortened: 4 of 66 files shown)


benjaminshafii:
## Full lifecycle E2E on real Daytona — passed

Ran against a **real Daytona sandbox**, provisioned **through den-api** (`PROVISIONER_MODE=daytona`, real `DAYTONA_API_KEY` from Infisical dev), from the published snapshot **`openwork-cloud-alpha-test`** (built from this branch's Dockerfile, 183MB). Idle knobs tightened for the test: `CLOUD_IDLE_STOP_MINUTES=2`, `CLOUD_IDLE_LOOP_SECONDS=20`.

| Step | Result |
|---|---|
| Sign in (seeded org, `cloud` capability on) → session bearer | ✅ |
| `GET /v1/cloud/instance` → real provision from snapshot | ✅ `provisioning` → **ready in ~10–28s** |
| Instance serves UI over signed preview URL | ✅ `/health` 200 · `/` 200 with `__OPENWORK_BOOTSTRAP__` |
| **Idle loop stops the sandbox unattended** | ✅ `healthy → stopped` (observed repeatedly); Daytona confirms `state: stopped`; old preview URL goes 400 |
| `GET /v1/cloud/instance` on a stopped worker | ✅ responds `waking` immediately, single-flight wake kicked |
| **Wake wall time** | ✅ **4–7s** across three wakes (budget: 20s) |
| File written via instance API | ✅ through the **manual-approvals flow**: write blocks pending → `GET /approvals` → host-token `allow` → 200 |
| **File intact after a full sleep→wake cycle** | ✅ `E2E_PERSISTENCE_PROOF_2026` read back |

### Findings worth knowing

1. **`--approval manual` blocks direct REST file writes in Cloud** — by design, but in Cloud there is no host UI attached to approve. The agent path (opencode) is unaffected; direct collaborator writes wait for approval. Needs a product decision before GA (auto-approve collaborator writes in Cloud, or surface approvals in the web UI).
2. Daytona's preview proxy sets a `daytona-sandbox-auth-<id>` session cookie on first response — that's how sub-resource auth works on the signed URL.
3. Cross-building the snapshot on Apple Silicon requires the native-builder pattern now in the Dockerfile (`FROM --platform=$BUILDPLATFORM`); QEMU kills both node (libuv assert) and bun (illegal instruction). Runtime version asserts stay on in CI, skippable locally via `RUNTIME_ASSERTS=0`.

### Honest gaps

- No `fraimz.html` yet — this was a driven E2E, not the coded flow; `evals/flows/cloud-instance.flow.ts` frames still need coding (voiceover frames 7–8 are now *true* but unproven by the runner).
- The heartbeat couldn't reach the local den-api from the sandbox (laptop not internet-reachable), so idle fell back to `updated_at` — exactly the designed fallback. In production the heartbeat URL is public and `last_active_at` drives it.
- Volume-subpath persistence wasn't independently isolated (sandbox-disk persistence is what the current architecture uses and is what's proven).

Repro: seeded DB `openwork_den_e2e` on the local MySQL container, `alex@acme.test / OpenWorkDemo123!`, snapshot `openwork-cloud-alpha-test` published in the Daytona org.

benjaminshafii:
## Coded eval: web sign-in → Cloud instance → approved action → mock MCP through Connect — PASSED

`pnpm evals --flow cloud-web-connect-e2e` — all 7 frames green against a **real Daytona sandbox** provisioned through den-api (daytona mode, real dev key), driving a real headless Chrome. Flow: `evals/flows/cloud-web-connect-e2e.flow.mjs` (`f59effd8b`).

| Frame | Claim | Evidence |
|---|---|---|
| 1 | No app without sign-in | screenshot: sign-in gate only |
| 2 | Admin signs in | screenshot: org chooser with Acme Robotics |
| 3 | Machine assigned **only after** sign-in | `ready` w/ bearer; **401 without** (machine-checked) |
| 4 | The instance is the full OpenWork app | screenshot: app shell served by the sandbox |
| 5 | Admin completes an action | file write through the **approvals flow** (pending → host-token allow → 200 → read-back) |
| 6 | Org mock MCP visible in Connect, in the instance | screenshot: "Connected to Acme Robotics" + **Acme Mock Tools — Ready** |
| 7 | Mock MCP **provably executes** through OpenWork Connect | `execute_capability` via `/mcp/agent` ran `acme_lookup`; the mock server's own request log names the tool; result text `acme lookup ok: record-7` |

Artifacts: `evals/results/2026-07-25T21-59-50-458Z/` (fraimz.html, report.md/json, 4 validated screenshots).

### Product findings from getting this green

1. **Bug: `ready` with a dead sandbox.** Den trusts `signed_preview_url_expires_at` without verifying the sandbox exists — after out-of-band deletion, `/v1/cloud/instance` returned `ready` + a 404 URL. Needs a liveness check or wake-on-404.
2. **Dynamic preview origins break static CORS.** Instance origins rotate on every re-sign; a fixed `CORS_ORIGINS` list can't track them. Evidence for the Phase-2 gateway owning browser↔Den traffic.
3. **Papercut: org-connections list fetches once on mount** and renders a stale "Failed to fetch" with no retry; the availability banner's Refresh doesn't refetch it.
4. **Expected + documented:** the in-instance "Agent access" card reads Degraded because the sandbox engine cannot reach a loopback Den — resolves under any public Den (prod) / the gateway.

### Eval-topology accommodations (stated honestly)

Local-only, not product behavior: den-api fronted by an `/api/den`-prefix proxy (den-web's shape, per `den-stack.ts`) with Origin-reflecting CORS for the rotating sandbox origin; the eval Chrome runs `--disable-web-security --disable-quic` because a **public-https page fetching loopback Den** trips Chrome's Local Network Access + a local DNS filter blocked `*.trycloudflare.com` (browser policy isn't under test — auth/approvals asserted server-side). In production Den is public https and none of this applies.

⚠️ Ops note: the dev `DAYTONA_API_KEY` was accidentally echoed into a local session transcript today — **rotate it**.

PR #3131, opened 2026-07-25
docs(enterprise): document outbound network access and guard it in CI

benjaminshafii:
## Why

Enterprise customers run OpenWork on networks where all outbound connections are blocked by default and IT approves each destination individually. Nothing in this repo listed what OpenWork needs to reach — so every deployment rediscovered it by hitting failures.

Two destinations were named **nowhere in the repo**, and both produce confusing symptoms:

| Host | Why it is needed | Symptom when only the obvious host is approved |
|---|---|---|
| `objects.githubusercontent.com` | Redirect target of every GitHub release-asset download (`redirect: "follow"`) | IT approves `github.com`; the download **starts and then dies partway through** |
| `registry.npmjs.org` | `npx -y openwork-ui-mcp` on the packaged desktop path | UI-control MCP is **silently unavailable** while the app otherwise opens normally |

## What this adds

**`docs/enterprise/outbound-access.md`** — written for customer IT and non-technical admins. Deliberately avoids the word "egress". Leads with a copy-pasteable minimum allowlist, then tables with purpose / required-or-optional / **what breaks when blocked**.

Desktop client access is documented **separately** from self-hosted Den access, because different teams approve them on different networks. It also records that self-hosted Den replaces the hosted origins and that the desktop needs only **one** origin (API and MCP are derived as `<baseUrl>/api/den/...`), and that Den's only hard requirements are its MySQL endpoint plus `ghcr.io` at image-pull time — everything else is per-feature.

**`docs/enterprise/outbound-access.json`** — machine-readable, 34 hosts:

| kind | count | | requirement | count |
|---|---|---|---|---|
| `fetched` | 23 | | `opt-in` | 16 |
| `link-only` | 7 | | `optional` | 11 |
| `subprocess` | 2 | | `required-in-practice` | 4 |
| `redirect-target` | 1 | | `required-for-cloud` | 2 |
| `schema-string` | 1 | | `required` | 1 |

The `kind` field is what lets indirectly-reached destinations be documented at all — a redirect target and an `npx` subprocess have no URL literal to find in the code.

**`scripts/check-outbound-access.mjs`** + a `ci-tests.yml` step — scans the shipped client surfaces (`apps/desktop/electron`, `apps/installer/src`, `apps/server/src`, `apps/app/src`, `packages/openwork-bootstrap/bin`) and fails when a hostname appears in code but not in the manifest. So a feature PR that introduces a new destination must document its enterprise impact. It also flags `fetched` entries whose host no longer exists in code, so the list cannot rot in either direction. RFC 2606 documentation names, loopback, and standards/schema hosts are ignored with a documented reason in the script.

## Classifications corrected while verifying

Every row was checked against a real call site rather than assumed. Corrections:

- `api.githubcopilot.com` — not a provider call path; an example custom MCP URL placeholder (`apps/app/src/i18n/locales/en.ts:803`).
- `ollama.com` — link-only help text (`ollama-config.tsx:312`); actual Ollama traffic is loopback `http://localhost:11434/v1`.
- `code.claude.com` — source comment only (`apps/server/src/claude-plugin-bundle.ts:11`), no runtime call.
- `linear.app` — demo transcript sample link (`session-surface.tsx:168`), not a fetch.
- `mail.google.com` — link-only Gmail draft/thread URLs emitted as output (`google-workspace.ts:925,929`).
- `schemas.agentskills.io` — schema identifier string, never fetched (`connect-skill-catalog.ts:11,15`).
- `registry.npmjs.org` — also covers the dev-only `npx -y @openwork/handsfree mcp` fallback (`computer-use.mjs:50`).
- Added `ghcr.io` for the self-hosted image/chart pull requirement (`values.yaml:184,212,241`).

## Compatibility

**No breaking changes for self-hosted deployments.** No product code, configuration, environment variable, API, Helm chart, or database surface is touched. This is documentation, one check script, one CI step, one `package.json` script.

One operational note for maintainers: because the check now runs in `ci-tests.yml`, an open PR that introduces a new outbound destination will fail until it adds a manifest line. That is the intended forcing function.

## Tests run

\`\`\`
node scripts/check-outbound-access.mjs
# Outbound access manifest covers 34 hosts. Scanned 31 external hosts from
# shipped client surfaces. No stale fetched entries.

pnpm check:outbound-access          # pass
node scripts/check-outbound-access.mjs --json   # valid JSON, ok: true
\`\`\`

**Negative controls**, run in three different files so the scan surface is genuinely covered:

Injected `https://collector.acme-telemetry.io/v1` into `apps/app/src/app/lib/desktop.ts`:
\`\`\`
Outbound access manifest check failed.
- Host collector.acme-telemetry.io is used in apps/app/src/app/lib/desktop.ts:601
  but is missing from docs/enterprise/outbound-access.json. Add it to
  docs/enterprise/outbound-access.json (see docs/enterprise/outbound-access.md).
exit=1
\`\`\`

Injected a fake host into `apps/installer/src/release-asset.ts` → caught. Deleted the `models.openworklabs.com` manifest entry → caught at `apps/server/src/opencode-models-url.ts:2`. Added a fake `kind: "fetched"` entry with no code literal → correctly reported as stale:
\`\`\`
- Manifest host stale-host.openworklabs.com has kind "fetched" but no
  https://stale-host.openworklabs.com literal was found in the scan roots.
\`\`\`

## Fraimz

Skipped, stated explicitly per AGENTS.md: documentation plus a CI check script, no runtime path changed. The executable proof is the four controls above.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -129,6 +129,9 @@ jobs:
       - name: Run eval runner unit tests
         run: pnpm test:eval-runner
 
+      - name: Check outbound access manifest
+        run: pnpm check:outbound-access
+
       - name: Typecheck Electron main against the IPC contract
         run: pnpm --filter @openwork/desktop typecheck:electron
 
--- a/docs/enterprise/outbound-access.json
+++ b/docs/enterprise/outbound-access.json
@@ -208,12 +208,21 @@
       "override": "OPENWORK_GOOGLE_WORKSPACE_TOKEN_BROKER_URL for token exchange/refresh only"
     },
     {
+      "host": "release-assets.githubusercontent.com",
+      "kind": "redirect-target",
+      "components": ["desktop", "installer"],
+      "purpose": "Current GitHub release-asset download target reached after github.com redirects installer and updater downloads. Verified live against a real OpenWork release asset.",
+      "requirement": "required-in-practice",
+      "blockedEffect": "OpenWork downloads start from github.com and then fail partway through when the redirect target is blocked.",
+      "override": null
+    },
+    {
       "host": "objects.githubusercontent.com",
       "kind": "redirect-target",
       "components": ["desktop", "installer"],
-      "purpose": "GitHub release-asset download target reached after github.com redirects installer and updater downloads.",
+      "purpose": "Legacy GitHub release-asset download target. GitHub has migrated to release-assets.githubusercontent.com, but allow both so older clients and any staged rollback keep working.",
       "requirement": "required-in-practice",
-      "blockedEffect": "OpenWork downloads start from github.com and then fail partway through when the redirect target is blocked.",
+      "blockedEffect": "OpenWork downloads start from github.com and then fail partway through if GitHub serves this legacy redirect target and it is blocked.",
       "override": null
     },
     {
--- a/docs/enterprise/outbound-access.json
+++ b/docs/enterprise/outbound-access.json
@@ -1,0 +1,310 @@
+{
+  "hosts": [
+    {
+      "host": "accounts.google.com",
+      "kind": "fetched",
+      "components": ["server"],
+      "purpose": "Google Workspace OAuth authorization page opened when a user connects Google Workspace.",
+      "requirement": "opt-in",
+      "blockedEffect": "Users cannot complete Google Workspace sign-in from OpenWork.",
+      "override": null
+    },
+    {
+      "host": "api.github.com",
+      "kind": "fetched",
+      "components": ["server"],
+      "purpose": "GitHub API calls for importing and previewing GitHub-hosted Claude Code/OpenWork plugins.",
+      "requirement": "opt-in",
+      "blockedEffect": "GitHub plugin import cannot inspect repositories, branches, or plugin manifests.",
+      "override": "OPENWORK_GITHUB_API_BASE"
+    },
+    {
+      "host": "api.githubcopilot.com",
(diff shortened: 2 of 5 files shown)


benjaminshafii:
## Correction pushed: the GitHub release-asset host has changed

While writing the companion private-network deployment guide I verified the redirect chain against a **live** OpenWork release asset instead of trusting the repo's own prose. The current target is **not** `objects.githubusercontent.com`:

```
$ curl -sIL -o /dev/null -w '%{url_effective}' \
    https://github.com/different-ai/openwork/releases/latest/download/latest-mac.yml

https://release-assets.githubusercontent.com/github-production-release-asset/1133911335/...
```

Full chain:
1. `github.com/.../releases/latest/download/latest-mac.yml` → 302
2. `github.com/.../releases/download/v0.18.1/latest-mac.yml` → 302
3. **`release-assets.githubusercontent.com/...`**

GitHub migrated the release-asset CDN. This PR originally named only `objects.githubusercontent.com`, so it would have told customer IT to open the wrong host — producing exactly the start-then-die download failure the page exists to prevent. That would have been worse than no doc at all.

Both are now listed: `release-assets.githubusercontent.com` as the host GitHub serves today, and `objects.githubusercontent.com` retained for older clients and any staged rollback. Manifest is now 35 hosts and the check still passes.

Worth noting for reviewers: the only reason this surfaced is that the repo describes this destination generically as *"GitHub's redirected release-asset host"* and never names it, so no code or doc anywhere pinned the value and nothing broke visibly when it changed.

PR #3136, opened 2026-07-25
test(evals): pin the semi air-gapped Den deployment contract

benjaminshafii:
## Why

Full air-gap is rare. **The common enterprise shape is a split:** laptops have internet *and* VPN, while Den web and Den API run inside the customer's private network.

That shape silently depends on four guarantees, and **no test held any of them.** A refactor could have changed on-prem behavior — or an enterprise firewall requirement — without anything turning red.

## What this adds

`evals/flows/semi-airgapped-den-contract.flow.mjs` (internal, `requiresApp: false`) plus its voiceover. Four frames, each backed by real evidence rather than prose:

| Frame | Guarantee | Why it matters behind a VPN |
|---|---|---|
| 1 | `resolveDenBaseUrls` ignores an explicitly configured API origin when a base URL is present | Keeps the firewall request to **one hostname**. If this flips, every customer's allowlist silently becomes wrong. |
| 2 | `url-guard` refuses private/reserved MCP addresses by default; `DEN_ALLOW_PRIVATE_MCP_URLS` is the only documented way in | If the default flips open, **hosted multi-tenant Den becomes SSRF-exposed** |
| 3 | The cloud catalog probe returns `untrusted_endpoint` and **performs no request** for an untrusted origin | A self-hosted deployment leaks no organization MCP token, and the diagnostic is *skipped*, not *failed* |
| 4 | `DEN_DIAGNOSTICS_ORIGIN` selects the outbound self-check target, public host as explicit default; bearer token minimum enforced | A VPN-only Den can point its self-check at an internal diagnostics service |

Evidence comes from focused `bun test` filters plus direct imports of the pure functions — following the in-flow test pattern already used by `installer-tls-trust.flow.ts`, and structured on `helm-custom-ca.flow.mjs`.

Values verified rather than assumed: default origin is `https://diagnostic.openworklabs.com`; the bearer token minimum is **24** characters (23 rejected, 24 accepted).

## Runs anywhere

No app, no database, no network. `requiresApp: false`, and it passes when handed a deliberately dead CDP URL:

\`\`\`
$ pnpm evals --flow semi-airgapped-den-contract --cdp-url http://127.0.0.1:1
  ✓ Frame 1  ✓ Frame 2  ✓ Frame 3  ✓ Frame 4
\`\`\`

## It can fail — verified on two different frames

A flow that cannot fail proves nothing, so this was checked in both directions, each time reverting the product source completely afterward.

Breaking frame 1 (make `resolveDenBaseUrls` prefer the explicit API origin):
\`\`\`
  ✗ Frame 1 — resolveDenBaseUrls ignores an explicit API origin when baseUrl is present
Result: FAILED — 0 passed, 1 failed, 0 skipped
\`\`\`

Breaking frame 2 (force `allowPrivateMcpUrls = true`):
\`\`\`
  ✗ Frame 2 — env.allowPrivateMcpUrls is false by default
Result: FAILED — 0 passed, 1 failed, 0 skipped
\`\`\`

`git status` is clean of both temporary edits; the final diff is only the two new eval files.

## Tests run

\`\`\`
pnpm evals  --flow semi-airgapped-den-contract   # PASSED (automation)
pnpm fraimz --flow semi-airgapped-den-contract   # PASSED (demo) + Voice-over script coverage ✓
pnpm evals:typecheck                             # pass
pnpm evals:test                                  # 4 tests, 4 pass, 0 fail
pnpm evals --list | grep semi-airgapped          # discovered
\`\`\`

## Fraimz

**Passed.** `fraimz.html` produced (32,938 bytes), 34 recorded assertions, every claim bound to an observable assertion via `witness(...)` with raw evidence attached through `ctx.output(...)`. Narration coverage passes with no drift against `evals/voiceovers/semi-airgapped-den-contract.md`.

## Compatibility

**No breaking changes.** Eval flow and voiceover only — no product code, configuration, chart, or docs. Nothing ships to users.

## Scope — what this is not

This proves the semi air-gapped contract **at the code level**. It does **not** stand up a VPN-isolated Den with a private CA and drive the real desktop app through the installer → sign-in → Cloud MCP chains. That remains worth doing; the existing cloud MCP flows are already parameterized by `OPENWORK_EVAL_DEN_API_URL` / `OPENWORK_EVAL_DEN_WEB_URL`, so pointing them at an internal Den with a private CA is the natural next step rather than new flows.

## Related

- #3135 documents this deployment shape for operators.
- #3132 and #3133 make the two operator switches asserted in frames 2 and 3 discoverable.

### DIFF
--- a/evals/flows/semi-airgapped-den-contract.flow.mjs
+++ b/evals/flows/semi-airgapped-den-contract.flow.mjs
@@ -1,0 +1,338 @@
+import { spawnSync } from "node:child_process";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { loadVoiceoverParagraphs } from "../runner/voiceover.mjs";
+
+// Narration is loaded from the approved script (evals/voiceovers/semi-airgapped-den-contract.md).
+// The runner fails this flow in demo mode if the narration drifts from that script.
+const FLOW_ID = "semi-airgapped-den-contract";
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), "..", "..");
+const APP_ROOT = join(ROOT, "apps", "app");
+const DEN_API_ROOT = join(ROOT, "ee", "apps", "den-api");
+const SERVER_ROOT = join(ROOT, "apps", "server");
+const COMMAND_TIMEOUT_MS = 120_000;
+const DEFAULT_DIAGNOSTICS_ORIGIN = "https://diagnostic.openworklabs.com";
+const CUSTOMER_DIAGNOSTICS_ORIGIN = "https://diagnostic.customer.example";
+const MIN_DIAGNOSTICS_BEARER_TOKEN_LENGTH = 24;
+const vo = await loadVoiceoverParagraphs(FLOW_ID);
+
+function run(command, args, cwd, env = process.env) {
+  return spawnSync(command, args, {
+    cwd,
+    env,
+    encoding: "utf8",
+    timeout: COMMAND_TIMEOUT_MS,
+    maxBuffer: 10 * 1024 * 1024,
+  });
+}
+
+function commandOutput(command, args, cwd, result) {
+  return [
+    `$ ${command} ${args.join(" ")}`,
+    `cwd: ${cwd}`,
+    `exit: ${String(result.status)}`,
+    "--- stdout ---",
+    result.stdout.trim(),
+    "--- stderr ---",
+    result.stderr.trim(),
+  ].join("\n").trim();
+}
+
+function combinedOutput(result) {
+  return `${result.stdout}\n${result.stderr}`;
+}
+
+function passingTestCount(output) {
+  const match = /\b(\d+)\s+pass\b/.exec(output);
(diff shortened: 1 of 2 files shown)


benjaminshafii:
## Daytona clean-sandbox validation — Passed

Ran this internal contract flow on a fresh Daytona Linux Electron sandbox with the combined enterprise-network branch checked out. The app itself was running, but the flow's `requiresApp: false` path used no CDP, network, database, or Den stack.

```text
✓ Frame 1 — single Den origin contract       187ms
✓ Frame 2 — private MCP default-deny         742ms
✓ Frame 3 — no token sent to untrusted host  203ms
✓ Frame 4 — diagnostics origin configurable 1464ms
```

Also on that same clean sandbox:

```text
Outbound access manifest: 35 covered, 31 observed, 0 stale
Desktop tests: 116 pass, 0 fail, 1 skipped
Installer tests: 53 pass, 0 fail
```

The flow therefore passes outside the developer's local machine and does not depend on ambient local services.

PR #3172, opened 2026-07-27
chore(settings): remove the Customization tab

benjaminshafii:
## What

Removes the **Customization** settings tab (internal tab id `shell`). It is now fully managed locally in the server, so the desktop surface is dead weight. No migration — pure cleanup.

Net **-462 lines**.

### Removed
- `apps/app/src/react-app/domains/settings/pages/shell-view.tsx` — deleted (469 lines: inline SVG shell wireframe preview, toggle-row primitives, four fake `cloudOnly` org toggles, reset-to-defaults footer).
- `apps/app/src/app/types.ts` — `"shell"` dropped from `SETTINGS_TAB_VALUES`. This also removes it from `control-provider.tsx`'s `open settings panel` action validation for free, since that validates against the same const.
- `settings-page.tsx` — icon / label / description cases removed, dropped from `getGlobalSettingsTabs`, and the now-unused `Layout` icon import removed. The sidebar and compact section dropdown are data-driven from these helpers, so no per-tab edits were needed there.
- `settings-route.tsx` — import, the `case "shell"` in `parseSettingsPath` (so `/settings/shell` and `/workspace/:id/settings/shell` no longer resolve), and the render case.

### Moved, not dropped
`window-section.tsx` — the Windows/Linux **"Display menu bar"** toggle now lives in **Appearance > Window**. It drives `applicationMenuVisible` over IPC (`__setApplicationMenuVisible` -> `apps/desktop/electron/app-menu.mjs`) to the native Electron menu bar, and the Customization tab was its only UI surface. Deleting it outright would have orphaned a working feature, so it moved rather than died.

### Deliberately untouched
- `shell/shell-config.tsx` — still read by 8 places (status bar, sidebar, notification bell, starter cards, welcome page, cloud sign-in promo). The tab wrote to it; it is not the tab.
- `ui-state-store.ts` — `applicationMenuVisible` is also published to agent context and consumed by the main process.
- The two eval flows that seed `localStorage["openwork.shell-config"]` (`dont-promote-ow-models-in-non-default-apps`, `openwork-connect-status`) — neither navigates to the tab, and the key they seed still exists.

### Docs
`.devcontainer/README.md` — the "Sign in -> Customization -> see locked toggles" verification step now points at the desktop policy banner, which is what actually still renders org restrictions (on every tab, via `DesktopPolicyBanner`).

Grep confirms zero remaining references to `shell-view`, `ShellCustomizationView`, the `shell` tab id, or `/settings/shell` across `apps/`, `packages/`, `ee/`, and `evals/`. Nothing server-side, IPC, Den-side, or in any migration ever referenced this tab.

## Tests run

| Command | Result |
|---|---|
| `pnpm typecheck` | clean |
| `pnpm --filter @openwork/app test` | **423 pass / 0 fail**, 1203 assertions across 83 files |

## Proof

Drove the running app through CDP at `localhost:5173/settings/general`, opened the settings sidebar, and enumerated the rendered nav items:

```
[ "Back to app", "Workspace", "Settings",
  "Preferences", "Permissions", "Extensions", "Advanced",
  "AI Providers", "Appearance", "Environment", "Updates", "Recovery",
  "Account", "OpenWork Connect\nBETA" ]

hasCustomization: false        // /Customization/i against document.body.innerText
```

The Global group now reads `AI Providers, Appearance, Environment, Updates, Recovery` — Customization is gone from the sidebar and absent from the entire DOM.

The moved toggle was verified by rendering `WindowSection` to static markup and asserting `"Display menu bar"`, its description, and the `windows:flex linux:flex` platform gating (temp test, not committed).

## Honest gap

I could not click the menu bar toggle in a real desktop runtime. `WindowSection` only mounts under `isDesktopRuntime()` and the row is gated to Windows/Linux, so the browser cannot show it and macOS would not either. Launching Electron locally failed: two stale Electron mains on this machine never produce page targets, and a fresh instance with an isolated `OPENWORK_ELECTRON_USERDATA` also came up windowless — the shell driving this has no GUI session.

To close it, run in the Linux Daytona sandbox (`daytona-electron-test`) and check **Settings > Appearance > Window**; the toggle should be visible there and should flip the native menu bar.


benjaminshafii:
Rebased onto `dev` (`c23f4872d`). Two conflicts, both resolved:

- `shell-view.tsx` — modify/delete. `dev` had reworked the file (469 -> 450 lines); the delete still wins, since removing the tab is the point.
- `settings-page.tsx` — `getGlobalSettingsTabs` gained a `capabilities` parameter upstream (`autoUpdate` / `localRuntimeControl` now gate the Updates and Recovery tabs). Kept the new upstream signature and gating, and only dropped `"shell"` from the array.

Re-verified on the new base:

| Command | Result |
|---|---|
| `pnpm typecheck` | clean |
| `pnpm --filter @openwork/app test` | **432 pass / 0 fail**, 1220 assertions across 84 files |

Grep for `shell-view`, `ShellCustomizationView`, the `shell` tab id, and `/settings/shell` across `apps/`, `packages/`, `ee/`, `evals/` still returns nothing.

benjaminshafii:
`openwork-tests` is red on both runners, but it is **the same single pre-existing failure that is already red on `dev`**, not something this PR introduced:

```
(fail) OpenWork capabilities knowledge plugin > retrieves the Connect-first member flow from bundled docs
       Expected to contain: "Settings` > `Connect"
```

Evidence — `OpenWork Tests` on `dev` at `c23f4872d`, which is the exact base commit of this branch, fails in the same step (`Run server unit tests`) on both `ubuntu-2204` and `macos-14` with that identical single assertion: https://github.com/different-ai/openwork/actions/runs/30258357257

This PR touches zero files under `apps/server`, so it cannot affect a bundled-docs assertion in `openwork-capabilities-knowledge.test.ts`. Everything else is green: `i18n-audit` and all five Vercel deployments pass, and `pnpm typecheck` plus the app suite (432 pass / 0 fail) are clean locally on the rebased base.

The `Settings > Connect` docs assertion looks like it needs a separate fix on `dev`.

PR #3178, opened 2026-07-27
fix(app): align starter suggestion titles across each row

benjaminshafii:
## Summary
- Starter suggestion cards on the new task screen kept both titles in a row on the same line.

## Why
The four suggestion cards are `<button>` elements that the grid stretches to a shared row height. A button's default styling vertically centers its content, so a card whose description fits on one line floated its title **8–9px lower** than the neighbour whose description wrapped to two lines, and the two headings stopped reading as a row.

Measured in the running app before the fix:

| Row | Left title y | Right title y | Drift |
|-----|--------------|---------------|-------|
| 1 | 402 | 394 | 8px |
| 2 | 487 | 496 | 9px |

Both cards in each row were the same height (86px) and shared a top edge (379px / 472px), so the drift was purely the centering.

## Issue
- Closes #

## Scope
- `session-empty-hero.tsx`: the suggestion card becomes a flex column, so its content starts at the top padding instead of being centered.
- New fraimz flow + approved voice-over proving titles stay aligned at three window widths.

## Out of scope
- Card copy, spacing, and the rest of the hero layout are untouched.

### Ran
- `pnpm fraimz --flow task-suggestion-card-alignment --cdp-url http://127.0.0.1:9826`
- `pnpm typecheck`
- `pnpm test:eval-runner`

### Result
- pass/fail: **all pass**
  - fraimz: `PASSED — 1 passed, 0 failed, 0 skipped` (3 frames + voice-over coverage)
  - typecheck: clean
  - app unit tests: `445 pass, 0 fail` (86 files)
  - eval runner tests: `10 pass, 0 fail`

**The flow is a real regression test, not a tautology.** With the one-line fix stashed, the same flow fails on frame 1:

```
✗ New task screen aligns the suggestion titles — At 1180px the card "Summarize my week"
  starts its title 24px below its top edge but its padding is only 15px, so the content
  is vertically centered instead of top-aligned.
```

Restoring the fix returns it to `PASSED`.

## CI status
- pass: to be confirmed by CI on this PR
- code-related failures: none known

## Manual verification
1. `pnpm dev` and land on the new task screen.
2. Look at the two rows of starter suggestions: in each row one description wraps to two lines and the other does not.
3. Both titles in a row sit on the same line, flush to the top of their card, at any window width.

## Evidence
fraimz proof posted as a PR comment (3 validated frames, one per window width). Each frame asserts, for every card, that `titleTop - cardTop` equals the card's top padding, that the two cards in a row share a height, that their descriptions do *not* share a height (the condition that used to break alignment), and that the two titles share a `y`.

## Risk
Very low — one `className` addition (`flex flex-col`) on a leaf presentational button. `truncate` on the title still works because the flex child stretches to the card width.

## Rollback
Revert this commit; the card returns to the centered layout.

### DIFF
--- a/apps/app/src/react-app/domains/session/chat/session-empty-hero.tsx
+++ b/apps/app/src/react-app/domains/session/chat/session-empty-hero.tsx
@@ -119,7 +119,7 @@ export function SessionEmptyHero(props: SessionEmptyHeroProps) {
           <button
             key={suggestion.title}
             type="button"
-            className="rounded-xl border border-border bg-background p-3.5 text-left transition-colors hover:bg-accent"
+            className="flex flex-col rounded-xl border border-border bg-background p-3.5 text-left transition-colors hover:bg-accent"
             onClick={() => fillPrompt(suggestion.prompt)}
           >
             <div className="truncate text-[13px] font-medium text-foreground">{suggestion.title}</div>
--- a/evals/flows/task-suggestion-card-alignment.flow.mjs
+++ b/evals/flows/task-suggestion-card-alignment.flow.mjs
@@ -1,0 +1,232 @@
+import { loadVoiceoverParagraphs } from "../runner/voiceover.mjs";
+
+// Narration is loaded from the approved script (evals/voiceovers/task-suggestion-card-alignment.md).
+// The runner fails this flow if the narration drifts from that script.
+const vo = await loadVoiceoverParagraphs("task-suggestion-card-alignment");
+
+const HERO_HEADING = "What do you need done?";
+const WINDOW_HEIGHT = 705;
+
+async function closeStaleDialogs(ctx) {
+  await ctx.eval(`(() => {
+    for (let index = 0; index < 3; index += 1) {
+      document.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+    }
+    const active = document.activeElement;
+    if (active && typeof active.blur === "function") active.blur();
+    return true;
+  })()`);
+}
+
+/** Emulate a narrower window; `null` restores the real one. */
+async function setWindowWidth(ctx, width) {
+  if (width === null) {
+    await ctx.client.send("Emulation.clearDeviceMetricsOverride");
+  } else {
+    await ctx.client.send("Emulation.setDeviceMetricsOverride", {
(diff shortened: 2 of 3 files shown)


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-27T12-48-34-733Z`

Full frame proof with validated screenshots: `evals/results/2026-07-27T12-48-34-733Z/fraimz.html` (re-run: `pnpm fraimz --flow task-suggestion-card-alignment`)

### ✅ task-suggestion-card-alignment — Starter suggestion titles align across each row even when a description wraps
_User-facing flow demo_

1. **Both titles in a row share one line while one description wraps to two**
   > 🎙 Opening OpenWork lands on the new task screen, and under the prompt box sit four starter suggestions in two rows of two. Each row stretches its two cards to the same height, but the descriptions are not the same length, so one of them wraps onto a second line while its neighbour stays on one. Both headings still start on exactly the same line, flush with the top of their card.
   - 📸 `task-suggestion-card-alignment-01-new-task-suggestions-aligned.png` — 5 validations passed
2. **Re-wrapping the descriptions at 880px keeps both titles on one line**
   > 🎙 Dragging the window narrower squeezes the cards and re-wraps the descriptions underneath the headings. The pair of cards still matches in height, the amount of text inside them still does not match, and the two headings still read as one row.
   - 📸 `task-suggestion-card-alignment-02-narrow-window-suggestions-aligned.png` — 5 validations passed
3. **Titles remain pinned to the top of each card at 700px**
   > 🎙 Narrowing further tucks the sidebar away and gives the cards a different shape again. Even here every heading stays pinned to the top of its own card, so no title drifts toward the middle while its neighbour stays put.
   - 📸 `task-suggestion-card-alignment-03-collapsed-sidebar-suggestions-aligned.png` — 5 validations passed
   - ✅ Script frame 1 narrated: "Opening OpenWork lands on the new task screen, and under the prompt box sit four starter"
   - ✅ Script frame 2 narrated: "Dragging the window narrower squeezes the cards and re-wraps the descriptions underneath"
   - ✅ Script frame 3 narrated: "Narrowing further tucks the sidebar away and gives the cards a different shape again. Ev"

PR #3181, opened 2026-07-27
chore: remove the openwork-orchestrator

benjaminshafii:
Deletes `apps/orchestrator`. **85 files, −11,797 lines.** Cloud already ran without it; this removes it everywhere else.

## Render workers are now purely openwork-server

Build installs the published `openwork-server` npm package instead of the orchestrator. The start command drops every orchestrator-only flag — `--remote-access` → `--host 0.0.0.0`; `--opencode-source`/`--opencode-bin` → `OPENWORK_MANAGE_OPENCODE` + `OPENWORK_OPENCODE_BIN`; `--allow-external` and `--connect-host` simply go.

It resolves extension plugins from the installed package and **fails loudly if they're absent**, rather than booting into the silent `extensions_plugin_missing` state Cloud just hit in production. `install-opencode.mjs` now reads the pinned OpenCode version from the repo's own `constants.json` instead of from a globally-installed orchestrator.

## The desktop stops shipping it

Every release was compiling, signing, notarizing and shipping **eight orchestrator binaries that nothing ever ran** — `engineStart` has always hardcoded the direct runtime, and both `startOrchestratorRuntime` and `startDirectRuntime` were defined and never called. Removed: the dead runtime paths, the detached-start IPC + its types, the Settings→Debug probe, and the sidecar packaging entries. Boot now calls `engineStart` directly instead of a vestigial `orchestratorWorkspaceActivate` that only delegated to it.

## Docker images converted, not stranded

The single-container and microsandbox images run `openwork-server` directly, with plugins shipped and `OPENWORK_EXTENSIONS_PLUGIN_DIR` set (mirroring the working Daytona snapshot).

**Deliberate exception:** their data directory keeps its historical `openwork-orchestrator` name. Existing deployments have volumes mounted there and renaming would silently orphan their state. I first had this behind a variable that hid the string from grep, and made it an explicit literal with a comment instead — honesty beats a clean sweep. It's a directory name; the CLI is gone.

## Release pipeline

Drops the sidecar job and the npm publish. `verify-tag.mjs`, `review.mjs` and `bump-version.mjs` no longer read a `package.json` that no longer exists — **they would have thrown on every release.**

## Preserved on purpose

Users still have `~/.openwork/openwork-orchestrator` on disk, so nuke keeps wiping it and the tests keep asserting that.

## Verification

| Check | Result |
|---|---|
| `apps/orchestrator` deleted + lockfile updated | ✅ |
| Typechecks: app · server · desktop-electron · den-api · den-web | all clean |
| 53 desktop electron tests · 20 den-api cloud · 20 server db/env · stats | pass |
| i18n audit (`--ci`) | passes all locales |
| `release/review.mjs` + `verify-tag.mjs` without the folder | run clean |
| Every changed YAML | parses |
| Server suite | **1083 pass / 7 fail — byte-identical to the pre-existing baseline** |

The 7 server failures (4 distinct: reload watcher/event timing, plus one suite-order plugin test) were confirmed identical on clean `dev` by stashing this change. The landing prebuild parity check also fails identically on clean `dev` — unrelated to this PR. Both are worth their own fix.

## Ops follow-ups (not code)

- **`openwork-orchestrator` stays published on npm.** Deleting source doesn't unpublish it — existing global installs keep working but go unmaintained. Worth `npm deprecate`-ing with a pointer to `openwork-server`.
- Docker images weren't built here (cross-arch, slow) — the conversions are unproven at image-build level and should get one CI build before anyone relies on them.


benjaminshafii:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-27T13-37-21-196Z`

Full frame proof with validated screenshots: `evals/results/2026-07-27T13-37-21-196Z/fraimz.html` (re-run: `pnpm fraimz --flow cloud-web-connect-e2e`)

### ✅ cloud-web-connect-e2e — Cloud web sign-in opens a real instance, completes an approved action, and executes a mock MCP through OpenWork Connect
_User-facing flow demo_

1. **No app without sign-in.**
   - ✅ Visible text includes "Welcome to"
   - ✅ Visible text does not include "New session"
2. **The web SPA shows the forced sign-in gate and the app shell is absent before authentication.**
   - 📸 `cloud-web-connect-e2e-01-frame-1-signin-gate.png` — 5 validations passed
3. **Admin signs in.**
   - ✅ Admin email/password sign-in returns HTTP 200
   - ✅ Admin sign-in response includes a bearer token
   - ✅ Signed-in admin can read /v1/me/orgs
   - ✅ The signed-in admin belongs to an organization
   - ✅ The active organization display name matches the eval org
   - ✅ The Den session token is available before browser seeding
   - ✅ The active organization is available before browser seeding
   - ✅ Den accepts the signed-in admin session
   - ✅ The session belongs to the signing-in admin
   - ✅ The browser carries the Den session
   - ✅ The forced sign-in gate is no longer blocking the app
4. **The admin's Den session is seeded into the browser and Den accepts it; the forced sign-in gate is gone.**
   - 📸 `cloud-web-connect-e2e-02-frame-2-signed-in.png` — 4 validations passed
5. **A machine is assigned only after sign-in.**
   - ✅ Unauthenticated cloud instance lookup returns 401
   - ✅ The authenticated cloud instance is ready
   - ✅ The Cloud instance URL is a Daytona HTTPS proxy URL
6. **The instance is the full OpenWork app.**
   - ✅ The Den session token is available before browser seeding
   - ✅ The active organization is available before browser seeding
   - ✅ The Cloud instance renders the OpenWork app shell
7. **The Cloud instance loads the full OpenWork app shell after the same signed-in organization session is restored.**
   - 📸 `cloud-web-connect-e2e-03-frame-4-instance-app.png` — 4 validations passed
8. **Admin completes an action (approved file write).**
   - ✅ The Cloud instance root HTML is fetchable
   - ✅ The instance HTML includes __OPENWORK_BOOTSTRAP__
   - ✅ The bootstrap JSON includes a client token
   - ✅ The instance client token can list workspaces
   - ✅ The Cloud instance exposes a workspace id
   - ✅ The host token can list pending approvals
   - ✅ The host token approves the pending file write
   - ✅ The file write request completes after approval
   - ✅ The approved file write returns ok:true
   - ✅ The written file can be read back from the Cloud instance
   - ✅ The action ran against a Cloud workspace
   - ✅ The file write exercised the approvals system before completing
9. **The org's mock MCP appears in Connect, in the instance.**
   - ✅ Admin can list manageable MCP connections
10. **Settings > Connect in the Cloud instance shows the org-wide Acme Mock Tools MCP connection.**
   - 📸 `cloud-web-connect-e2e-04-frame-6-connect.png` — 5 validations passed
11. **The mock MCP provably executes through OpenWork Connect.**
   - ✅ The mock MCP witness log is readable
   - ✅ Admin can mint an agent MCP token
   - ✅ The minted MCP token is present
   - ✅ MCP tools/call returned HTTP 200
   - ✅ MCP tools/call returned a data frame
   - ✅ MCP tools/call returned no JSON-RPC error
   - ✅ search_capabilities finds the mock MCP acme_lookup capability (or mock_echo fallback)
   - ✅ MCP tools/call returned HTTP 200
   - ✅ MCP tools/call returned a data frame
   - ✅ MCP tools/call returned no JSON-RPC error
   - ✅ The mock MCP witness log is readable
   - ✅ The mock MCP witness log grew after execute_capability
   - ✅ A new mock MCP request names acme_lookup
   - ✅ The execute_capability result contains the Acme lookup proof text
   - ✅ The mock MCP witness log captured the executed tool
   - ✅ The OpenWork Connect execution returned the Acme lookup proof

PR #3201, opened 2026-07-27
feat(diagnostics): capture MCP registration errors and cloud endpoint transport evidence

benjaminshafii:
## Problem

Self-hosted Den behind corporate TLS (Blue Yonder POC): `openwork-cloud` MCP registration fails on some machines while sign-in, Den API, models, and inference all work. Three weeks of debugging produced **zero captured error cause**, because:

- `server.ts` parsed only `entry.status` from OpenCode's `POST /mcp` registration response and **discarded `entry.error`** — the exact TLS message from the failing runtime (OpenCode's failed status carries `{ status: "failed", error: string }`).
- `cloud-tool-catalog` fails closed (`untrusted_endpoint`) for self-hosted origins, so its transport classifier never runs.
- `mcps[].origin` was hardcoded `null` and `path` a constant — endpoint drift between users was invisible.

## Change

1. **Persist OpenCode's registration error** (sanitized via `sanitizeDiagnosticString`, bounded to 400 chars) through the registration record into `engine-mcp-sync.details.failedRegistrations[].errorSummary`, plus a classified `transportCause` (`tls_incomplete_chain`, `tls_untrusted_ca`, `dns_error`, `connection_refused`, …) covering Node and Bun error shapes.
2. **Scrubbing instead of whole-string redaction**: URL/path segments become `[url]`/`[path]` placeholders so endpoint-bearing TLS errors survive (`failed to connect to [url] unable to verify the first certificate`), while credential-bearing strings still fully redact.
3. **New `cloud-endpoint-transport` check** — credential-free (bare `node:tls` handshake, no HTTP, no headers): exact verify error code, served chain length + subject/issuer CNs (the "is the proxy serving leaf-only?" question), DNS/TCP reachability, and local trust evidence (`tls.getCACertificates("system").length`, `NODE_EXTRA_CA_CERTS` presence + bundle cert count). Runs only when the managed cloud MCP is not connected.
4. **Real `origin` + `path`** for managed `openwork-cloud` MCP entries (no query/hash/userinfo; strict schema validators added).
5. Credentialed-probe trust gating (`untrusted_endpoint` contract, `semi-airgapped-den-contract` eval) untouched; report-safety claims stay truthful (`rawEngineErrorsIncluded: false`, new `sanitizedEngineErrorSummariesIncluded` / `credentialFreeTransportProbePerformed` flags).

## Proof (real run, real TLS-broken endpoint)

[...]

PR #3206, opened 2026-07-27
fix(diagnostics): omit SNI servername for IP-literal cloud endpoints

benjaminshafii:
## Problem
dev CI went red on d1c395bde (#3201): 3 transport-probe tests fail on the CI runtime with `verifyErrorCode: ERR_INVALID_ARG_VALUE` — only for `127.0.0.1` endpoints (the DNS-hostname test passed).

Root cause: `tlsOptions()` set `servername: url.hostname` unconditionally. SNI forbids IP literals — **Node throws `ERR_INVALID_ARG_VALUE` synchronously** ("Setting the TLS ServerName to an IP address is not permitted"). Local Bun 1.3.10 tolerates it, which is why the PR's local runs were green; CI (and the production embedded-server runtime, Node) enforce it. Reproduced with a 5-line script under `node`.

## Fix
Set `servername` only when the host is not an IP literal (`net.isIP`), stripping IPv6 brackets from URL hostnames for the raw `host` option. One function, plus regression tests: connector-captured options for DNS/IPv4/IPv6 endpoints, and an explicit not-`ERR_INVALID_ARG_VALUE` assertion on the refused-connection path.

## Tests
- `bun test` (apps/server, full suite CI runs): **589 pass / 0 fail** (599 tests, 80 files) — includes the 3 previously CI-failing tests
- Node-runtime proof (production runtime): raw `tls.connect` with the fixed option shape → `ECONNREFUSED` on closed port, `DEPTH_ZERO_SELF_SIGNED_CERT` on a self-signed server, no servername in options
- `pnpm --filter openwork-server typecheck` ✅

### DIFF
--- a/apps/desktop/electron/updater.mjs
+++ b/apps/desktop/electron/updater.mjs
@@ -262,7 +262,7 @@ export function registerUpdaterIpc({
   app,
   ipcMain,
   getMainWindow,
-  loadAutoUpdater,
+  loadAutoUpdater = () => import("electron-updater"),
   manifestChannel = "latest",
   shipItDefaultsDomain = SHIP_IT_DEFAULTS_DOMAIN,
 }) {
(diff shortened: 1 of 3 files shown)


benjaminshafii:
Added a second commit unbreaking dev's Electron typecheck (red since #3204, before this branch): `registerUpdaterIpc` required `loadAutoUpdater` but `main.mjs`/test fixtures never passed it. The electron-updater import is now the declared parameter default. `typecheck:electron` ✅, desktop tests 152/0 ✅. With both commits this PR takes dev CI from red (since 6b12dd013) back to green.

PR #3247, opened 2026-07-28
fix(app): retry model access from composer hint

reachjalil:
## Summary

- make the composer’s unavailable-model hint itself retry organization model assignment refresh
- show an in-place spinner and “Refreshing…” state while the existing sync pipeline runs
- keep the hint compact with a single-line truncated message and a short, fixed retry action
- surface a small toast if the refresh request fails
- add a focused composer contract for the retry and no-wrap behavior

## Context and root cause

Current `dev` already refreshes assigned models when the model picker opens and exposes a separate “Refresh organization models” button beside the composer warning. The remaining composer state is easy to miss, can wrap awkwardly, and gives no progress feedback after clicking.

This change reuses the existing `refreshOrganizationModels` path. It does not add a new sync mechanism or change model assignment policy.

## Checks

Checks not run (level 0, per the requested/default verification scope). `git diff --check` passed.

## Manual verification

1. Sign in as a user with no organization models assigned and open a workspace.
2. Confirm the unavailable-model hint beneath the composer is a compact, single-line retry control.
3. Assign a model to that user from the organization dashboard.
4. Click the composer hint without restarting OpenWork.
5. Confirm it shows “Refreshing…” and the model becomes available after sync.
6. Narrow the window and confirm the message truncates instead of wrapping into an awkward multi-line control.

## Risk and untested paths

The change is limited to the React composer presentation and its existing refresh callback. The live assignment journey, failure toast, and responsive layout were not exercised automatically or manually in this pass.


reachjalil:
Live Fraimz evidence for the managed-model retry follow-up is available here:

https://openwork-model-availability-retry-e.vercel.app

Verified at commit `5761b9ee6`:
- Live Electron + isolated Den flow: no assigned models → assign GPT-5.4 → click the compact Retry notice → stale warning clears and the composer becomes runnable without restarting (passed)
- Focused app tests: 7 passed
- `@openwork/app` TypeScript check: passed

The live run also exposed and fixed the stale `cloudProviderList` snapshot: manual retry imported the new model but previously left the empty-state snapshot unchanged.

PR #3259, opened 2026-07-29
[superseded by #3313] experimental: passive OpenWork Station MVP

reachjalil:
> **Superseded.** This experiment became the OpenWork Apps platform. Station is
> now an independently distributed application rather than host product source.
>
> | | |
> |---|---|
> | Contract, manifest standard, packaging tools | [#3313](https://github.com/different-ai/openwork/pull/3313) |
> | Host installer, sandboxed runtime, Preferences lifecycle | [#3313](https://github.com/different-ai/openwork/pull/3313) |
> | Station itself | [different-ai/openwork-station](https://github.com/different-ai/openwork-station) |
> | First released package | [v0.1.0](https://github.com/different-ai/openwork-station/releases/tag/v0.1.0) |
>
> Nothing here is lost. The branch is retained, and the design work it produced
> — including the uncommitted navy `#011627` treatment that was still in the
> worktree — was carried into the new repository rather than discarded.
>
> **Do not merge this.** It embeds Station in the host, which is exactly what
> the replacement removes.

---

## What this experiment established

This PR built OpenWork Station inside the monorepo: an ambient assistant with
real Realtime transcription, continuous opportunity detection, connected-source
research, priority cards, and thread handoff. It worked, and it proved the
product idea.

It also made the architectural problem obvious. Station needed the microphone, a
realtime model, read access to Slack/Gmail/Calendar, an always-on-top window, a
global shortcut, and the ability to start threads. Granting all of that to code
living inside the host means every future app of this shape has to live inside
the host too — and there is no point at which a user is asked whether they want
any of it.

So the work turned into the general question: **what would it take for anyone to
publish an application like this, and for a user to install it from a URL with
the same guarantees?**

## What replaced it

A user pastes a public GitHub repository URL. OpenWork reads a standardized
`openwork.app.json` at the commit the release tag resolves to — without cloning,
building, or executing anything — downloads the released `.owapp` package,
verifies it against a closed hashed file list, and shows the app's identity,
publisher, immutable commit, licence, package digest, and every permission it
wants with the author's own stated reason.

The app installs **switched off**. Setup and enablement are separate steps, so
installing something can never start its microphone. It runs sandboxed, reaches
only the hosts it declared, and every privileged call passes a broker that
checks the permission, a fresh single-use gesture where one is required, a
quota, and the current workspace.

Updates that ask for more than was approved are withheld until reviewed.
Rollback restores a retained verified package. Disable, uninstall, quarantine,
and revocation all stop an app the same way.

This has been run end to end against the real published release: preview,
review, install, disabled state, correct files on disk, candidate replay
refused. The transcript is in #3313.

## What carried over

- The product behaviour — continuous detection during speech rather than after a
  silence, proposing a goal before researching, cards that only cite sources
  actually retrieved, the transcript display being independent of whether
  Station is on — all lives in
  [`src/core`](https://github.com/different-ai/openwork-station/tree/main/src/core),
  now as pure modules with tests that need no microphone and no network.
- The visual language, including the navy treatment from this worktree's
  uncommitted CSS.
- The interaction fixes: Escape scoped to what Station has open, arrow keys and
  Enter only while Station holds focus, dismissal applied locally so it does not
  flicker.

## What is honestly not proven

Live connected-source retrieval and live microphone/realtime inference through
the new capability path both need a profile signed into OpenWork Cloud with an
authorized source and a configured key. That was the open blocker on this PR and
it remains the open blocker on the replacement. It is not claimed as done.


reachjalil:
Closing this superseded experiment to keep the merge queue accurate. The replacement remains #3313, as documented in this PR; the original branch and history are retained.

PR #3303, opened 2026-07-29
feat(apps): OpenWork Apps manifest standard, contract, and packaging tools

reachjalil:
## What this is

The first of a reviewable sequence that makes OpenWork hackable through
independently distributed applications. This PR adds **only the public contract
and the tooling** — no host wiring, no runtime, no UI, no Station. Nothing in
this diff changes OpenWork's behaviour; it is all new packages.

Later PRs in the sequence layer on top:

1. **This PR** — contract, manifest standard, packaging and validation tools
2. Host installer, persistence, and capability broker
3. Sandboxed Electron runtime and native surfaces
4. Dynamic right-rail contributions and the Preferences lifecycle
5. The public `different-ai/openwork-station` repository
6. Station release plus host integration against it
7. Documentation, compatibility policy, and final evidence

## Architecture

**Extension stays the umbrella; App is the new executable kind.** OpenWork
already has an extension manifest model in `apps/app/src/app/extensions.ts` with
a source-format union, resources, contributions, lifecycle, and declarative
enablement. This work layers onto it rather than replacing it, so existing
extension behaviour is untouched.

**Secrets stay host-owned.** The manifest declares *which* environment variables
an app needs. The host keeps using `apps/server/src/env-file.ts` (`EnvService`,
atomic `0o600` writes, reserved-prefix policy) and tells an app only whether a
key is configured. There is no vocabulary for reading a value.

**Install only prebuilt releases.** No clone, no build scripts, no package
manager hooks, no manifest-provided shell command, no mutable-branch install.

### `packages/app-contract`

`openwork.app.json`, `manifest_version: 1` — identity, publisher, repository,
license, icons, engine and App API compatibility windows, platforms,
distribution, entrypoints, contributions, permissions, environment requirements,
privacy disclosure, and update/rollback metadata.

Design choices worth reviewing:

- **App ids are reverse-DNS and must contain a dot.** Built-in extension ids
  never contain one, so an installed app structurally cannot shadow a built-in.
- **Unknown fields are rejected, not stripped.** A misspelled key in a
  permission block fails the manifest instead of silently dropping authority.
- **Duplicate JSON keys are rejected.** A strict parser replaces `JSON.parse`,
  because last-wins duplicate keys are a parser-differential: a linter and the
  installer could read different permission sets from the same file.
- **Compatibility is `{ min, max_exclusive }`, not a range expression.** There
  is no npm-style grammar to mis-parse, and the failure mode of a mis-parsed
  range is installing an app the host cannot run.
- **Disclosure and enforcement must agree.** A host named in
  `privacy.third_parties` must be permitted by `network.host`; requesting
  `audio.microphone` requires disclosing `microphone-audio`. A manifest that
  promises less than it asks for is refused.
- **One canonical validator** backs the CLI, CI, server preview, install, and
  update. A second validator would drift, and a drifted validator is how an app
  gets installed with authority the review screen never showed.
- **JSON Schema is generated from the same Zod schemas**, so it cannot drift
  from the validator.

### Permission table

| Permission | Risk | Consent | Grants |
|---|---|---|---|
| `audio.microphone` | critical | enable | Microphone capture, only with a visible listening state |
| `openwork.connect.read` | critical | enable | Read-only Connect queries, limited to declared scopes |
| `runtime.background.continuous` | high | enable | A background runtime that keeps running |
| `openwork.threads.start` | high | use | Start an OpenWork thread, per user gesture |
| `network.host` | high | install | Network access to exactly the listed hostnames — no wildcards |
| `ai.realtime` | moderate | enable | A host-minted, short-lived realtime credential |
| `ai.inference.transient` | moderate | install | Structured one-shot inference |
| `desktop.floatingSurface` | moderate | enable | A host-owned floating window |
| `openwork.attachments.create` | moderate | use | Attach a document to a thread, per user gesture |
| `desktop.globalShortcut` | low | enable | Register the named global shortcuts |
| `storage.app` | low | install | App-scoped storage within a quota |

The three consent stages are independent on purpose: `install` is recorded at
trust review, `enable` when the runtime first starts, and `use` additionally
requires a fresh host-issued gesture token for **every** call. Installing an app
can never start its microphone.

The lifecycle model keeps installation, setup, enablement, and runtime as four
independent axes for the same reason.

### What an app cannot request

There is no vocabulary for shell execution, arbitrary filesystem access, Node,
native binaries, accessibility or screen capture, reading raw environment values
or provider credentials, arbitrary MCP execution, arbitrary IPC, or arbitrary
OpenWork server routes. These are **absent rather than gated** — an app cannot
request what cannot be spelled.

### `packages/app-tools`

The deterministic `.owapp` format and the `openwork-app` CLI.

An archive holds `openwork.app.json`, the app's assets, and
`META-INF/openwork-package.json` carrying per-file sizes and SHA-256 digests
plus the source repository, release tag, and immutable commit. The metadata
*closes* the archive — every other entry must be listed and every listed entry
must be present — so a smuggled extra file and a missing file are both caught
without trusting the ZIP directory.

The writer and reader are hand-rolled rather than a dependency, for two reasons:

- **Determinism.** General archivers embed mtimes and host attributes, so the
  same input produces different bytes on different machines and the digest stops
  meaning anything. Entries are sorted, timestamps fixed at the 1980 ZIP epoch,
  no extra fields, no directory entries, zeroed external attributes.
- **The reader is the attack surface.** Zip-slip, symlink escape, decompression
  bombs, and ZIP64 confusion are bugs in extractors. This reader fails closed on
  all of them, cross-checks every local header against the central directory,
  and never touches the filesystem — callers decide what to write, after
  verification.

Extraction is separately defended: paths re-checked against the destination
root, exclusive-flag writes so a planted symlink is an error rather than a
redirected write, and an atomic rename so a crash never leaves a runnable
half-install.

## Trust model — stated honestly

v1 provenance is a GitHub release binding plus content hashing. **It is not a
publisher signature** and does not attest who produced the bytes. The manifest
reserves room for signing and transparency later without a v1 break, and the CLI
says this in its own `verify` output rather than letting a checksum read as
cryptographic publisher identity.

## Tests

Two bugs in this PR's own code were found by these tests and fixed rather than
worked around:

- Unknown fields inside permission objects were being silently **stripped**
  instead of rejected — precisely the fail-closed requirement.
- The ZIP reader read the compression method at local-header offset +10 (the
  timestamp field) instead of +8. Stored entries passed by coincidence; deflated
  entries did not.

Coverage includes: invalid manifest, unsupported manifest version, engine and
platform incompatibility, duplicate contribution and permission ids, dangling
contribution references, unreferenced entrypoints, wildcard and scheme-bearing
network hosts, undisclosed third-party hosts, path traversal, symlink entries,
directory entries, duplicate archive names, local/central header mismatch,
encrypted entries, data descriptors, ZIP64, decompression bombs, oversized
entries, CRC mismatch, hash mismatch, size mismatch, undeclared and missing
files, manifest/metadata mismatch, identity and version drift against a pinned
candidate, extraction path escape, and refusing to overwrite an existing install.

```
pnpm --filter @openwork/app-contract test      100 pass, 0 fail
pnpm --filter @openwork/app-contract typecheck clean
pnpm --filter @openwork/app-contract build     clean
pnpm --filter @openwork/app-tools test          62 pass, 0 fail
pnpm --filter @openwork/app-tools typecheck    clean
pnpm --filter @openwork/app-tools build        clean
```

End-to-end from a real app directory on this branch:

```
openwork-app validate  ->  ok: com.openworklabs.station 1.0.0 — 4 permission(s), 3 contribution(s)
openwork-app pack      ->  2196 bytes, sha256:c2884ca6…7103
openwork-app verify    ->  verified, provenance shown, signature explicitly disclaimed
shasum -a 256          ->  c2884ca6…7103   (independently agrees)
unzip -l               ->  5 entries, all timestamped 1980-01-01 (determinism controls active)
```

`unzip` reading the archive also confirms `.owapp` is a spec-compliant ZIP, not
a bespoke container.

## Fraimz

Not applicable to this PR and stated explicitly rather than skipped: it adds two
new packages with no runtime path in the app, server, or desktop. Nothing
observable in the product changes. The first observable behaviour arrives with
the installer and Preferences work, and `pnpm fraimz --flow
openwork-app-install-station` lands with the end-to-end integration.

## Known limitations

- `@openwork/app-contract` and `@openwork/app-tools` are publishable in shape
  but not yet published; the `@openwork` npm scope is unclaimed. External
  consumption before publication is via this repository. Publication needs
  separate authority.
- v1 has no publisher signing, by design and by explicit statement.
- Cross-zlib-version byte reproducibility is not claimed. Integrity does not
  depend on it — every digest is over uncompressed content and the finished
  archive.

## Rollback

Both packages are new and unreferenced by any existing code path. Reverting this
commit removes them with no migration and no effect on any current behaviour.


reachjalil:
Consolidated into a single PR against `dev`: **#3313**.

Same commits, one review surface. Closing this so there is one place to look.

PR #3307, opened 2026-07-29
fix(server): a no-op provider patch must not reload the engine

benjaminshafii:
Class-level follow-up to #3301. Reloading the opencode engine disposes it, which **aborts any in-flight agent run** and disconnects the SSE event stream. A patch that changes nothing should be inert.

## Why this matters

In production, den-api re-materialized providers on a ~18s poll and every patch reloaded the engine, mid-tool-call:

```
"disposing instance" /tmp/openwork-workspace
ERROR process ... error=Aborted
"event disconnected"
```

991 disposals in a single user instance. #3301 stopped den-api sending redundant patches; this makes the server itself robust so no caller can cause it again.

## The defect

`writeRuntimeOpencodeConfig` already computed the right answer:

```ts
if (row?.valueJson === configJson) return { config: next, changed: false };
```

…but `PATCH /runtime-config/providers` ignored `result.changed` and unconditionally rewrote the file, reloaded the engine, and reported `reload: "reloaded"`.

## Fix

Skip condition is `!result.changed && !fileResult.changed` — deliberately **not** just `if (result.changed)`. The DB row can be unchanged while the on-disk runtime config is missing or stale (first boot after a checkpoint restore, or a hand-edited file); that case must still write and reload. `writeOpenworkRuntimeConfigFile` now byte-compares before its atomic write and returns `{ path, changed }`, so the route can tell the difference. `reload` now reports `"reloaded"` vs `"skipped"` honestly.

No other route was found with the same unconditional reload-after-write pattern against this store.

## Tests

Regression fails before (1 pass / 1 fail), and asserts the reload genuinely does not happen rather than trusting the response field. Also covered: a real update still reloads exactly once; DB-unchanged-but-file-stale still writes and reloads; removal still reloads.

| Command | Result |
| --- | --- |
| `pnpm --filter openwork-server test` | 592 pass / 10 skip / 0 fail |
| `pnpm --filter openwork-server typecheck` | exit 0 |

Note: this ships in the sandbox image, so it takes effect on the next snapshot + recycle — unlike #3301, which was live immediately.

### DIFF
--- a/apps/server/src/cli.ts
+++ b/apps/server/src/cli.ts
@@ -45,9 +45,9 @@ if (!config.opencodeBaseUrl && process.env.OPENWORK_MANAGE_OPENCODE === "1") {
   const workspace = findManagedEngineWorkspace(config.workspaces);
   if (workspace) {
     // Server-managed config file: the engine re-reads it from disk on every
-    // instance rebuild, and keepOpenworkRuntimeConfigFileFresh rewrites it
+    // instance rebuild, and keepOpenworkRuntimeConfigFileFresh synchronizes it
     // on every runtime-DB write — so disposes always pick up current state.
-    const runtimeConfigPath = await writeOpenworkRuntimeConfigFile(config, workspace.id);
+    const { path: runtimeConfigPath } = await writeOpenworkRuntimeConfigFile(config, workspace.id);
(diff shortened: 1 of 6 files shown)


benjaminshafii:

PR #3308, opened 2026-07-29
fix(app): advance the session-groups event cursor instead of refetching from 0

benjaminshafii:
The session-groups poller re-requested the entire event history every 3 seconds, forever.

## Evidence (production CDP trace)

```
+0.27s  -> GET /workspace/ws_49372819c6f0/events?since=0
+3.27s  -> GET /workspace/ws_49372819c6f0/events?since=0
+6.27s  -> GET /workspace/ws_49372819c6f0/events?since=0
...never advancing
```

## Cause

The poller preferred `response.cursor`, and a stale `0` overwrote the cursor derived from the returned events on every tick — so it reset itself each poll.

## Fix

Server semantics are **exclusive** (`event.seq > since`), now matched explicitly. The poller commits the highest returned `seq` only *after* events are applied, so a failure can't skip events. Empty responses preserve the cursor rather than resetting to 0. Errors, gaps, and reset signals fall back to 0 deliberately. Cursor state is keyed per endpoint/workspace and dropped when a workspace goes away, so no workspace can inherit another's cursor.

Shared with desktop; the hook is common to both paths.

## Tests

Regression fails before — `Expected: [0, 2] / Received: [0, 0]` (1 pass / 4 fail) — and passes after (5 pass). Covers cursor advance, empty-response preservation, workspace isolation, boundary correctness against the server's exclusive semantics, and error/gap fallback.

| Command | Result |
| --- | --- |
| `apps/app` suite | 533 pass / 0 fail |
| `tsc --noEmit` + `build:web` | clean |

### DIFF
--- a/apps/app/src/react-app/shell/session-group-event-poller.ts
+++ b/apps/app/src/react-app/shell/session-group-event-poller.ts
@@ -1,0 +1,27 @@
+import type { OpenworkSessionGroupEvent } from "@/app/lib/openwork-server";
+
+export type SessionGroupEventResponse = {
+  items: OpenworkSessionGroupEvent[];
+  cursor?: number;
+};
+
+export class SessionGroupEventPoller {
+  private cursorByWorkspace = new Map<string, number>();
+
+  async poll(
+    key: string,
+    request: (options: { since: number }) => Promise<SessionGroupEventResponse>,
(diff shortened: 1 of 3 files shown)


benjaminshafii:

PR #3309, opened 2026-07-29
fix(app): type sonner CSS custom properties so tsc is clean

benjaminshafii:
`apps/app` typecheck was failing on `dev`. Green gate restored.

```
src/components/ui/sonner.tsx(42,9): error TS2353: Object literal may only specify known properties,
and '"--normal-bg"' does not exist in type 'Properties<string | number, string & {}>'.
```

It only reproduces on a **fresh install** of unmodified `dev`, not against stale `node_modules` — which is why it went unnoticed, and why I initially misattributed it to an unrelated PR's dependency change.

## Fix

The style object is declared as `React.CSSProperties & Record<`--${string}`, string>`, so CSS custom properties are legal without a cast. No `as`, no `any`. Built-JS grep confirms all four custom properties and their values are unchanged.

| Command | Result |
| --- | --- |
| `tsc --noEmit` | exit 0, no output |
| `apps/app` suite | 528 pass / 0 fail |
| `build:web` | exit 0 |

### DIFF
--- a/apps/app/src/components/ui/sonner.tsx
+++ b/apps/app/src/components/ui/sonner.tsx
@@ -14,6 +14,13 @@ function useTheme() {
   )
 }
 
+const toasterStyle: React.CSSProperties & Record<`--${string}`, string> = {
+  "--normal-bg": "var(--popover)",
+  "--normal-text": "var(--popover-foreground)",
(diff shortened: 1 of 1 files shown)


benjaminshafii:

PR #3311, opened 2026-07-29
feat(apps): host installer, sandboxed runtime, and Preferences lifecycle

reachjalil:
## What this is

The host half of the OpenWork Apps platform, and the proof it works against a
real published application.

**Depends on [#3303](https://github.com/different-ai/openwork/pull/3303)** — this
branch is stacked on it. Review that first; this diff is only the commits on top.

By the end of this PR, OpenWork can install
[different-ai/openwork-station](https://github.com/different-ai/openwork-station)
from its public GitHub release, over the real network, with nothing mocked.

## The journey, run for real

Against published release `v0.1.0`:

```
1. Paste https://github.com/different-ai/openwork-station

2. Preview — no repository code executed
   app          com.openworklabs.station 0.1.0
   publisher    OpenWork Labs
   release      v0.1.0
   commit       e64d21712a85473c5cd17e2a0977b452c5b1d981
   asset        openwork-station-0.1.0.owapp
   digest       sha256:d3bd3c4cb808b2b0e17449d5ff00b52404e98aa762fe5acf327054afb7968627
   compatible   true
   candidate    c79cec546d40549d… expires in 900s

3. Permissions for review, critical first
   [critical] audio.microphone
   [critical] openwork.connect.read (slack.search, gmail.search, calendar.events.read)
   [high    ] network.host (api.openai.com)
   [high    ] openwork.threads.start
   [high    ] runtime.background.continuous
   [moderate] ai.inference.transient
   [moderate] ai.realtime
   [moderate] desktop.floatingSurface (stays above other windows)
   [moderate] openwork.attachments.create
   [low     ] desktop.globalShortcut (CommandOrControl+Shift+Space)
   [low     ] storage.app (up to 1024 KB)
   environment: OPENAI_API_KEY (not configured)

4. Installed
   installation installed
   enablement   disabled          <- switched off, as designed
   setup        setup_required
   digest match true

5. On disk: LICENSE, openwork.app.json, assets/icon.svg,
   dist/background.js, dist/station/{index.html,index.js,style.css}

6. Replaying the candidate → refused

7. Audit: trust_granted, installed
```

Reproduce it:

```bash
OPENWORK_APPS_LIVE=1 pnpm --filter @openwork/server test src/apps/live-install.e2e.test.ts
```

It is network-gated so it does not fail offline, and it **reports when it is
skipped** rather than passing quietly.

### `apps/server/src/apps` — installation and enforcement

**Source adapter.** Reads metadata and downloads one asset. No clone, no build,
no package-manager hook, no manifest-provided command. It resolves a release tag
to an immutable commit, dereferencing annotated tags, and deliberately ignores
`target_commitish` because that can be a branch — the mutable thing this step
exists to eliminate.

**Preview/install split.** Preview downloads, hashes, and verifies the package,
then pins every input the user is about to see into a short-lived single-use
candidate. Install consumes that candidate and those exact bytes and re-resolves
nothing. A release swapped between review and install therefore installs *what
was reviewed*; a replayed or expired candidate is a distinct reported failure.
There is a test that moves the tag and swaps the asset for one demanding the
microphone, and asserts the original permissions are what get installed.

**Approval matching.** Approved permissions are compared against the candidate's
as an unordered set of parameterised keys, so a client can neither approve a
narrower list and get a wider one, nor slip a permission past a review that
never showed it.

**Install disabled.** Setup, enablement, and runtime are independent axes. This
is what makes "installing an app never starts its microphone" structural rather
than conventional.

**Updates.** An update that adds or widens a permission is verified and withheld
as pending until reviewed. Removals and narrowings apply immediately. Rollback
restores the retained previous package — a real rollback, not a re-download —
and leaves it stopped, because resuming would restart whatever went wrong.

**Crash quarantine** counts inside a rolling window, so an app crashing once a
day is never quarantined by accumulation.

**Capability broker.** Every privileged call passes schema → activation →
permission → fresh single-use gesture → quota → workspace binding, in that
order. Connect scopes are checked against the grant, not merely the vocabulary.
Denials are audited without secrets, payloads, or query text.

### `apps/desktop/electron/apps` — the sandboxed runtime

`openwork-app://` resolution is pure and total: an app can address only files
inside its own installed version, only servable types, and only while running —
so disabling an app cuts off its content immediately rather than eventually.

Surfaces run with `sandbox`, `contextIsolation`, no node integration in any
form, a per-app session partition, `default-src 'none'` CSP, exact-host network
filtering at the session level, denied window creation, and external links
handed to the browser instead of loaded in the sandbox.

Teardown is exhaustive and continues through failures — a wedged window must not
leave a global accelerator bound to a dead app, and there is a test that makes
`destroy()` throw and asserts the shortcut still comes back.

The preload bridge exposes two functions and no channel parameter.
`ipcRenderer` is never exposed in any wrapped form.

### `apps/app/src/react-app/domains/apps` — contributions and Preferences

Contributions are **derived from state, never registered imperatively**, so
disabling, uninstalling, quarantining, or revoking a permission withdraws them
on the next render. There is no unregister call to forget. `resolveRuntimePlan`
is the whole stop mechanism: nothing tells the supervisor to stop an app, an app
simply stops appearing in the plan.

Registration keys are `<appId>/<contributionId>`; built-in ids contain no slash
and app ids must contain a dot, so an app can neither claim a built-in slot nor
collide with another app. Ordering is deterministic — built-ins keep their order
and stay ahead of installed apps, so installing something never reshuffles an
interface the user already knows.

The install flow is a state machine, so the approved permission set is read from
the preview being *displayed*. What is submitted is necessarily what was shown.

## Security model

| Concern | How it fails closed |
|---|---|
| Repository code execution during preview | Never cloned, built, or run. Metadata + one asset download. |
| Mutable release | Tag resolved to an immutable commit; manifest read at that commit. |
| TOCTOU between review and install | Candidate pins digest, commit, app id, version, permissions. Install re-resolves nothing. |
| Candidate replay / expiry | Single-use, 15-minute TTL, distinct typed failures. |
| Client approving different permissions | Unordered parameterised-key set comparison against the candidate. |
| Zip-slip, symlink escape, zip bomb, ZIP64 | Hardened reader (in #3303); atomic exclusive-flag extraction. |
| Undeclared or missing file in package | Package metadata closes the archive over per-file hashes. |
| App reaching an unapproved host | Session-level exact-host filter plus CSP `connect-src`. |
| App reading a secret | No vocabulary for it. Only key *names* cross the boundary; status is configured/missing. |
| Silent permission expansion on update | Update withheld pending review. |
| App starting a thread on its own | Fresh single-use gesture token, seconds-long TTL, app-bound, constant-time compare. |
| Stale request after a workspace change | Requests are workspace-bound and in-flight work is aborted. |
| Crash loop destabilising OpenWork | Rolling-window quarantine, exhaustive teardown. |

## Migration for `featureFlags.station`

Station is **not** embedded in this PR and never was on `dev` — the in-tree
prototype lives only on the branch behind
[#3259](https://github.com/different-ai/openwork/pull/3259), which this
supersedes. Users who enabled `featureFlags.station` are pointed at
Preferences → Apps to install the public release. Nothing is silently installed
or executed.

## Tests

```
pnpm --filter @openwork/server test        661 pass, 10 skip, 0 new failures
pnpm --filter @openwork/server typecheck   clean
pnpm --filter @openwork/desktop test       210 pass, 1 pre-existing skip
pnpm --filter @openwork/app test           571 pass, 0 fail
pnpm --filter @openwork/app typecheck      clean
OPENWORK_APPS_LIVE=1 … live-install        6 pass  (real github.com)
```

New coverage: TOCTOU release swap, candidate replay and expiry, permission
mismatch in both directions, install-disabled, setup gating and regression,
update withheld on added and on widened permission, downgrade refusal, rollback
with and without a retained package, crash quarantine and window expiry, repair
with a mismatched package, uninstall retention both ways, registry corruption
dropped on load, broker denial for every gate, gesture expiry/replay/wrong-app,
connect scope outside grant, quota rolling window, protocol traversal and
symlink and type refusals, exact-host matching including lookalike hosts, CSP
shape, teardown through a throwing window, and deterministic contribution
ordering.

The repository's bare-fetch guard caught the GitHub adapter reaching for the
global. Fixed properly — external egress now goes through `externalFetch` so it
carries the configured proxy and CA trust — rather than by relaxing the check.

## Fraimz

Not yet. The Preferences view is new UI and needs a frame-by-frame journey; the
flow it drives is proven at the service layer by the live install above.
`pnpm fraimz --flow openwork-app-install-station` is the next piece of work on
this branch, and I have not claimed it as done.

## Known limitations

- `@openwork/app-contract` and `@openwork/app-tools` are publishable in shape
  but unpublished; the `@openwork` npm scope is unclaimed. Station consumes the
  tools from a checkout, which is why its packaging script takes a `--tools`
  path. Publication needs separate authority.
- Live **connected-source** research is unproven: it needs a profile signed into
  OpenWork Cloud with an authorized Slack/Gmail/Calendar source. The broker path
  is covered by tests; a real cited retrieval is not claimed.
- Live **microphone and realtime inference** through the new capability path is
  likewise unproven end to end for the same reason — it needs a configured
  `OPENAI_API_KEY` on a signed-in profile.
- The Preferences view is not yet mounted into the settings router.

## Rollback

Every server, desktop, and app change is additive and behind new modules. The
only edits to existing files are the wiring in `apps/server/src/server.ts`, one
line in the desktop test script, and two `package.json` dependency additions.
Reverting the commits removes the platform with no migration and no effect on
any existing extension.


reachjalil:
Consolidated into a single PR against `dev`: **#3313**.

Same commits, one review surface. Closing this so there is one place to look.

PR #3327, opened 2026-07-30
feat(den-api): convert Office uploads to native Google Docs, and stop advertising an impossible inline base64 limit

benjaminshafii:
## Why

Two separate defects, found while trying to get a `.docx` agreement into Drive as an editable Google Doc.

**1. The Drive upload capability cannot produce a Google Doc.** `POST /v1/capabilities/google-workspace/drive-files` built its multipart metadata as `{ name, parents }` with no `mimeType`, so an uploaded `.docx` could only ever land as a `.docx`. There was no path to a native, editable, commentable Google Doc — which is what you actually want for anything that needs redlines.

**2. The `dataBase64` descriptions instruct an operation that cannot succeed.** The schema told agents to *"Read the file from the active workspace and base64-encode it"* and advertised `maxLength: 13981016` (10 MiB).

That 10 MiB ceiling is real server-side but unreachable in practice. I audited every hop and **there is no byte limit anywhere in our code**:

| Hop | Cap |
|---|---|
| `execute_capability` `inputSchema.body` | none — `z.unknown()` (`mcp/agent.ts:517`) |
| `normalizeToolBody` → `JSON.stringify` → `app.fetch` | none (`mcp/invoke.ts:29,97,132`) — in-process, never crosses a socket |
| den-api Hono app | none — no global `bodyLimit` |
| `@hono/mcp` transport | none |
| `uploadDriveFileBodySchema` | 10 MiB |

The actual ceiling is the **model's tool-call output budget**: the base64 must be emitted by the model as a JSON string argument. A ~137KB payload (a small `.docx`) truncated mid-generation and arrived as `{}` — no server error, because the request never became a well-formed tool call. So we were advertising ~100× more headroom than is reachable, and telling agents to walk straight into it.

## What changed

- **`convertTo`** (`document` | `spreadsheet` | `presentation`), optional. Sets the metadata mimeType to the target Google editor type while the media part keeps the caller's source mimeType — the documented Drive conversion mechanism. `.docx`→Doc, `.xlsx`→Sheet, `.pptx`→Slides.
- Omitting `convertTo` is **byte-identical to today** (no `mimeType` key emitted at all). Pinned by the pre-existing test, which asserts the exact metadata JSON `{"name":"plan.pdf","parents":["folder_1"]}`.
- Reworded both `dataBase64` descriptions to state the practical inline limit and keep 10 MiB labelled as a *server* ceiling, not a usable one.
- Updated `evals/flows/google-workspace-gmail-draft-attachments.flow.mjs:77`, which asserted the old wording verbatim.

No validation limits changed. Change 2 is descriptions only.

## Tests

```
$ pnpm --filter @openwork-ee/den-api exec bun test test/google-workspace-capabilities.test.ts
 31 pass
 0 fail
 188 expect() calls

$ pnpm --filter @openwork-ee/den-api exec tsc --noEmit
(exit 0, no output)
```

Three new tests: conversion sets the right metadata mimeType *and* preserves the source media mimeType; invalid `convertTo` 400s without calling Google; the existing omission test is unchanged and still passes.

## Validation honesty

Per AGENTS.md: **there is no live Drive proof in this PR, and I want to be explicit about that.** Verifying an actual `.docx`→Google Doc conversion requires a connected Google Workspace account hitting real Drive, against a deployed den-api carrying this change. I ran the unit suite against the fake Google, which asserts the exact multipart bytes we send — that proves *we send the documented conversion request*, not that Google converts it.

**Reviewer check:** upload a `.docx` with `convertTo: "document"` and confirm `file.mimeType` comes back as `application/vnd.google-apps.document` and `webViewLink` opens in the Docs editor.

## Deliberately not in this PR

- **Path- or URL-based upload.** This is the real fix for defect 2 — bytes should never transit model context. The correct pattern already exists at `apps/server/src/extensions/google-workspace.ts:91-104` (takes `attachments: [{ path }]`, server reads the bytes at `:759`), but it has no Drive equivalent and is gated off for Cloud users (`extensions/index.ts:61-72`). Closing that gap spans `apps/server`↔den-api auth, or needs server-side URL fetching with real SSRF hardening. That's a design decision, not a patch.
- **`packages/docs/openapi.json`.** It is already stale on `dev`. Regenerating it pulls in ~230 unrelated keys (cloud gateway routes, `/v1/org.delete`, install-connect, marketplace-capabilities) and 7 unrelated description edits. That belongs in its own chore PR. The runtime path reads live-generated OpenAPI (`mcp/catalog.ts`), so this change takes effect regardless.
- **The same trap in reverse.** `gmail-attachment` returns up to 10 MiB of base64 *into* model context (`google-workspace.ts:553`) and will fail the same way on large attachments.

### DIFF
--- a/ee/apps/den-api/src/routes/org/google-workspace.ts
+++ b/ee/apps/den-api/src/routes/org/google-workspace.ts
@@ -38,6 +38,12 @@ const GOOGLE_WORKSPACE_API_TIMEOUT_MS = 30_000
 const MAX_GMAIL_ATTACHMENT_BYTES = 10 * 1024 * 1024
 const MAX_GMAIL_ATTACHMENTS_TOTAL_BYTES = 20 * 1024 * 1024
 const MAX_GMAIL_ATTACHMENT_BASE64_LENGTH = Math.ceil(MAX_GMAIL_ATTACHMENT_BYTES / 3) * 4
+const uploadDriveFileConversionSchema = z.enum(["document", "spreadsheet", "presentation"])
+const DRIVE_CONVERSION_MIME_TYPES: Record<z.infer<typeof uploadDriveFileConversionSchema>, string> = {
+  document: "application/vnd.google-apps.document",
+  spreadsheet: "application/vnd.google-apps.spreadsheet",
+  presentation: "application/vnd.google-apps.presentation",
+}
 const GMAIL_REPLY_SUBJECT_RE = /^\s*(re|fwd?)\s*:/i
 
 const CONNECT_GOOGLE_ACCOUNT_MESSAGE = "Connect your Google account first: open Settings > Connect and use Connect your account on the Google Workspace row, or connect from the OpenWork Cloud dashboard."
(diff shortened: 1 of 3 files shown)


reachjalil:
Thanks for digging into this and for identifying the base64/tool-call limitation. My product preference is that OpenWork preserve users’ files exactly as supplied rather than converting Office files on their behalf—or allowing the agent to choose a conversion implicitly. If someone uploads a `.docx`, `.xlsx`, or `.pptx`, we should keep the original bytes, filename, and MIME type; if they want a native Google Doc, Sheet, or Slides file, they can make that conversion explicitly in Google Drive.

I’m addressing the underlying upload/download transport separately with a byte-outside-model architecture: scoped, expiring file references; multipart Google Drive uploads for small files; resumable uploads for larger files; and download materialization without returning file bytes through model context.

Given that direction, I recommend that we do not merge the `convertTo` behavior from this PR. The investigation and corrected guidance around the impractical inline-base64 limit are valuable, and I’ll carry that finding into the parallel transport fix.

PR #3346, opened 2026-07-30
fix(app): open Browser before a session starts

reachjalil:
## Summary

- Keep the unified right panel addressable while the workspace is on **New session**.
- Let **Browser** create and display a native browser tab before any chat session has started.
- Preserve the existing per-session panel behavior once a real session is selected.
- Add deterministic evaluator coverage for the session-free Browser path.

## Root cause

The right panel was keyed and rendered only by a selected chat session, so its state update was dropped when no session existed. After making the panel addressable, the destination button also forwarded React's click event into the optional browser URL parameter, which native IPC could not clone into a tab request.

## What changed

- Introduced a stable internal panel key for null or blank session identifiers.
- Routed no-session panel state and tabs through that key.
- Removed the selected-session render gate for the general side panel.
- Wrapped destination activation so UI events are never forwarded as action arguments.
- Updated the right-panel evaluator to start from a clean, closed panel and prove Browser tab creation with no selected session.

## Verification

- `bun test tests/right-panel-empty.test.tsx` — 5 passed, 0 failed.
- `pnpm evals:typecheck` — passed.
- `node --check evals/flows/right-panel-action-empty-state.flow.mjs` — passed.
- `pnpm fraimz --flow right-panel-action-empty-state --cdp-url <isolated-candidate>` — 1 passed, 0 failed, 0 skipped.
- Full app suite observation: 557 passed and 2 unrelated existing failures in `message-list-loading.test.tsx` (`target.addEventListener is not a function` in the test's partial window shim). This branch does not touch that surface.

## Risk

Low and localized to right-panel state ownership and destination activation. The fallback key is internal-only and real session IDs continue to use their existing isolated state.

### DIFF
--- a/apps/app/src/react-app/domains/session/chat/session-page.tsx
+++ b/apps/app/src/react-app/domains/session/chat/session-page.tsx
@@ -68,6 +68,7 @@ import { isCollectibleArtifactTarget, isLocalhostBrowserTarget, isOpenableFileTa
 import type { OpenTargetOptions } from "@/lib/target-provider";
 import { VoicePanel } from "../voice/voice-panel";
 import { SidePanel } from "../panel/side-panel";
+import { getSidePanelSessionKey } from "../panel/side-panel-session";
 import { TerminalDock } from "../terminal/terminal-dock";
 import { useActivePanelTab, usePanelTabStore, useSessionPanelState } from "../panel/panel-tab-store";
 import { useWorkspaceShellLayout } from "../../../shell/workspace-shell-layout";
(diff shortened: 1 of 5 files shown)


reachjalil:
## fraimz — ✅ PASSED

1 passed · 0 failed · 0 skipped — run `2026-07-30T11-16-11-696Z`

Full frame proof with validated screenshots: `evals/results/2026-07-30T11-16-11-696Z/fraimz.html` (re-run: `pnpm fraimz --flow right-panel-action-empty-state`)

### ✅ right-panel-action-empty-state — The built-in browser opens before a chat session starts
_User-facing flow demo_

1. **Browser remains available before any chat session is selected or started**
   - 📸 `right-panel-action-empty-state-01-browser-open-without-session.png` — 4 validations passed

PR #3351, opened 2026-07-30
refactor(app): replace MCP connected count polling with global SSE event listener (#3126)

### LINKED ISSUE #3126: [Enhancement]: Replace polling on /workspace/:id/events and /session-groups/events with Server-Sent Events (SSE)
### Problem

The client currently polls two endpoints on a tight interval (sub-second) to stay in sync with workspace state:

- `GET /workspace/:id/events`
- `GET /workspace/:id/session-groups/events`

Sample from local dev logs (~1 second window):

```
GET /workspace/ws_8901851c2c65/events 200 5ms
OPTIONS /workspace/ws_8901851c2c65/session-groups/events 204 0ms
GET /workspace/ws_8901851c2c65/session-groups/events 200 0ms
OPTIONS /workspace/ws_8901851c2c65/events 204 0ms
GET /workspace/ws_8901851c2c65/events 200 2ms
```

This pattern repeats continuously for the lifetime of an open workspace, generating roughly 6–8 requests/second per client (each `GET` preceded by an `OPTIONS` preflight).

**Why this is a problem:**
1. **Wasted round trips** - most polls return no meaningful state change, so the majority of requests do nothing but consume server and network resources.
2. **CORS preflight overhead** - every `GET` is preceded by an `OPTIONS` request, doubling actual request volume for zero payload gain. This suggests `Access-Control-Max-Age` isn't set/cached properly on the client side, but even with that fixed, polling itself remains wasteful.
3. **Scales linearly with active workspaces** - each open client sustains this load indefinitely. In multi-tenant or hosted environments this compounds quickly.
4. **No backoff when idle** - polling continues at the same rate regardless of whether anything has changed, or whether the tab/window is even in the foreground.
5. **Endpoint naming already implies push** - both endpoints are literally named `events`, suggesting the original intent was a streaming interface that ended up implemented as poll-based, likely as a stopgap.

### Proposed solution

Migrate both endpoints from polling to Server-Sent Events:

- **Auth cannot use native `EventSource`.** All requests currently require `Authorization: Bearer <token>` (`openwork-server.ts`). Native `EventSource` can't send custom headers, and passing the token as a query param is a non-starter (leaks into logs/history/referrer). Implementation must use a `fetch()`-based readable stream reader instead  which also means **losing `EventSource`'s built-in auto-reconnect**; reconnect-with-backoff and `Last-Event-ID` tracking need to be hand-rolled.
- **Server idle timeout will kill long-lived connections.** `server.ts` sets `idleTimeout: 120s`, mirrored in `serve-node.ts`'s `keepAliveTimeout`. Any SSE stream idle for 2 minutes gets silently dropped by the Node HTTP layer. Requires either server-sent heartbeat comments (`:\n\n`) every ~30s, or a separate timeout policy for streaming routes.
- **No pub/sub layer exists yet.** `ReloadEventStore` and `SessionGroupEventStore` (`events.ts`, `session-groups.ts`) are in-memory ring buffers with no subscribe/notify mechanism. This needs to be built, including per-workspace connection tracking and cleanup on client disconnect, before any streaming can happen  this is new infrastructure, not a refactor of existing code.
- **Desktop/Tauri routing dependency.** `openwork-server.ts` already has an `OPENWORK_STREAM_URL_RE` regex that routes `/events`-matching URLs to `window.fetch` instead of Tauri's IPC-based `desktopFetch`, because Tauri's fetch blocks until the body closes (which would freeze the desktop webview on an infinite stream). `/events` and `session-groups/events` both appear to match this regex already, but this needs explicit testing confirmation, not just static analysis  verify in a real Tauri build, not just reading the regex.

VisheshDDev:
### Summary
Fixes #3126

Replaces `useMcpConnectedCount` interval polling with real-time `mcp.*` and `server.*` event listeners via `GlobalSDKProvider`.

### Testing
* Verified with unit test in `use-mcp-connected-count.test.tsx` (`bun test`).

### DIFF
--- a/apps/app/src/react-app/domains/connections/use-mcp-connected-count.test.tsx
+++ b/apps/app/src/react-app/domains/connections/use-mcp-connected-count.test.tsx
@@ -1,0 +1,27 @@
+import React from "react";
+import { renderToString } from "react-dom/server";
+import { useMcpConnectedCount } from "./use-mcp-connected-count";
+
+// Declare Bun test globals so TypeScript doesn't throw module errors
+declare const describe: (name: string, fn: () => void) => void;
+declare const test: (name: string, fn: () => void) => void;
+declare const expect: (actual: any) => { toContain: (expected: string) => void };
+
+function TestComponent({ client, directory }: { client: any; directory: string }) {
+  const count = useMcpConnectedCount(client, directory);
+  return React.createElement("div", { "data-count": count }, count);
+}
+
+describe("useMcpConnectedCount", () => {
+  test("returns 0 when client or directory is missing", () => {
+    const htmlWithNullClient = renderToString(
+      React.createElement(TestComponent, { client: null, directory: "/test/dir" })
+    );
+    expect(htmlWithNullClient).toContain('data-count="0"');
+
+    const htmlWithEmptyDir = renderToString(
+      React.createElement(TestComponent, { client: {} as any, directory: "" })
+    );
+    expect(htmlWithEmptyDir).toContain('data-count="0"');
+  });
+});
--- a/apps/app/src/react-app/domains/connections/use-mcp-connected-count.ts
+++ b/apps/app/src/react-app/domains/connections/use-mcp-connected-count.ts
@@ -2,16 +2,23 @@ import { useEffect, useState } from "react";
 
 import { unwrap } from "../../../app/lib/opencode";
 import type { Client, McpStatusMap } from "../../../app/types";
-
-const REFRESH_INTERVAL_MS = 60_000;
+import { useGlobalSDK } from "../../kernel/global-sdk-provider";
 
 /**
- * Live count of connected MCP servers for the active workspace, polled from
- * opencode's mcp.status. Used by the session status bar so it reflects real
- * connectivity instead of a hardcoded value.
+ * Live count of connected MCP servers for the active workspace, driven by real-time
+ * Server-Sent Events (SSE) from the global SDK stream instead of polling on an interval.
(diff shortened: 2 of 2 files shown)


VisheshDDev:
Hey @benjaminr-bi! Closed the previous PR to clean up the commit history—this fresh PR contains only 1 clean commit on top of dev for #3126. Whenever you have time to take a look, no rush!

Pallavikumarimdb:
@VisheshDDev  Thanks for the PR! 
However, this PR replaces the 60s polling on `useMcpConnectedCount` rather than the sub-second polling on the actual target endpoints `/workspace/:id/events` and `/workspace/:id/session-groups/events` (defined in `use-engine-reload.ts` and `use-session-group-sync.ts`). 
Because of this, the core issue of high-frequency round trips on those two endpoints remains unresolved and does not satisfy the fix of the issue  #3126

VisheshDDev:
> @VisheshDDev Thanks for the PR! However, this PR replaces the 60s polling on `useMcpConnectedCount` rather than the sub-second polling on the actual target endpoints `/workspace/:id/events` and `/workspace/:id/session-groups/events` (defined in `use-engine-reload.ts` and `use-session-group-sync.ts`). Because of this, the core issue of high-frequency round trips on those two endpoints remains unresolved and does not satisfy the fix of the issue #3126

I'm working on #3126. Before implementing the SSE routes, I'd like to confirm the intended architecture. My current plan is:

add a subscription API to ReloadEventStore and SessionGroupEventStore
use that pub/sub layer to back SSE endpoints
replace polling in the React client with fetch-based SSE + reconnect

Does this align with the intended implementation?

PR #3352, opened 2026-07-30
fix(desktop): prevent renderer crash on windows during app exit (#2908)

### LINKED ISSUE #2908: [Bug]: Windows: hard `throw` in getComputerUseMcpCommand() crashes desktop renderer (blank screen) on 0.17.30 and 0.17.32
### Summary

**Environment**
- OS: Windows 11
- OpenWork versions tested: 0.17.30, 0.17.32 (both affected identically)
- Install path: %LOCALAPPDATA%\Programs\@openworkdesktop\
- Config: %APPDATA%\openwork\, ~/.config/opencode/opencode.json

**Symptom**
App window goes blank/unresponsive on startup (or when opening Settings, depending on build). No log files are written by default; running the .exe from a terminal surfaces:

    Error: OpenWork Computer Use is missing from this OpenWork build.

**Root cause (confirmed via app.asar extraction)**
In `server/dist/extensions/google-workspace.js`:

    function computerUseHelperAppPath() {
      const explicitApp = process.env.OPENWORK_COMPUTER_USE_APP?.trim();
      const candidates = [
        explicitApp,
        process.resourcesPath ? path.join(process.resourcesPath, "helpers", COMPUTER_USE_HELPER_APP_NAME) : null,
        path.resolve(__dirname, "..", "resources", "helpers", COMPUTER_USE_HELPER_APP_NAME),
      ].filter(Boolean);
      return candidates.find((candidate) => existsSync(candidate)) ?? null;
    }

    function getComputerUseMcpCommand() {
      const helperExecutable = computerUseHelperExecutablePath();
      if (helperExecutable) return [helperExecutable, "mcp"];
      if (app.isPackaged) {
        throw new Error("OpenWork Computer Use is missing from this OpenWork build.");
      }
      ...
    }

`COMPUTER_USE_HELPER_APP_NAME = "OpenWork Computer Use.app"` and the executable path is built as `Contents/MacOS/ComputerUse` — this is exclusively a macOS `.app` bundle structure. There is no `win32` branch anywhere in `computerUseHelperAppPath()`/`computerUseHelperExecutablePath()` that resolves a Windows helper path or binary (confirmed — the file has many other `process.platform === "win32"` branches for unrelated concerns like shell selection and auto-update assets, but none for this lookup). So on Windows, `helperExecutable` is always `null`, `app.isPackaged` is `true` for an installed build, and the function throws unconditionally.

This is invoked via IPC channel `getComputerUseMcpCommand` (registered in `server/dist/extensions/openai-image-generation.js`), with no `try/catch` around the call in either file. If the renderer calls this channel during initial mount (e.g. to enumerate available MCP tools) without handling the rejection, the result is an uncaught error that blanks the window.

**Suggested fix**
Return `null` instead of throwing when the helper is unavailable on non-macOS platforms, and have the caller treat `null` as "Computer Use unavailable" rather than propagating an exception. Ideally, gate the whole Computer Use extension registration behind `process.platform === "darwin"` until Windows/Linux support exists, so it never appears as an available tool on unsupported platforms in the first place.

**Confirmed via**: extracting app.asar from the installed build and locating the throw site directly (see code above).

### To Reproduce

**Repro steps**
1. Install OpenWork on Windows (0.17.30 or 0.17.32).
2. Launch normally, or open Settings.
3. Window goes blank / unresponsive.
4. Run the .exe from cmd.exe to see the thrown error in stdout.

### Expected behavior

Window going blank / unresponsive.

### Actual behavior

Window goes blank / unresponsive.

### Screenshots (optional)

_No response_

_No response_

_No response_

VisheshDDev:
Summary
Fixes #2908

On Windows, closing or quitting the desktop app could trigger a native access violation crash when `showShutdownScreen()` attempted to interact with a destroyed renderer window or when active pty terminal sessions emitted IPC messages into destroyed web contents.

### Changes
* Added `win.webContents?.isDestroyed()` guard inside `showShutdownScreen()`.
* Ensured active terminal processes are explicitly killed during `disposeRuntimeBeforeQuit()`.

### Testing
* Verified typecheck with `pnpm --filter @openwork/desktop typecheck:electron`.

### DIFF
--- a/apps/desktop/electron/main.mjs
+++ b/apps/desktop/electron/main.mjs
@@ -1158,7 +1158,7 @@ let runtimeBootstrapPromise = null;
 
 function showShutdownScreen() {
   const win = mainWindow;
-  if (!win || win.isDestroyed()) return;
+  if (!win || win.isDestroyed() || win.webContents?.isDestroyed()) return;
   try {
     win.show();
     win.webContents.loadURL(`data:text/html;charset=utf-8,${encodeURIComponent(`<!doctype html>
@@ -1187,11 +1187,14 @@ function showShutdownScreen() {
     // Ignore renderer teardown races during quit.
   }
 }
-
 async function disposeRuntimeBeforeQuit() {
   if (runtimeDisposedForQuit || runtimeDisposeInProgress) return;
   runtimeDisposeInProgress = true;
   try {
+    // Kill all lingering pty terminal processes before webContents teardown
+    for (const [terminalId] of terminalProcesses.entries()) {
+      killTerminal(terminalId);
+    }
     await runtimeManager.dispose().catch(() => undefined);
     runtimeDisposedForQuit = true;
   } finally {


VisheshDDev:
@benjaminr-bi Submitted the fix for #2908! Prevents the Windows renderer access violation on app exit by guarding destroyed webContents and terminating lingering terminal processes during shutdown.

PR #3371, opened 2026-07-30
revert(app): revert signed-in Den startup gate before Run Task (#3300)

benjaminshafii:
## Summary

- Reverts #3300 (`64786c44d`, "fix(app): wait for signed-in Den startup before Run Task"), restoring the previous Run Task / composer submit behavior.
- Removes the app-level Cloud MCP readiness gate: `cloud-mcp-app-readiness.ts` is deleted and `session-route.tsx`, `session-surface.tsx`, `composer.tsx`, `new-task-composer.tsx`, `session-empty-hero.tsx`, and `use-session-mcp-maintenance.ts` go back to consuming `cloudMcpSubmissionState` directly, with no `onRetryCloudConnection` plumbing.
- One conflict in `apps/app/src/react-app/shell/session-route.tsx` was resolved by hand: a commit landed after #3300 renamed the connect handler, so the newer `onOpenConnect: () => handleOpenExtensions()` is kept (instead of the older `navigate("/settings/extensions")`) and `handleOpenExtensions` stays in the memo dependency list since it is still used by `onOpenSettingsSection`.

### Build verification
- `pnpm --filter @openwork/app build` — passed (vite build succeeded, only the pre-existing chunk-size warnings).
- `pnpm tsc --noEmit` in `apps/app` — passed, clean. This is the check that matters most here, since it confirms the hand-resolved conflict left no dangling references to the reverted symbols (`effectiveCloudMcpSubmissionState`, `cloudMcpAppSubmission*`, `IDLE_CLOUD_MCP_SUBMISSION_GATE_STATE`).

### Test verification
- `pnpm test` in `apps/app` (`bun test --isolate tests/`) crashes the runner with exit 133 on `tests/opencode-stream-timeout.test.ts`. **This is pre-existing, not caused by this revert**: the identical crash reproduces on the base commit `2c558bcff` with the same exit code and same file.
- `bun test tests/opencode-stream-timeout.test.ts` on its own passes 4/4, so the crash only appears under `--isolate` in the full-suite run.
- Every test that ran before the crash point passed on this branch.

### UI verification
No UI verification was performed — I did not boot the app or drive the flow. This is a straight revert of a previously reviewed commit, so the reviewer should confirm the Run Task behavior matches what it was before #3300 landed.

### API verification
No API changes.

## Test instructions

1. `pnpm --filter @openwork/app dev`
2. Open a session while signed in to Den and confirm Run Task and the composer are no longer held/locked by the startup readiness gate.
3. Confirm the composer's "Connect" affordance still routes to Settings > Extensions (this path was kept at its newer `handleOpenExtensions` implementation, not reverted).


OmarMcAdam (approved):

src-opn (approved):

benjaminshafii:
The red `openwork-tests (blacksmith-4vcpu-ubuntu-2204)` check here isn't this PR's fault — the install died compiling `better-sqlite3` from source because 11.10.0 ships no Node 24 prebuild, and the `node-gyp` header download timed out:

```
prebuild-install warn install No prebuilt binaries found (target=24.13.0 ... platform=linux)
gyp ERR! stack AggregateError [ETIMEDOUT]
```

Fixed in #3378 (bump to `better-sqlite3@^12.11.1`, which has node-v137 prebuilds, plus a pnpm store cache in the tests workflow). Once that merges, merge `dev` in here or re-run the job.

PR #3375, opened 2026-07-31
feat(connections): show real provider logos across the connect flow

benjaminshafii:
## Summary

Providers were identifiable only by a two-letter monogram unless they were one of the three brands with a bundled inline SVG, so the connect dialog read as a wall of grey initials.

- **New logo resolution ladder** (`provider-logo-src.ts`): inline SVG → Simple Icons → favicon → monogram last. `ProviderIcon` walks the candidates and advances on each `img` error, so a dead source degrades silently per row rather than breaking the list. Long-tail catalog ids are slugified domains, so `302ai` is unslugged to `302.ai` to recover a real mark instead of "30".
- **Connect dialog** is grouped into Connected / All providers, with 36px logo tiles. The filtered list is connected-first and headers are inserted at the group boundary, which keeps the existing arrow-key/Enter navigation indexing straight into display order instead of needing a parallel index.
- **API-key step** leads with the provider's logo, name and slug, shows env vars as chips, and has a single footer action bar. The stacked double footer, duplicate Back button, and the "Keys are stored locally by OpenCode" caption are gone.
- **Model picker's API-keys row** is dynamic: with no keys it shows an Anthropic/OpenAI/Google logo cluster and "Add your keys"; once keys exist it shows your own providers and "Connect more providers".

## Test plan

Ran and passing:

- `npx tsc --noEmit` — clean
- `bun test src/react-app/design-system/provider-icon.test.tsx` — 9 pass
- `pnpm build` — clean

I verified the logo sources live rather than trusting the spec: Simple Icons returns 200 for `anthropic`, `googlegemini`, `mistralai`, `deepseek`, `ollama`, `openrouter` but 404s for `openai` and `abacus`; the favicon step returns real images for `302.ai`, `groq.com`, `abacus.ai`. Those misses are encoded so they skip straight to the favicon step.

**Not yet verified visually.** I have not driven this dialog in a running app — it needs a backend to populate the provider list, so I have not seen it render a real grid of logos end to end. Reviewer repro: open Settings → Connect, confirm providers show brand marks rather than initials, and that a provider with no resolvable logo still falls back cleanly.

### DIFF
--- a/apps/app/src/components/model-select.tsx
+++ b/apps/app/src/components/model-select.tsx
@@ -47,6 +47,13 @@ import {
 import { openModelPickerEvent, openProviderAuthEvent } from "@/react-app/shell/new-providers-listener";
 import { newProvidersEvent } from "@/app/lib/provider-events";
 
+/** Shown with their logos when no keys are connected yet. */
+const SUGGESTED_KEY_PROVIDERS = [
+  { id: "anthropic", name: "Anthropic" },
+  { id: "openai", name: "OpenAI" },
+  { id: "google", name: "Google" },
+];
+
 function getProviderDisplayName(providerId: string) {
   return providerId
     .split("-")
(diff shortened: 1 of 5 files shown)


benjaminshafii:
## Verification

Mounted the real `ProviderAuthModal` in a browser through the app's Vite pipeline (real React, real network) rather than a static render, so the logo resolution ladder actually executes.

**Provider list**
- Every provider resolved a real brand mark: Anthropic, OpenRouter, OpenAI, 302.AI, Abacus.AI, Cohere, DeepSeek, Google, Groq, Hugging Face, Mistral.
- The ladder was exercised end to end: `cdn.simpleicons.org` returned 404 for `302ai` and `abacus`, both fell through to the favicon step and rendered correctly. A deliberately unknown vendor exhausted every step and fell back to a monogram.
- Connected providers sort into a `CONNECTED` group ahead of `ALL PROVIDERS`, with green Connected badges and OAuth/API key method chips.

**API-key step**
- Header shows provider logo + name + id.
- Env vars render as a chip (`ABACUS_API_KEY`).
- "Keys are stored locally by OpenCode." is gone.
- Single action bar: Back on the left, Close + Save key on the right.

Also run: typecheck, `bun test`, and a production build — all green.

One note: the shared `inset-s-*` logical utilities used by `components/ui/dialog.tsx` don't resolve in a bare harness page, so dialog centering had to be overridden inline for screenshots. That's pre-existing shared-primitive behavior, untouched by this PR.

src-opn (approved):

benjaminshafii:
@OmarMcAdam @src-opn re-requesting a look, please.

After your approval the single commit on this branch was **rewritten to be signed** — the default-branch ruleset requires verified signatures and the original commit was unsigned. The new commit was created through GitHub's API so it carries a verified signature, and the PR then auto-merged.

The rewrite was content-preserving: `git diff` between the originally-approved SHA and the merged tree returns zero lines, so what landed is byte-identical to what you approved. The approval carried over rather than being dismissed, which is why it merged without a second review.

Flagging it explicitly so the re-review is a conscious sign-off rather than an inherited one.

PR #3387, opened 2026-07-31
feat(evals): make surface placement explicit, and own browser disposal

benjaminshafii:
## Why

`resolveHost()` reads one ambient env var (`OPENWORK_EVAL_DAYTONA_SANDBOX`), so a whole process got one host: everything local, or everything in one sandbox. `desktop()` called it internally and `DesktopOptions` had no host field, so a spec could not express "app here, browser there", or two desktops in two sandboxes. Topology was inherited from the environment rather than declared — the same shape of problem as env-var-gated skips.

## What changed

- `desktop({ host })` and a new `chrome({ host })` accept a placement.
- `localHost()` / `daytonaSandbox(id)` name one, keeping `REPO_ROOT` wiring inside the package. `daytonaSandbox()` throws from *inside* a sandbox, where the CLI indirection has nothing to target.
- `resolveHost()` default behaviour is unchanged, so every existing spec is untouched.

```ts
await using sandboxA = daytonaSandbox("openwork-test-a");
await using app     = await desktop({ host: sandboxA });
await using browser = await chrome({ host: daytonaSandbox("openwork-test-b") });
```

## Bug fixed along the way

`attachSurface` only closes the CDP socket — it never kills the process. `first-run-cloud-share` hand-rolled `spawnChrome` + `attachSurface`, so `host.disposeSurface` was never reached and a Chrome process survived every run. `chrome()` owns disposal: stop the client, then dispose the surface.

## Proof (Daytona sandbox, real den)

| Spec | Result | Exercises |
|---|---|---|
| `first-run-cloud-share` | **PASS x3** — 47.0s / 35.7s / 41.0s | the new `chrome()` seam |
| `app-smoke` | **PASS x2** — 51.8s / 51.7s | `desktop()` default path, unchanged |

Process check before and after all runs, using `ps -eo comm=` (self-match-proof, unlike `pgrep -f`):

```
BASELINE:          none
AFTER-ALL-RUNS:    none
```

`tsc -p evals` clean; `pnpm --dir evals test` 95 pass / 0 fail.

## Limits, stated

- **A split topology is not proven end-to-end.** It typechecks and the seam is exercised, but putting the app and browser in *different* sandboxes needs two sandboxes and a laptop-side vitest run (`OPENWORK_EVAL_DAYTONA_SANDBOX` mode). Not done here.
- **No capability model yet.** `Host.kind` is still a bare string; `requires([...])` and honest skip reasons are the follow-up this unblocks.
- No `fraimz.html`: the spec lane writes `evals/results/rolls/<stamp>-<name>/index.html`. Frames for these runs are validated there. Worth reconciling with AGENTS.md separately.

## Unrelated finding

`org-connection-lifecycle` fails on this sandbox: DOM waits pass (`NEEDS YOUR SIGN-IN`, `OAuth required` present in `body.innerText`) but the vision check reports the Extensions panel blank. Not touched by this PR and not in #3359's green list either, so it may never have passed. Needs its own investigation — either a reveal/scroll issue before the frame, or a real rendering defect.


src-opn (dismissed):

benjaminshafii:
## Conflict resolved + re-verified after merging `dev`

The conflict was squash-merge divergence: this branch carried all 50 commits from `fix/app-spec-den-bootstrap` while `dev` has them as one squash (`915c43d18`). Only `c830ab530` is actually new here.

Merged `origin/dev` in (`d6431becc`). Exactly one file conflicted — `first-run-cloud-share.slow.test.ts`, add/add, because the merge base predates the file. The two differing hunks were precisely this PR's intentional change vs dev's pre-change version, so taking this branch's side loses nothing.

Re-verified on the merged tree, which pulls in dev's app changes that plausibly touch these paths (`Revert "wait for signed-in Den startup before Run Task"` #3371, `gateway checkpoint ladder` #3376, `provider logos in the connect flow` #3375), plus a dependabot lockfile bump requiring a fresh `pnpm install`:

| Spec | Result | Exercises |
|---|---|---|
| `app-smoke` | **PASS** — 52.3s | `desktop()` default host path |
| `first-run-cloud-share` | **PASS** — 41.9s | the new `chrome()` seam |

Process check after both runs (`ps -eo comm=`): `none`. `tsc -p evals` clean, `pnpm --dir evals test` 95 pass / 0 fail on the merged tree.

src-opn (approved):

PR #3388, opened 2026-07-31
fix(evals): make specs placement-portable — host workspace root, sandbox-aware ports, profile pruning

benjaminshafii:
> ## Status update (2026-08-01)
> Everything below the rule was written when mode B first ran; the PR has since grown the missing pieces and they are proven:
> - **The two-member connector spec IS now in this PR** (`org-connector-two-members.slow.test.ts`) and is **green twice on two separate Daytona sandboxes** — 133.0s / 135.2s, driver outside all sandboxes. Both members OAuth-connect their own account and call the connector's tool from the composer; the mock witnesses two distinct bearer fingerprints. Full evidence: [the run report](https://github.com/different-ai/openwork/pull/3388#issuecomment-5150725360).
> - Mode B is green (`two-daytona-desktops.slow.test.ts`, plus the connector runs above). The renderer-over-proxy hang was renderer JS starvation (workspace scan + full disk), not the proxy — fixed by the /tmp workspaces and profile pruning already in this PR.
> - Of the "What's left" table: #1 resolved via `startMockMcp({ publicUrl })` (the seam existed but was dead code — health-check and stale-log bugs fixed here); #2 unnecessary (the eval image's real browser completes the OAuth hop on-sandbox); #3 landed (args + per-credential `tokenId`, asserted by the spec); the model knob exists (`OPENWORK_EVAL_MODEL`); CA trust/TLS not needed (Daytona preview URLs are real HTTPS).
> - The earlier "blocked on a product defect (#3375)" claim is retracted in the spec header itself; the spec stays opt-in because it needs provisioned placement (two sandboxes, reachable Den, published mock), which nightly does not have.

---

Groundwork for the two-member connector spec (two Electrons in two sandboxes, den + connector reachable from both). I set out to write that spec; driving mode B for the first time surfaced three real bugs that had to land first. The spec itself is **not** in this PR — see "What's left".

## Three bugs, each found by running

**1. Specs assumed the driver's filesystem is the app's.** `app-smoke` passed `process.cwd()` as a workspace path. Correct only when driver and app share a disk. Driving a sandbox from a laptop, the app was told to open a directory that exists only on the laptop — the symptom was onboarding hanging 120s on "Power your first task" with no error, reading like a broken app.

`Host.workspaceRoot` now states it (`repoRoot` locally, `/workspace` on Daytona) and `DesktopHandle` re-exposes it, for specs that genuinely need the repo (`skills-local` reads `.opencode/skills`).

**2. Daytona surface ports were allocated from a local counter.** The local host finds a free port by binding one; the Daytona host cannot, since the port must be free on another machine. It guessed from `{ primary: 9825, next: 9830 }` — and the OpenCode sidecar, which picks its own port at boot, was observed holding **9825**, the CDP primary. Electron's debugger never bound; the preview URL timed out after 180s naming nothing. Now it asks the sandbox which ports are listening and skips them.

**3. Stale surface profiles filled the disk.** Disposal removes a profile on the happy path; every killed or failed run leaks one at ~50MB. **27 of them took a 10GB sandbox to 99% full.** The symptom was not "disk full" — it was the renderer failing `Runtime.evaluate` for 240s. That cost four debug cycles. `clearStaleSurfaces` already pkills by `rootDir` and so already assumes exclusive ownership; it now deletes the directories it orphans.

## A regression I caught and reverted

Routing `app-smoke` through `workspaceRoot` made it open the **whole monorepo** (`/workspace`, `node_modules` included) instead of the smaller `/workspace/evals` that `process.cwd()` happened to give. Engine scan blocked the renderer past 240s. It now uses a fresh `/tmp/openwork-app-smoke-<stamp>` — small, and valid on any host. Only caught because I re-ran **mode A** after each change; worth keeping that habit.

## Proof

Mode A (in-sandbox), after the disk was freed:

| Spec | Result |
|---|---|
| `app-smoke` | **PASS** — 59.1s, then 58.3s |
| `skills-local` | **PASS** — 79.1s |

Pruning verified: `profiles-left=1` (the current run only) where 27 had accumulated; disk steady at 84%.

`tsc -p evals` clean. `pnpm --dir evals test` 95 pass / 0 fail. Three `Host` fakes in runner tests needed `workspaceRoot` — the type system caught them, which is the seam working.

## Mode B status, honestly

Mode B (vitest on a laptop, Electron in a sandbox, CDP over a preview URL) went from **hanging at 120s** to **completing workspace creation in 68.5s**. It is not yet green end to end. Two things remain, both documented rather than fixed:

- **The driver needs the vision key.** `validate()` runs driver-side, so `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` must be on the driver — `daytona-spec.sh` sources `/daytona-secrets/*.env` *inside* the sandbox, which does nothing for mode B. Same family as bugs 1 and 3: something moved sides.
- **Renderer responsiveness over the proxy** during engine boot still needs work.

## What's left for the connector spec

| # | Gap |
|---|---|
| 1 | `startMockMcp({ host })` — labs are local-only. `publicUrl` already exists as an attach hatch, so a generic `Host.startService()` is the clean shape. |
| 2 | `captureOpenedUrls({ host })` — the `xdg-open` shim lands on the driver, not where Electron runs |
| 3 | Mock MCP call log: record tool **arguments** and token identity (it already logs `toolNames` + `authorized`) |
| 4–7 | CA trust, TLS terminator, `requires([...])` capabilities, model knob |

Note this spec **cannot run in nightly** — nightly is `blacksmith-4vcpu-ubuntu-2204` with no Daytona access. That is why the capability model matters: without it, it would skip silently and read green.

### DIFF
--- a/evals/packages/behaviors/src/den.ts
+++ b/evals/packages/behaviors/src/den.ts
@@ -82,6 +82,7 @@ export async function ensureMemberSession(
   } catch {
     // Bootstrap the missing member through the real invitation flow.
   }
+  let markVerifiedWarning = "";
   const invite = await denFetch(den, "/v1/invitations", {
     method: "POST",
     headers: auth(admin),
@@ -98,8 +99,21 @@ export async function ensureMemberSession(
   if (!signUp.response.ok) {
     throw new Error(`Member sign-up failed: HTTP ${signUp.response.status} ${preview(signUp.body)}`);
   }
-  doInternalMarkEmailVerified(input.markVerifiedCmd ?? "", input.email);
-  const member = await signIn(den, input);
+  // Best effort: verification is a convenience for dens that require it. If the
+  // helper is absent or fails (its DB credentials drift), the sign-in below is
+  // the real test — failing here would hide a member who can already sign in.
+  try {
+    doInternalMarkEmailVerified(input.markVerifiedCmd ?? "", input.email);
+  } catch (error) {
+    markVerifiedWarning = error instanceof Error ? error.message : String(error);
+  }
+  const member = await signIn(den, input).catch((signInError: unknown) => {
+    const detail = signInError instanceof Error ? signInError.message : String(signInError);
+    throw new Error(
+      `${input.email} could not sign in after sign-up: ${detail}`
+      + (markVerifiedWarning ? `\nEmail verification also failed first: ${markVerifiedWarning}` : ""),
+    );
+  });
   const accept = await denFetch(den, "/v1/orgs/invitations/accept", {
     method: "POST",
     headers: auth(member),
--- a/evals/packages/behaviors/src/desktop.ts
+++ b/evals/packages/behaviors/src/desktop.ts
@@ -157,7 +157,14 @@ export async function fill(app: Surface, selector: string, value: string, opts: 
 
 export async function go(app: Surface, hashPath: string): Promise<void> {
   const hash = hashPath.startsWith("#") ? hashPath : `#${hashPath}`;
-  await evalIn(app, `(() => { window.location.hash = ${jsValue(hash)}; return true; })()`);
+  // Setting the hash is idempotent, so retry through renderer freeze bursts
+  // rather than letting one blocked evaluate fail a whole spec. Contention (two
+  // desktops on one host) makes those bursts routine, and a bare 20s evaluate
+  // here was the single most common way a long journey died near its end.
+  await waitFor(app, `(() => { window.location.hash = ${jsValue(hash)}; return true; })()`, {
+    timeoutMs: 60_000,
+    label: `navigate to ${hash}`,
+  });
 }
 
 export async function currentHash(app: Surface): Promise<string> {
--- a/evals/packages/behaviors/src/desktop.ts
+++ b/evals/packages/behaviors/src/desktop.ts
@@ -161,8 +161,11 @@ export async function go(app: Surface, hashPath: string): Promise<void> {
   // rather than letting one blocked evaluate fail a whole spec. Contention (two
   // desktops on one host) makes those bursts routine, and a bare 20s evaluate
   // here was the single most common way a long journey died near its end.
+  // 120s: with several desktops and their engines on one host the renderer can
+  // stay blocked well past a minute. Retrying an idempotent hash assignment
+  // costs nothing when the app is healthy.
   await waitFor(app, `(() => { window.location.hash = ${jsValue(hash)}; return true; })()`, {
-    timeoutMs: 60_000,
+    timeoutMs: 120_000,
     label: `navigate to ${hash}`,
   });
 }
--- a/evals/packages/cdp/src/cdp.ts
+++ b/evals/packages/cdp/src/cdp.ts
@@ -187,10 +187,17 @@ export function connect(
           const id = nextId;
           nextId += 1;
           return new Promise((innerResolve, innerReject) => {
+            // Name WHAT timed out. The rejection fires from a timer, so the
+            // caller's stack is gone by then: "CDP call Runtime.evaluate timed
+            // out" alone cannot tell you which of a spec's dozens of
+            // evaluations blocked, which turns a one-line fix into a hunt.
+            const subject = typeof params.expression === "string"
+              ? ` evaluating: ${params.expression.replace(/\s+/g, " ").trim().slice(0, 160)}`
+              : "";
             const timer = timeoutMs > 0
               ? setTimeout(() => {
                 pending.delete(id);
-                innerReject(new Error(`CDP call ${method} timed out after ${timeoutMs}ms.`));
+                innerReject(new Error(`CDP call ${method} timed out after ${timeoutMs}ms.${subject}`));
               }, timeoutMs)
               : null;
             pending.set(id, { resolve: innerResolve, reject: innerReject, timer });
--- a/evals/packages/hosts/src/daytona.ts
+++ b/evals/packages/hosts/src/daytona.ts
@@ -318,6 +318,40 @@ function portSet(values: number[] | undefined): Set<number> {
   return ports;
 }
 
+/**
+ * Ports already listening INSIDE the sandbox.
+ *
+ * The local host finds a free port by binding one; this host cannot, because the
+ * port has to be free on a different machine. Allocating from a local counter
+ * alone silently collides: the OpenCode sidecar picks its own port at boot and
+ * was observed holding 9825 — the CDP primary — so Electron's debugger never
+ * bound and the preview URL timed out after 180s with no hint of the cause.
+ */
+async function sandboxListeningPorts(exec: DaytonaExec, sandbox: string): Promise<Set<number>> {
+  const result = await exec(["exec", sandbox, "--", "bash -lc 'ss -ltn 2>/dev/null || netstat -ltn 2>/dev/null'"], { timeoutMs: 30_000 });
+  const ports = new Set<number>();
+  if (result.code !== 0) return ports;
+  for (const line of result.stdout.split(/\r?\n/)) {
+    // Match the local-address column's :PORT, e.g. "127.0.0.1:9825" or "*:3005".
+    const match = /[:.](\d{2,5})\s+\S+\s*$/.exec(line.trim()) ?? /[:.](\d{2,5})\s/.exec(line.trim());
+    if (!match) continue;
+    const port = Number.parseInt(match[1] ?? "", 10);
+    if (Number.isInteger(port) && port > 0 && port <= 65_535) ports.add(port);
+  }
+  return ports;
+}
+
+/** Allocate a port that is free in the sandbox, not merely unused by this host. */
+async function allocateSandboxPort(allocation: PortAllocation, exec: DaytonaExec, sandbox: string): Promise<number> {
+  const taken = await sandboxListeningPorts(exec, sandbox);
+  for (let attempt = 0; attempt < 64; attempt += 1) {
+    const port = allocatePort(allocation);
+    if (!taken.has(port)) return port;
+    // Keep it marked used so we never hand it out again this session.
+  }
+  throw new Error(`Could not find a port free inside sandbox ${sandbox} after 64 attempts.`);
+}
+
 function appendExtraEnv(assignments: Map<string, string>, env: Record<string, string> | undefined): void {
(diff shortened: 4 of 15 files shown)


benjaminshafii:
## Update: connector spec written; blocked by a product regression it caught

Continued past the groundwork. The two-member connector spec now exists (`evals/specs/org-connector-two-members.slow.test.ts`) and everything up to the org connections surface is verified. The surface itself is broken on current `dev`.

### Also landed

**Mock MCP is now a witness.** It records tool **arguments** and a sha256 fingerprint of the caller's bearer token, exposed as `conn.toolCalls({ name, atLeast, timeoutMs })`. That makes per-member isolation provable from the connector rather than from the app's own "Connected" text — distinct `tokenId`s are the assertion.

**Email verification is best effort.** `ensureMemberSession` aborted the whole spec at 2s when `mark-verified.sh` failed — it uses `-uroot` with no password while den's MariaDB root has one. The member had already been created and could sign in. Now it warns and lets the following sign-in be the test, reporting both causes if that fails too.

Three spec-side bugs found and fixed by running:
- the card poll watched the **session** surface, because the app opens a freshly created session on its own and navigates away from settings — the wait now steers back, as a person would
- the connections route needed a **settled** wait plus the `Refresh` click the proven spec uses
- bare 20s `Runtime.evaluate` probes died on renderer freeze bursts, which two contending desktops make routine — probes are now short and retry-tolerant

### The blocker

`/settings/extensions/connections` renders **empty** for a signed-in org member: correct route, `document.body.innerText === ""`.

This is not the new spec. **`org-connection-lifecycle`, untouched by this branch, regressed the same way today** — it got *past* `waitForConnectionCard` earlier and now fails "Connection card did not render". Between those runs the sandbox pulled merged `dev`, which includes **#3375 "show real provider logos across the connect flow"** — that surface. Worth bisecting there first.

The new spec is therefore **opt-in** (`OPENWORK_EVAL_CONNECTOR_SPEC=1`) so a product defect cannot masquerade as a nightly failure. Remove the gate when the surface renders; the assertions are already written.

### Proof on this branch

| Spec | Result |
|---|---|
| `app-smoke` | **PASS** |
| `skills-local` | **PASS** |
| `first-run-cloud-share` | **PASS** |
| `org-connector-two-members` | **SKIPPED** (gated on the defect above) |

`tsc -p evals` clean, `pnpm --dir evals test` 95 pass / 0 fail.

### Still open

Placement is one line away — both desktops use the ambient host; `desktop({ host: daytonaSandbox(id) })` moves either onto its own sandbox once **mode B** is green (driver needs the vision key; renderer responsiveness over the preview-URL CDP needs work). TLS/CA and the capability model remain as described in the PR body.

benjaminshafii:
## Status: 26 of ~29 steps green, including the headline assertion

### I was wrong about the product regression — retracting it

I claimed `/settings/extensions/connections` was broken by #3375. **It is not.** A probe found the surface renders fine: 31 buttons, 557 chars of text, **zero errors, zero console errors**.

The real cause was mine. **The app canonicalises the route** — `#/workspace/<id>/settings/extensions/connections` becomes `#/workspace/<id>/extensions/connections`. My steer-back loop checked for the *pre-rewrite* form, so it re-navigated every iteration, fighting the rewrite until the surface never settled. That produced the blank page I blamed on dev. #3375 is exonerated; nobody should bisect it on my account.

### What now passes

Everything through member B connecting their own account:

- arrangement: connector, two members, org connection created and readable per-member over the API
- member A: desktop boots, signs in, connections surface, card renders, `NEEDS YOUR SIGN-IN`, detail opens, `OAuth required`
- member A connects: authorize observed at the connector, `Connected with your own account.`, API flips to `connectedForMe: true`
- **member B still shows `NEEDS YOUR SIGN-IN` while A is connected — the per-member isolation assertion, the whole reason two desktops exist**
- member B connects their own account
- three vision frames pass

### What is still blocked, and why it argues for the original design

The tool-call phase (steps 27–29) never starts: `go(.../session)` cannot execute because appA's renderer is wedged. Not a burst — it survived a 15s probe retried across 120s.

**This is resource exhaustion, not product behaviour.** Two Electrons, two OpenCode engines, den and MySQL on one 9GB sandbox. Which is an argument *for* the two-sandbox topology this branch was groundwork for: one desktop and one engine per sandbox removes the ceiling entirely.

Cheapest next step without more hardware: reorder so only one desktop is heavily active — A connects → A's tool call → *then* create B's desktop → B sees `NEEDS YOUR SIGN-IN` (isolation still valid, B is created after A connected) → B connects → B's tool call.

### Resilience fixes landed while getting here

- **CDP timeouts now name the expression.** The rejection fires from a timer, so the caller's stack is gone; `CDP call Runtime.evaluate timed out` alone can't say which of dozens of evaluations blocked. This one change turned a guessing game into a one-line fix twice.
- **`go()` retries.** Setting a hash is idempotent; a bare 20s evaluate was the most common way a long journey died near its end.
- **Frames wait for paint, not DOM.** `waitForText` proves the DOM; the detail panel paints slightly later, so the vision check passed one run and failed the next.
- **Reveal matches rendered text.** `innerText` is render-aware, so a badge styled `text-transform: uppercase` reads `NEEDS YOUR SIGN-IN` while `textContent` holds `Needs your sign-in` — the reveal could never agree with the wait.

The spec stays opt-in (`OPENWORK_EVAL_CONNECTOR_SPEC=1`) until the tool-call phase runs. `app-smoke`, `skills-local`, and `first-run-cloud-share` remain green on this branch.

benjaminshafii:
## org-connector-two-members: green on two Daytona sandboxes — twice

The last opt-in spec now completes its tool-call phase. The fix was placement, as predicted: each member's desktop gets its own sandbox via `desktop({ host: daytonaSandbox(id) })`, so two Electrons + two engines never contend for one 9GB box.

**Runs (driver on a Mac outside all sandboxes, mode B):**

| run | result | duration |
|---|---|---|
| 1 | green | 133.2s |
| 2 (final code) | green | 133.0s |
| 3 (final code) | green | 135.2s |

```
✓ specs/org-connector-two-members.slow.test.ts  two members each connect their own
  account to one org connector and call its tools  133042ms / 135211ms
```

**Topology:** Den server sandbox (`test-server-on-daytona.sh`, seeded demo org) + mock OAuth MCP on that sandbox with `ISSUER=<public preview URL>` + desktop A and desktop B on separate `openwork-eval-vnc` sandboxes. Env: `OPENWORK_EVAL_DAYTONA_SANDBOX_A/_B`, `OPENWORK_EVAL_CONNECTOR_MOCK_PUBLIC_URL`, `OPENWORK_EVAL_MODEL=big-pickle` (signed-in org desktops only offer the org's Zen models — local `openai/*` is filtered by desktop policy).

**The connector's own witness log** (the authority, not the UI): each run served both markers under distinct bearer fingerprints —

```
08:52:19 mock_echo {"text":"a-1785574331698"} token: a7d2f7da4e15
08:52:26 mock_echo {"text":"b-1785574334716"} token: 4787b0fe26b7
08:54:52 mock_echo {"text":"a-1785574483795"} token: 952a847864c1
08:54:57 mock_echo {"text":"b-1785574486615"} token: 8744fda9ad6e
```

**Two real bugs found by running it:**
1. `startMockMcp({ publicUrl })` was a dead seam — `child?.exitCode !== null` on the null child declared the never-spawned server "exited". Never exercised until now.
2. A long-lived mock's request log satisfied `toolCalls({ atLeast: 2 })` with the *previous* run's calls. `toolCalls` now takes `sinceIso` (and each call carries the mock's timestamp), which also protects the token-distinctness assertion from cross-run contamination.

**Vision-validated frames** (run 2, `evals/results/rolls/2026-08-01T08-50-30-689Z-org-connector-two-members/`):

| member A must connect | member A connected | member B NOT inherited | member B's tool call ran |
|---|---|---|---|
|  |  |  |  |

**Also ran:** `pnpm exec tsc -p evals` clean; evals unit tests 95 pass / 0 fail.

The spec header's earlier "blocked on a product defect (#3375)" claim is now retracted in the code itself; the honest reason it stays opt-in is that it needs provisioned placement (two sandboxes + a reachable Den + a published mock), not that anything product-side is broken.

Repro:
```
OPENWORK_EVAL_APP_SPECS=1 OPENWORK_EVAL_CONNECTOR_SPEC=1 \
OPENWORK_EVAL_DEN_API_URL=<den api url> OPENWORK_EVAL_DEN_WEB_URL=<den web url> \
OPENWORK_EVAL_DAYTONA_SANDBOX_A=<id-a> OPENWORK_EVAL_DAYTONA_SANDBOX_B=<id-b> \
OPENWORK_EVAL_CONNECTOR_MOCK_PUBLIC_URL=<mock url> OPENWORK_EVAL_MODEL=big-pickle \
pnpm --dir evals exec vitest run --config vitest.config.ts --project nightly specs/org-connector-two-members.slow.test.ts
```

benjaminshafii:
Addendum: CodeQL flagged a polynomial-ReDoS in the mock's bearer-header regex (`js/polynomial-redos`, high) — replaced with a shared linear-time parse in 2b62df0a1. Redeployed the mock at that tip and re-ran the spec end to end: **green, 127.9s**, fresh witness log:

```
09:04:51 mock_echo {"text":"a-1785575083225"} token: 187d39be4aac
09:04:57 mock_echo {"text":"b-1785575086025"} token: 78c0feb69a8c
```

CodeQL, openwork-tests (ubuntu + macos), and both Analyze jobs now pass on the tip. That makes four greens for this spec today (133.2s, 133.0s, 135.2s, 127.9s).

benjaminshafii:
@OmarMcAdam @src-opn — requesting your approval; auto-merge (squash) is armed, so one approval ships it.

30-second review guide:
- **Scope:** evals-only + `scripts/mock-oauth-mcp-server.mjs`. No app/server/Den code.
- **What it does:** makes eval specs placement-portable (host-owned workspace roots, sandbox-aware ports, profile pruning) and lands the two-member org-connector spec, proven green 4x with each member's desktop on its own Daytona sandbox — evidence with frames and the connector's own token-fingerprint log is in [this comment](https://github.com/different-ai/openwork/pull/3388#issuecomment-5150725360).
- **Risk:** nightly/CI unaffected — the new spec self-skips without its opt-in env (nightly has no Daytona). All checks green, incl. CodeQL after fixing a ReDoS it caught in the mock.

src-opn (approved):

Issue #3372, opened 2026-07-31
Settings page renders blank/white when a local MCP server uses string `command` (`config.command?.some is not a function`)


a743110094:
### Summary

Opening **Settings** in the desktop app crashes with `TypeError: Jn.config.command?.some is not a function` and renders a blank/white page whenever a configured **local** MCP server uses the string form of `command` (Claude Code style: `command: "python3"` + separate `args`). The settings UI assumes `config.command` is always an array.

### Environment

- OpenWork **0.18.12** (macOS arm64)
- Electron 35.7.5 / Chrome 134

### Steps to reproduce

1. In `~/.config/opencode/opencode.json`, configure a local MCP server with a string command + separate args:

   ```jsonc
   {
     "mcp": {
       "servers": {
         "command": "python3",
         "args": ["${OPENCODE_PLUGIN_ROOT}/mcp_servers/browser/server.py"],
         "enabled": true,
         "type": "local"
       }
     }
   }
   ```

2. Launch OpenWork, open **Settings** (the MCP/Providers section).
3. The page goes completely blank/white.

### Expected

Settings page renders; a local MCP server whose `command` is a string is shown correctly (or normalized before use).

### Actual

Blank white page. Console shows:

```
TypeError: Jn.config.command?.some is not a function
    at app-C-XsK2US.js:2077:36901
    at Array.some (<anonymous>)
    at app-C-XsK2US.js:2077:36835
    at Object.xs [as useMemo] (index-CRIL1fbt.js:48:56853)
    ...
```

After the crash `document.getElementById('root')` has `0` children — the whole Settings route unmounts.

### Root cause

The Settings MCP view computes MCP statuses in a `useMemo` that does:

```js
mcpServers.some(Jn =>
  Jn.name === "opencode-chrome-devtools" ||
  Jn.config.command?.some(Ei => Ei.includes("chrome-devtools"))
)
```

`?.` only guards `null`/`undefined`, not strings. A string `command` (`"python3"`) has no `.some`, so it throws during render and unmounts the route.

This is **inconsistent with the server side**, which already accepts both forms and normalizes them:

- `server/dist/cloud-plugins.js` → `mcpCommandFromConfig()`: `if (Array.isArray(config.command)) …; else return [command, ...readStringArray(config.args)]`
- `server/dist/validators.js` — local MCP **requires** a command array (`"Local MCP requires command array"`).

So the same config is valid on the server but crashes the renderer.

### Suggested fix

[...]

Issue #3137, opened 2026-07-25
[Bug]: The app is usuable


krixx646:
### Summary

I cannot use open work because of this

<img width="849" height="577" alt="Image" src="https://github.com/user-attachments/assets/f452765c-ca5d-492f-82af-35599813339c" />

### To Reproduce

I opened my existing open work version and saw this error message. 

<img width="849" height="577" alt="Image" src="https://github.com/user-attachments/assets/7316dea5-03b3-4d1a-bc6f-e5191de21cac" />

I manually installed the newest version, followed the prompt, but still had this obstacle

### Expected behavior

It should clearly work.

### Actual behavior

Not working

### Screenshots (optional)

<img width="849" height="577" alt="Image" src="https://github.com/user-attachments/assets/599c52d8-8637-4847-a1ed-69b248635022" />

krixx646:
Same error message, different root cause — worth checking before landing #3134.

I hit `ENOENT: no such file or directory, mkdir '\\?'` on the boot screen, but on my
machine it had nothing to do with Electron returning verbatim `\\?\` paths. I dumped
every path from the running main process and they were all clean:

| call | value |
|---|---|
| `app.getPath('home')` | `C:\Users\<me>` |
| `app.getPath('appData')` | `C:\Users\<me>\AppData\Roaming` |
| `app.getPath('userData')` | `C:\Users\<me>\AppData\Roaming\com.differentai.openwork` |
| `os.homedir()` | `C:\Users\<me>` |

The actual trigger: a registered workspace pointing at a **drive that no longer exists**
(an external drive I had unplugged the day before).

Minimal reproduction, no Electron required:

```js
await fs.mkdir("D:\\youtube", { recursive: true });
// ENOENT: no such file or directory, mkdir '\\?'

Pallavikumarimdb:
During boot, OpenWork checks all registered local workspaces and runs `fs.mkdir(..., { recursive: true })` on their paths. If a workspace is on an unplugged or missing drive, Node/Bun throws that `ENOENT ... mkdir '\\?'` error, crashing the app startup. I've wrapped this workspace check in a try/catch block so the app logs a warning and proceeds to boot normally instead of crashing. I'll open a PR with the fix and test coverage.

krixx646:
Thanks, I will be awaiting the fix, probably on the next update. I would love to contribute to this project. Who do I meet to be joined up?

PR #3402, opened 2026-07-31
fix(app): show a recovery screen instead of a blank window when a render throws

pranshu26:
## Summary
- Add a last-resort React error boundary around the whole app, so a render throw shows a recovery screen with the error and stack instead of a blank window.

## Why
React unmounts the entire tree when a render throws. The renderer has only two boundaries today — `LexicalErrorBoundary` (`composer/editor.tsx:1249`) and the tool-part boundary in `components/chat/message-list.tsx:148` — and nothing above the routes. The comment on that second one already describes this failure mode:

> Tool inputs from streamed or interrupted runs can violate their type contracts (partial/undefined input); without this boundary a single bad part unmounts the entire app (white screen). Seen in production on v0.15.3 via a todowrite part with missing `input.todos`.

The same shape reaches the whole app from route-level render work. #3372 is a live example: a local MCP server configured with a string `command` throws `config.command?.some is not a function` inside a `useMemo` in `settings-route.tsx:1797`, and the user gets a blank window with nothing to report. `composer.tsx:1605` throws the same way and takes out the main chat screen.

This does not fix any individual crash — #3373 is the right fix for #3372 specifically. It changes what a crash *looks like*: an error card with a copyable stack instead of a white screen, which also makes the next one of these reportable.

## Issue
- Related to #3372, #3137 — blank/unusable screens where the underlying throw is invisible to the user. Not claiming to close either; the boundary changes presentation, not the root cause.

## Scope
- `apps/app/src/react-app/shell/app-error-boundary.tsx` (new) — the boundary and its fallback UI.
- `apps/app/src/index.react.tsx` — wrap the tree, inside `StrictMode` so provider failures are caught too.
- `apps/app/tests/app-error-boundary.test.tsx` (new).

## Out of scope
- Fixing the individual crashes (#3372 has #3373 open).
- Global `window.addEventListener("error"/"unhandledrejection")` capture in production. `debug-logger.ts:243,255` registers these but `isEnabled()` returns `false` under `import.meta.env.PROD` unless a localStorage flag is set, so production has no global capture. Separate concern, happy to take it in a follow-up.

### Ran
- `bun test --isolate tests/app-error-boundary.test.tsx`
- `bun test --isolate tests/` (full app suite)

### Result
- pass: 4/4 new tests.
- pass: full app suite 580 pass / 0 fail across 105 files.
- pass: typecheck exits 0.

One note on how the tests are written: `react-dom/server` rethrows instead of running error boundaries, so `renderToStaticMarkup(<Boundary><Throws/></Boundary>)` cannot exercise the catch path — I tried that first and it fails. The tests instead drive the state transition directly (`getDerivedStateFromError` → `render()`), which covers the same code the boundary runs. The actual catching is verified in the running app below.

## Manual verification
1. `pnpm --filter @openwork/app dev`
2. Temporarily rendered a component that throws `config.command?.some is not a function` inside the router, simulating #3372.
3. Loaded the app.

Before: blank window, nothing rendered, no indication of what happened.
After: the recovery screen renders — heading "OpenWork hit an unexpected error", the full stack in a scrollable block (`at Boom (http://localhost:5199/src/index.react.tsx:28:9)` etc.), and working Reload / Copy details buttons. Correct in dark theme.

The throwing component was removed before committing; the diff contains no test scaffolding.

## CI status
- pass: n/a — the test workflow does not run on this PR.
- code-related failures: none observed locally.
- external/env/auth blockers: yes. As a fork PR, `OpenWork Tests` and `i18n Audit` sit in `action_required` pending maintainer approval, and the Vercel checks fail with "Authorization required to deploy". Both are independent of this diff.

## Evidence
- Screenshot of the recovery screen available on request — I can attach it to this thread if useful.

## Risk
Low. The boundary is inert until something throws; the passthrough case is covered by a test asserting the children render byte-identically. Worst case is a cosmetic issue on a screen only reachable when the app has already broken.

Two deliberate choices worth flagging for review:
- It renders plain elements rather than `@/components` primitives, and does **not** call `t()`. `initLocale()` runs before the tree mounts and has itself been a startup failure (#2767), so the screen that reports a crash shouldn't depend on i18n or on component code that may be implicated in the crash. This means the strings are English-only — tell me if you'd rather take the i18n dependency and I'll switch it.
- It's placed inside `StrictMode` but outside `QueryClientProvider`, so provider-level failures are caught too.

## Rollback
Revert the commit — remove the wrapper in `index.react.tsx` and delete the two new files. Nothing else imports the boundary.

### DIFF
--- a/apps/app/src/index.react.tsx
+++ b/apps/app/src/index.react.tsx
@@ -16,6 +16,7 @@ import {
   PlatformProvider,
 } from "./react-app/kernel/platform";
 import { AppProviders } from "./react-app/shell/providers";
+import { AppErrorBoundary } from "./react-app/shell/app-error-boundary";
 import { AppRoot } from "./react-app/shell/app-root";
 import { setWebNotificationHandler } from "./react-app/shell/desktop-notifications";
 import { startDeepLinkBridge } from "./react-app/shell/startup-deep-links";
@@ -41,16 +42,18 @@ const Router = isDesktopRuntime() ? HashRouter : BrowserRouter;
 
 ReactDOM.createRoot(root).render(
   <React.StrictMode>
-    <QueryClientProvider client={queryClient}>
-      <TooltipProvider>
-        <PlatformProvider value={platform}>
-          <AppProviders>
-            <Router>
-              <AppRoot />
-            </Router>
-          </AppProviders>
-        </PlatformProvider>
-      </TooltipProvider>
-    </QueryClientProvider>
+    <AppErrorBoundary>
+      <QueryClientProvider client={queryClient}>
+        <TooltipProvider>
+          <PlatformProvider value={platform}>
+            <AppProviders>
+              <Router>
+                <AppRoot />
+              </Router>
+            </AppProviders>
+          </PlatformProvider>
+        </TooltipProvider>
+      </QueryClientProvider>
+    </AppErrorBoundary>
   </React.StrictMode>,
 );
--- a/apps/app/src/react-app/shell/app-error-boundary.tsx
+++ b/apps/app/src/react-app/shell/app-error-boundary.tsx
@@ -1,0 +1,82 @@
+/** @jsxImportSource react */
+
+import * as React from "react";
+
+interface AppErrorBoundaryState {
+  error: Error | null;
+}
+
+function formatError(error: Error): string {
+  return error.stack ?? `${error.name}: ${error.message}`;
+}
+
+/**
+ * Last-resort boundary around the whole application.
+ *
+ * React unmounts the entire tree when a render throws, so before this existed a
+ * single bad value rendered the app as a blank white window with nothing in the
+ * UI to explain it. The same failure mode is documented on the tool-part
+ * boundary in components/chat/message-list.tsx, which was added after it was
+ * seen in production.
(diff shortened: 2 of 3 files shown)


yomgui:
Thanks for this — the boundary and its placement were right, and it's exactly what was needed recently when a provider threw during render and produced a blank window.

This PR never reached a mergeable state for reasons unrelated to the code: as a fork PR the `OpenWork Tests` and `i18n Audit` workflows failed at the workflow-file stage without running any job, the Vercel deploys sat in `Authorization required`, and it never got a review, so it aged out of sync with `dev`. The repo also now requires journey-spec evidence in `evals/specs` rather than a unit test.

Rather than push to your fork, I rebased your commit onto current `dev` and continued in #4697 (superseding this one): your boundary + unit test are kept, with Copy details (message, stack, app version, distribution flavor), Open logs folder, Reload, and a Daytona journey spec that drives the real app through the crash. You're credited with a `Co-authored-by` trailer. Closing this in favor of #4697 once it merges.

yomgui:
Superseded by #4697, now merged to dev (61b4190) with your commit credited via Co-authored-by. Thanks again.

PR #3405, opened 2026-07-31
Fix duplicate local MCP servers on startup

### LINKED ISSUE #3325: [Bug]: Local MCP servers are started twice on launch
### Summary

On startup, every custom-configured local MCP server spawns two separate child process trees under the opencode sidecar. Only one is used; the second is an idle duplicate. Bundled MCP servers, such as imagegen-mcp-server, spawn only once, however.

### To Reproduce

1. Configure local stdio MCP servers (either in opencode.json or via OpenWork Settings > Extensions).
2. Launch OpenWork.
3. Check the process tree under opencode sidecar.

**Process tree:**
```
opencode-x86_64-pc-windows-msvc.exe
 |-- cmd.exe -> mcp-server1.exe
 |-- cmd.exe -> mcp-server1.exe
 |-- uvx.exe -> mcp-server2.exe
 |-- uvx.exe -> mcp-server2.exe
...
```

### Expected behavior

Each MCP server is started only once.

### Actual behavior

Each MCP server is started twice, doubling child process count.

### Screenshots (optional)

_No response_

- OS: Windows 10 x64
- OpenWork version: v0.18.10

During desktop server initialization in [`embedded.ts`](https://github.com/different-ai/openwork/blob/723c9db710244c35a1c869eb3a869d14c8d1889b/apps/server/src/embedded.ts), OpenWork calls [`syncAllWorkspacesRuntimeMcpToEngine()`](https://github.com/different-ai/openwork/blob/723c9db710244c35a1c869eb3a869d14c8d1889b/apps/server/src/embedded.ts#L125), which sends HTTP POST /mcp requests to register the MCP servers via the engine API. Prior to this, the OpenCode sidecar already loaded MCP definitions on startup from config files (opencode.json, `$env:OPENCODE_CONFIG`, etc.), resulting in duplicate MCP server registrations.

Deduplicating or skipping the startup sync call for servers already loaded by the engine should prevent the duplicate process tree.

Note: Bundled OpenWork MCP servers (such as imagegen-mcp-server) or MCPs belonging strictly to inactive workspaces spawn only once because they are registered through POST /mcp only.

mohitdebian:
Fixes #3325

### Description
This PR resolves a bug where local MCP servers were spawning duplicates on startup. It turns out that any `POST /mcp` request makes the OpenCode sidecar respawn every MCP server from the config file without killing the old ones.

### Changes Made
By removing the `mcp` section from the generated `runtime-opencode-config.json`, the engine natively spawns 0 MCP servers from the config file. OpenWork instead exclusively pushes all runtime MCP servers dynamically via `POST /mcp`. This relies on a single source of truth for MCP registration and avoids the sidecar respawning any duplicates.

### Testing
- Automated E2E tests (`pnpm --filter openwork-server test`) updated and passing (All 600 tests pass).
- Verified process tree generation is correctly deduplicated.

### DIFF
--- a/apps/server/src/openwork-runtime-config.ts
+++ b/apps/server/src/openwork-runtime-config.ts
@@ -131,7 +131,6 @@ export function buildOpenworkRuntimeConfigObjectFromSnapshot(
       ...runtimePluginList(runtimeConfig),
     ],
     ...(disabledProviders.length ? { disabled_providers: disabledProviders } : {}),
-    mcp: runtimeMcpMap(runtimeConfig),
     ...(Object.keys(provider).length ? { provider } : {}),
   };
 }


mohitdebian:
Fixes #3325

Based on new findings, the previous attempt was reverted. It turns out any `POST /mcp` request makes the OpenCode sidecar respawn every MCP server from the config file without killing the old ones.

To fix this, we've removed the `mcp` section from the generated `runtime-opencode-config.json`. This causes the engine to natively spawn 0 MCP servers from the config file, while OpenWork handles all runtime MCP server registrations dynamically via `POST /mcp`. This establishes a single source of truth for MCP registration and prevents the sidecar from respawning duplicates.

PR #3407, opened 2026-08-01
feat(i18n): default to Simplified Chinese when system language is zh

988hj7tczd-oss:
## Summary

When no stored language preference exists, `initLocale` currently always returns `"en"`. This makes first-run Chinese users see an English UI even though a complete Simplified Chinese translation ships with the app.

This change follows the browser/system language instead: if it starts with `zh`, default to `zh` (Simplified Chinese); otherwise fall back to `en`.

- Users can still switch manually via **Settings > Appearance > Language**
- A stored preference (localStorage `openwork.language`) always takes priority
- No behavior change for English users

Only `apps/app/src/i18n/index.ts` is modified (11 insertions). No other files touched.

### DIFF
--- a/apps/app/src/i18n/index.ts
+++ b/apps/app/src/i18n/index.ts
@@ -213,6 +213,17 @@ export const initLocale = (): Language => {
     console.warn("Failed to read language preference:", e);
   }
 
+  // When no stored preference exists, follow the system language:
+  // Chinese systems default to Simplified Chinese, everything else to English.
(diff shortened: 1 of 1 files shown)


988hj7tczd-oss:
Hi @benjaminshafii 👋

Small follow-up on this one: when no stored language preference exists, `initLocale` defaults to English. This PR makes Chinese-system users start in Simplified Chinese (the complete zh translation already ships with the app). It's only 11 lines in `index.ts`, no behavior change for English users.

Let me know if anything needs adjusting. Thanks!

PR #3413, opened 2026-08-01
fix(server): release embedded runtime config listener

### LINKED ISSUE #3321: openwork-server: no-op provider patch still reloads the engine (#3307 guard ineffective in production)

ousamabenyounes:
## Summary
- release the embedded server's runtime-config listener when its handle stops
- cover an in-process server replacement sharing the same runtime database
- add an app-less Fraimz flow that executes the real lifecycle regression test

## Why
- a stopped embedded server kept its runtime-config subscription alive and could rewrite the shared generated OpenCode config from its stale workspace
- that second writer made an identical provider PATCH appear file-changing and reloaded the engine
- Related PR #3349 improves no-op identity inside `mergeRuntimeProviderUpdate`, but the unchanged lifecycle regression test remains RED on its exact head (`bb090e630cda81421af72d4b9188587032f76dea`). This patch addresses the separate listener lifecycle that causes the file rewrite.

## Issue
- Closes #3321

## Scope
- embedded managed-OpenCode server listener registration and shutdown
- one server lifecycle integration test
- one internal Fraimz proof flow and voiceover

## Out of scope
- provider merge semantics covered by #3349
- CLI-owned process lifetime

### Ran
- `bun test src/embedded-runtime-config-lifecycle.test.ts`
- `bun test` (full server suite, compared with an identically provisioned `origin/dev` worktree)
- `pnpm fraimz --flow embedded-runtime-config-listener-lifecycle`
- real `curl -X PATCH /runtime-config/providers` after the two-server lifecycle setup

### Result
- targeted test: 1 pass, 0 fail
- typecheck: pass
- full server baseline: 597 pass, 10 skip, 4 fail, 1 error
- full server patched: 598 pass, 10 skip, 4 fail, 1 error; identical pre-existing failure set plus the new passing test
- Fraimz: PASSED, 1 passed, 0 failed, 0 skipped
- curl response: `changed:false`, `reload:"skipped"`

## Test verification (RED → GREEN)

Upstream `origin/dev` (`3f9c38a37125450e9b11ff74061ed500626254c9`) with only the new test:

```text
expected { changed: false, reload: "skipped" }
received { changed: false, reload: "reloaded" }
0 pass
1 fail
```

The same unchanged test on PR #3349 (`bb090e630cda81421af72d4b9188587032f76dea`):

```text
expected { changed: false, reload: "skipped" }
received { changed: false, reload: "reloaded" }
0 pass
1 fail
```

With this patch:

```text
PATCH /runtime-config/providers 200
PATCH /runtime-config/providers 200
1 pass
0 fail
```

Reverting only `apps/server/src/embedded.ts` returns the test to the same RED failure.

## CI status
- pass: targeted lifecycle test, TypeScript typecheck, Fraimz runtime proof
- code-related failures: none
- external/env/auth blockers: the pinned local Bun container has the same four server-suite failures and one `node:sqlite` runner error on upstream and patched trees; no baseline passing test regressed

## Manual verification
1. Start and stop an embedded managed server against a shared runtime DB.
2. Start a replacement server, patch a provider, then mutate the stopped workspace in that shared DB.
3. Repeat the provider PATCH and observe HTTP 200 with `changed:false` and `reload:"skipped"`.

## Evidence
- `evals/results/2026-08-01T12-15-15-633Z/fraimz.html` generated locally; every internal claim passed against the executed pinned-Bun integration test
- video/screenshot: N/A — internal server lifecycle with no visible UI surface

## Risk
- low: the change only retains and invokes an unsubscribe callback already returned by the runtime-config helper
- writes already queued before shutdown remain ordered by the existing per-path write queue, so the replacement server's later initial write lands last

## Rollback
- revert this commit; the previous behavior leaks the listener for the lifetime of the process

### DIFF
--- a/apps/server/src/embedded-runtime-config-lifecycle.test.ts
+++ b/apps/server/src/embedded-runtime-config-lifecycle.test.ts
@@ -1,0 +1,147 @@
+import { chmod, mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { describe, expect, test } from "bun:test";
+
+import { startEmbeddedServer } from "./embedded.js";
+import { writeRuntimeOpencodeConfig } from "./runtime-opencode-config-store.js";
+
+const HOST = "127.0.0.1";
+const SERVER_TOKEN = "server-token";
+const HOST_TOKEN = "host-token";
+const PROVIDER_ID = "lpr_anthropic";
+const PROVIDER = { id: "anthropic", name: "Anthropic", env: ["ANTHROPIC_API_KEY"] };
+const DEV_MODE_ENV = "OPENWORK_DEV_MODE";
+const RUNTIME_DB_ENV = "OPENWORK_RUNTIME_DB";
+const OPENCODE_BASE_URL_ENV = "OPENWORK_OPENCODE_BASE_URL";
+const FIRST_WORKSPACE_DIR = "first-workspace";
+const SECOND_WORKSPACE_DIR = "second-workspace";
+const RUNTIME_DB_FILE = "runtime.sqlite";
+const FIRST_CONFIG_FILE = "first-server.json";
+const SECOND_CONFIG_FILE = "second-server.json";
+const HTTP_OK = 200;
+const PATCH_METHOD = "PATCH";
+const CONTENT_TYPE = "application/json";
+const SKIPPED = "skipped";
+const STALE_MCP_URL = "https://stale.example.test/mcp";
+
+function restoreProcessEnv(name: string, value: string | undefined): void {
+  if (value === undefined) {
+    delete process.env[name];
+    return;
+  }
+  process.env[name] = value;
+}
+
+async function writeFakeOpencodeBin(root: string): Promise<string> {
+  const binPath = join(root, "fake-opencode.mjs");
+  await writeFile(binPath, [
+    "#!/usr/bin/env bun",
+    "const portIndex = process.argv.indexOf(\"--port\");",
+    "const requestedPort = Number(process.argv[portIndex + 1] ?? 0);",
+    "const server = Bun.serve({ port: requestedPort, fetch: () => Response.json({ ok: true }) });",
+    "console.log(`opencode server listening on http://127.0.0.1:${server.port}`);",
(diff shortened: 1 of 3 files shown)


benjaminshafii:

PR #3414, opened 2026-08-01
fix(release): install both macOS CPU variants so Intel builds get node-pty

### LINKED ISSUE #2831: [Bug]: macOS Intel Darwin Binary missing
### Summary

App starts with Error Message:

```
Uncaught Exception:
Error: The @lydell/node-pty package supports your platform (darwin-x64), but it could not find the platform-specific package for it: @lydell/node-pty-darwin-x64

This can happen if you use the "--omit=optional" (or "--no-optional") npm flag.
The "optionalDependencies" package.json feature is used to install the correct
binary executable for your current platform. Remove that flag to use @lydell/node-pty.

This can also happen if the "node_modules" folder was copied between two operating systems
that need different binaries - including "virtual" operating systems like Docker and WSL.
If so, try installing with npm rather than copying "node_modules".
at requirePlatformSpecificPackage (/Applications/OpenWork.app/Contents/Resources/app.asar/node_modules/node-pty/index.js:19:13)
at Object.<anonymous> (/Applications/OpenWork.app/Contents/Resources/app.asar/node_modules/node-pty/index.js:47:18)
at Module._compile (node:internal/modules/cjs/loader:1738:14)
at Module._extensions..js (node:internal/modules/cjs/loader:1904:10)
at Module.load (node:internal/modules/cjs/loader:1472:32)
at Module._load (node:internal/modules/cjs/loader:1289:12)
at c._load (node:electron/js2c/node_init:2:17950)
at TracingChannel.traceSync (node:diagnostics_channel:322:14)
at wrapModuleLoad (node:internal/modules/cjs/loader:242:24)
at Module.require (node:internal/modules/cjs/loader:1494:12)
```
Root cause: The OpenWork.app was built/packaged with the wrong platform-specific binary.
Since it's running under Rosetta (x86_64), it needs the darwin-x64 binary, but the app only
shipped the darwin-arm64 binary.

Fix: @lydell/node-pty-darwin-x64 package into the app's
node_modules, not just cp...

### To Reproduce

Download latest DMG for Intel Macs
Copy App to Application folder
Start App. 

### Expected behavior

Working :-) 

### Actual behavior

Has missing Darwing Binary to work with Rosetta
 OpenWork.app is an Intel (x86_64) binary running under
Rosetta 2
- node-pty correctly detects the runtime platform as darwin-x64 (Rosetta)
- But only @lydell/node-pty-darwin-arm64 was installed — the darwin-x64 binary is missing
The fix is to install the missing x64 platform binary

### Screenshots (optional)

_No response_

- OpenWork v0.17.30
- macOS 26.5.1 (25F80)

_No response_

mvanhorn:
## Summary
- The macOS Electron release job installed dependencies with a plain `pnpm install --frozen-lockfile`, which resolves CPU-specific optional deps (notably `node-pty`) for the runner's own architecture only. The x64 artifact built on an arm64 runner therefore shipped without a usable Intel `node-pty`, and the terminal failed on Intel Macs.
- This installs both CPU variants on macOS and adds a regression test that asserts the workflow keeps doing so.

## Why
- Intel Mac users hit a broken terminal in the released build; the arm64 build was fine, which is why it was easy to miss.
- The failure lives in release CI, not in app code, so nothing in the existing test suite covered it. Without a test asserting the workflow shape, the next person editing the install step can silently reintroduce it.

## Issue
- Closes #2831

## Scope
- `.github/workflows/release-macos-aarch64.yml`: split the single `Install dependencies` step into a macOS step that runs `pnpm install --frozen-lockfile --cpu=arm64 --cpu=x64` and a Linux/Windows step that keeps the existing `pnpm install --frozen-lockfile`, selected by `matrix.os_type`.
- `apps/desktop/electron/electron-builder-config.test.mjs`: new test `installs both macOS CPU variants for Electron releases`, which parses the release workflow and asserts the macOS matrix covers both `electron-macos-arm64` and `electron-macos-x64` and that the macOS install step requests both CPUs.

## Out of scope
- No change to `node-pty` itself, its version, or how the app loads it.
- No change to Linux or Windows install behavior; those keep the original single-CPU install.
- No change to signing, notarization, or publishing steps.
- Not addressing any other cross-arch dependency; this fixes the install step that caused the reported terminal failure.

## Testing
Ran the Electron distribution config suite locally, which now includes the new workflow assertion.

### Ran
- `node --test electron/electron-builder-config.test.mjs` (from `apps/desktop`)

### Result
- pass/fail: pass
- 5 tests, 5 passing, 0 failing, including the new `installs both macOS CPU variants for Electron releases`.
- if fail, exact files/errors: n/a

## CI status
- pass: the config test suite above passes locally.
- code-related failures: none observed.
- external/env/auth blockers: I cannot run the actual macOS release job from a fork; it needs the repo's signing and notarization secrets. The regression test asserts the workflow's shape instead, which is the part that regressed.

## Manual verification
1. Check out this branch and run `node --test electron/electron-builder-config.test.mjs` from `apps/desktop`; the new case passes.
2. Revert only the workflow hunk (restore the single `pnpm install --frozen-lockfile`) and re-run; the new test fails, showing it actually pins the behavior rather than passing vacuously.
3. On a real release run, confirm the macOS install step logs both `--cpu=arm64` and `--cpu=x64`, and that the x64 artifact contains an Intel `node-pty` binary.

## Evidence
- N/A (CI-configuration change; the regression proof is the test in step 2 above, which fails against the previous workflow and passes with this change).

## Risk
- Low. The install command changes only on macOS, and adding a second `--cpu` target is additive: arm64 resolution is unchanged, x64 binaries are added alongside.
- Install time on the macOS runner grows slightly because both optional-dependency sets are fetched.
- If a dependency ever ships an arm64 build but no x64 build, the macOS install would fail loudly at CI time rather than producing a silently broken artifact. I consider that the better failure mode, but it is a behavior change worth knowing about.

## Rollback
- Revert this PR. The two hunks are independent of each other and of the rest of the release pipeline: restoring the single `Install dependencies` step returns the previous behavior exactly, and dropping the test removes the assertion. No state, cache, or published artifact needs cleanup.

### DIFF
--- a/.github/workflows/release-macos-aarch64.yml
+++ b/.github/workflows/release-macos-aarch64.yml
@@ -463,7 +463,12 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-electron-pnpm-
 
-      - name: Install dependencies
+      - name: Install dependencies (macOS)
+        if: matrix.os_type == 'macos'
+        run: pnpm install --frozen-lockfile --cpu=arm64 --cpu=x64
+
+      - name: Install dependencies (Linux and Windows)
+        if: matrix.os_type != 'macos'
         run: pnpm install --frozen-lockfile
 
       - name: Write Electron notary API key (macOS)
--- a/apps/desktop/electron/electron-builder-config.test.mjs
+++ b/apps/desktop/electron/electron-builder-config.test.mjs
@@ -11,6 +11,19 @@ async function readConfig(name) {
   return YAML.parse(await readFile(path.resolve(dirname, "..", name), "utf8"));
 }
 
+async function readReleaseWorkflow() {
+  return YAML.parse(
+    await readFile(
+      path.resolve(
+        dirname,
+        "../../..",
+        ".github/workflows/release-macos-aarch64.yml",
+      ),
+      "utf8",
+    ),
+  );
+}
+
 describe("Electron distribution configs", () => {
   it("uses a stable Linux desktop identity and ships integration icons", async () => {
     const packageMetadata = JSON.parse(
@@ -33,7 +46,10 @@ describe("Electron distribution configs", () => {
     assert.equal(config.appId, "com.differentai.openwork");
     assert.equal(config.productName, "OpenWork");
     assert.equal(config.protocols[0].schemes[0], "openwork");
-    assert.equal(config.artifactName, "openwork-${os}-${arch}-${version}.${ext}");
+    assert.equal(
+      config.artifactName,
+      "openwork-${os}-${arch}-${version}.${ext}",
+    );
   });
 
   it("defines an enterprise flavor with the standard app identity and release provider", async () => {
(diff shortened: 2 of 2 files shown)


mvanhorn:
The four red Vercel checks here aren't from this change — they're all returning a `vercel.com/git/authorize` link rather than a build failure, which is Vercel declining to build a fork PR without team authorization. Nothing in the diff touches the app builds.

Flagging it so it doesn't read as a broken PR. Happy to rebase or do anything else that helps, but I can't clear those from my side.

benjaminshafii:

PR #3469, opened 2026-08-03
ci(changelog): auto-generate release notes on stable releases

benjaminshafii:
## What

Automates the changelog workflow we just did by hand for v0.18.9–v0.18.13. On every stable release (`release: published`, tags matching `vX.Y.Z` — alphas and signpath tags skipped), a new **Changelog** workflow:

1. **Gathers deterministic facts by script** — previous stable tag, commit list (`git log --no-merges prev..tag`), `git diff --shortstat` LOC line, publish timestamp, docs date label, tracker filename. The LLM never computes numbers or dates.
2. **Drafts the prose with a new `changelog` agent** (`.opencode/agents/changelog.md`, same pattern as the existing `triage`/`duplicate-pr` CI agents): one `<Update>` entry in `packages/docs/changelog.mdx` + a structured section in `changelog/release-tracker-<date>.md`, following the exact formats already in those files.
3. **Verifies the output with `scripts/check-changelog-output.mjs`** — only the two allowed paths changed, balanced `<Update>` tags, exact compare link exactly once, strictly descending version order, tracker LOC line recomputed independently and matched verbatim, all 16 tracker headings present in order.
4. **Opens a signed, auto-merging PR against `dev`** (same signing + auto-merge pattern as `update-models.yml`). Mintlify/warden checks remain the merge gate.

Also dispatchable manually (`workflow_dispatch` with a `tag` input) for backfills; it exits cleanly when the tag is already documented (idempotent) or not a stable tag.

## Testing

CI/agent-config change; the workflow itself cannot run pre-merge, so it was rehearsed end-to-end locally in a worktree:

- Built the exact CI prompt for **v0.18.8** with the workflow's own shell logic, ran the real `changelog` agent, then the guardrail: all checks green, entry inserted in the correct backfill position (directly below v0.18.9), tracker format exact, LOC line verbatim.
- Guardrail negative tests: clean tree → fails with "no changes"; removed one `</Update>` → fails on tag balance; unexpected changed paths → fails and names them (this caught real `.opencode/package-lock.json` runtime churn, now discarded by a dedicated workflow step before verification).
- Guardrail positive test against the real historical v0.18.13 docs diff → pass.
- `node --check scripts/check-changelog-output.mjs` and `actionlint` (clean apart from the expected unknown Blacksmith runner label).

**Post-merge proof**: I will dispatch `changelog.yml` with `tag=v0.18.8` (the one undocumented version below the new entries) — the run doubles as a real backfill and the end-to-end CI validation of the full pipeline including the auto-merge PR.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,193 @@
+name: Changelog
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Stable release tag to document (e.g. v0.18.13)
+        required: true
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: changelog-${{ github.event.release.tag_name || inputs.tag }}
+  cancel-in-progress: false
+
+jobs:
+  changelog:
+    if: github.repository == 'different-ai/openwork'
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    timeout-minutes: 30
+    env:
+      BASE_BRANCH: dev
+      TAG: ${{ github.event.release.tag_name || inputs.tag }}
+      BRANCH_NAME: automation/changelog-${{ github.event.release.tag_name || inputs.tag }}-${{ github.run_id }}
+
+    steps:
+      - name: Validate tag
+        id: guard
+        run: |
+          if [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "stable=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "stable=false" >> "$GITHUB_OUTPUT"
+            echo "Skipping changelog automation because $TAG is not a stable release tag."
+          fi
+
+      - name: Checkout
+        if: steps.guard.outputs.stable == 'true'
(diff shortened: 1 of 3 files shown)


benjaminshafii:
Review guide (3 files, ~5 min):

- **`.github/workflows/changelog.yml`** — the security-relevant one. Secrets used: `OPENCODE_API_KEY` (LLM draft step only) and `COMMIT_SIGNING_KEY` (same signing pattern as `update-models.yml`). Permissions are `contents: write, pull-requests: write`. The LLM output is hard-verified by script before anything is committed.
- **`.opencode/agents/changelog.md`** — prose agent, file tools only (`read/glob/grep/edit/write`), no bash.
- **`scripts/check-changelog-output.mjs`** — the guardrail: only 2 allowed paths may change, balanced `<Update>` tags, exact compare link, descending version order, LOC line recomputed independently and matched verbatim.

Why review is needed at all: warden intentionally refuses to self-clear PRs touching `.github/` — every check is otherwise green and auto-merge is armed, so one approval lands it.

After merge I'll dispatch the proof run (`changelog.yml` with `tag=v0.18.8`) which backfills the one undocumented version and validates the pipeline end-to-end in CI.

src-opn (approved):

Issue #2101, opened 2026-06-06
Can't use custom agents from the app — the agent picker is commented out and @mentions don't cut it


zantak:
Hey! I've been setting up custom agents (`~/.config/opencode/agent/*.md`) and realized there's currently no way to actually use them from OpenWork.

Dug around a bit and found the agent picker in the composer is commented out in `composer.tsx`:

```tsx
{/* TODO: Decide what to do with agent selection before showing this control again.
```

All the wiring is still there (`selectedAgent` state, `listAgents`, `onSelectAgent`, and it ends up as `agent:` in `session.promptAsync`) — there's just no button to reach it anymore.

So I tried `@mentioning` the agent instead, but that becomes an agent *part*, which the engine treats as a subagent call. With a `mode: primary` agent the model just answers:

> The subagent type X is not available. The task tool only supports: explore, general

The sneaky part: it *looks* like your agent took over, but checking the opencode db, every reply actually came from `build`. The same agents work fine in the opencode TUI with Tab.

One more thing — even when `selectedAgent` does get set, it's a plain `useState(null)` in `session-route.tsx`, so it doesn't survive a reload. `modelVariant` goes through prefs, the agent selection probably should too.

So ideally:

- bring the picker back (the commented control looks ready, and the `listAgents` filter already excludes hidden/subagent entries)
- persist the selection per session so you don't silently fall back to `build`

I'm on 0.15.1/0.15.2, Linux AppImage, engine is an external `opencode serve` via `OPENWORK_OPENCODE_BASE_URL`. Happy to test a fix!

PR #3475, opened 2026-08-03
fix(app): restore per-session agent selection when switching between sessions

### LINKED ISSUE #3461: [Bug]: Chosen agent per session keep being overwritten by the latest chat
### Summary

previously If I have two agents ( ex. review agent and coding agent.) and had two chats going on at the same time, each chat's agent would "stick" or stay the same even if I keep switching between them, now this is no longer the case and when I switch between sessions / chats the agent stays on whatever the previous session was.

### To Reproduce

1. create session1 with an agentA.
2. create session2 with agentB.
3. switch back to session1

### Expected behavior

new messages to session1 would be with agentA.

### Actual behavior

new messages to session1 are set to agentB and you have to manually change it to agentA.

### Screenshots (optional)

_No response_

_No response_

_No response_

PriyeshPandey2000:
## Summary
- Agent selection lived in one global slot (`local.prefs.selectedAgent`) shared by every session, so picking an agent in one conversation overwrote it for all of them
- Adds `session-agent-store.ts`, a localStorage store keyed by session ID, mirroring the existing `session-model-store.ts`
- `session-route.tsx` reads/writes the active session's agent; the new-task composer still uses the global pref

## Why
- Returning to a conversation silently changed its agent, so the next message ran under the wrong one
- Per-session model choice already works this way; agent selection was the outlier

## Issue
- Closes #3461

## Scope
- `session-agent-store.ts` (new): per-session agent map under `openwork.sessionAgents.v1`, capped at 200 entries, tolerant of missing or malformed storage
- `session-route.tsx`: `selectedAgent` derives from the active session; `setSelectedAgent` writes per-session, or to the global pref when no session is selected

## Out of scope
- `local.prefs.selectedAgent` stays as the new-task composer's source of truth, preserving the reload persistence from `#2101`
- New sessions start at the default agent rather than inheriting the composer's choice — `session-model-store.ts` behaves the same way today, so this keeps the two consistent rather than changing one in a bugfix
- No server, schema, or sync changes — local UI state only

### Ran
- `pnpm typecheck`
- `bun test --isolate tests/` (apps/app unit suite)
- `pnpm test:session-switch`
- `pnpm test:sessions`
- `pnpm test:health`

### Result
- pass/fail: pass
- if fail, exact files/errors: `tests/message-list-loading.test.tsx` has 2 failures (`Invalid hook call` / `TypeError: null is not an object (evaluating 'resolveDispatcher().useState')` in `renderToStringImpl`). These are pre-existing on `dev` — I re-ran that file with this branch's two files reverted to their `dev` state and got the identical 2 failures, so they are unrelated to this change. Everything else is green: 600 pass in the unit suite, 7/7 steps in `test:session-switch`, `ok: true` for `test:sessions` and `test:health`.

## CI status
- pass: typecheck, session-switch, sessions, health
- code-related failures: none from this change

## Manual verification
1. Set the `build` agent on session A and `plan` on session B
2. Switch back to session A — the picker shows `build` (before this change it showed `plan`)
3. Reload the app and switch between both sessions — each still restores its own agent
4. With no session selected, pick an agent in the new-task composer and reload — the choice is restored
5. Clear a session's agent back to default — the entry is removed and the session falls back to the default agent

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/session-agent-store.ts
+++ b/apps/app/src/react-app/domains/session/surface/session-agent-store.ts
@@ -1,7 +1,8 @@
 // Per-conversation agent memory. Each session can pick its own agent from
 // the composer; the choice is remembered in localStorage so returning to a
-// conversation restores the agent it last used. Sessions without a remembered
-// choice fall back to the global default agent preference.
+// conversation restores the agent it last used. A session with no stored
+// choice has no entry here and falls back to the default agent. The global
+// agent preference is only used outside of any session (the new-task composer).
 import { create } from "zustand";
 
 const STORAGE_KEY = "openwork.sessionAgents.v1";
@@ -68,7 +69,3 @@ export const useSessionAgentStore = create<SessionAgentStore>((set) => ({
       return { bySessionId };
     }),
 }));
-
-export function getSessionAgent(sessionId: string): string | null {
-  return useSessionAgentStore.getState().bySessionId[sessionId] ?? null;
-}
--- a/apps/app/src/react-app/domains/session/surface/session-agent-store.ts
+++ b/apps/app/src/react-app/domains/session/surface/session-agent-store.ts
@@ -1,0 +1,74 @@
+// Per-conversation agent memory. Each session can pick its own agent from
+// the composer; the choice is remembered in localStorage so returning to a
+// conversation restores the agent it last used. Sessions without a remembered
+// choice fall back to the global default agent preference.
+import { create } from "zustand";
+
+const STORAGE_KEY = "openwork.sessionAgents.v1";
+const MAX_REMEMBERED_SESSIONS = 200;
+
+function readStoredAgents(): Record<string, string> {
+  if (typeof window === "undefined") return {};
+  try {
+    const raw = window.localStorage.getItem(STORAGE_KEY);
+    if (!raw) return {};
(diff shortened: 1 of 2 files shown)


benjaminshafii:

PriyeshPandey2000:
Closing per the maintainer's note — the issue no longer reproduces on the latest `dev`.

**Covered by #4841** (`fix(app): keep conversation modes independent and switchable while running`), which landed `apps/app/src/react-app/domains/session/surface/session-mode-memory.ts`. That implementation:
- Stores per-session agent choice under the **same localStorage key** this PR introduced (`openwork.sessionAgents.v1`, 200-entry cap), a strict superset of this approach.
- Handles explicit "Default", split-pane isolation, and keeps the new-task global preference separate.

**Evidence:** `apps/app/tests/session-mode-memory.test.tsx` on `dev` asserts the exact #3461 scenario — *"remembers independent choices and explicit Default across reload"* sets `plan` on session A and `build` on session B, then verifies each restores its own — plus malformed-storage tolerance and reload persistence. That suite passes on `dev`; the bug is gone.

A rebase of #3475 onto current `dev` would conflict with the `useSessionAgentSelection` refactor in `session-route.tsx` and duplicate `session-mode-memory.ts`, adding no behavior `dev` doesn't already have. Closing instead.

PR #3487, opened 2026-08-04
fix(server): keep engine credentials out of the trusted-process identity

reachjalil:
## Summary

- Stop embedding the managed OpenCode engine's basic-auth username and password in the trusted-process identity string, in both entry points ([embedded.ts](https://github.com/different-ai/openwork/blob/dev/apps/server/src/embedded.ts) and [cli.ts](https://github.com/different-ai/openwork/blob/dev/apps/server/src/cli.ts)).
- The identity is opaque — it is hashed immediately by `registerTrustedOpencodeProcess` and only ever compared, never reported — so it only needs per-boot uniqueness. A `randomUUID()` nonce plus the pid provides that without routing credentials through the fast SHA-256 fingerprint.

## Why

Resolves CodeQL alert #115 (`js/insufficient-password-hash`, open on `dev`): the engine password flowed from the identity join into `hashToken` (plain SHA-256) in `utils.ts`. The alert also blocks the CodeQL check on any PR whose diff touches that flow path (e.g. #3486). This removes the tainted flow at its source rather than dismissing the alert.

## Behavior

No behavior change. Registration and `clearTrustedOpencodeProcess` compare the same opaque string held by the caller; nothing reconstructs the identity from credentials.

## Tests

- `pnpm --filter openwork-server typecheck` — passed
- `bun test src/agent-context-diagnostics.test.ts src/mcp.engine-sync.e2e.test.ts src/cloud-mcp-reconcile.e2e.test.ts` (bun 1.3.10) — 109 passed, 0 failed

## Note

#3486 rewrites the surrounding block in `embedded.ts`; whichever lands second needs a trivial rebase of the identity lines.

### DIFF
--- a/apps/server/src/cli.ts
+++ b/apps/server/src/cli.ts
@@ -1,5 +1,6 @@
 #!/usr/bin/env bun
 
+import { randomUUID } from "node:crypto";
 import { mkdir } from "node:fs/promises";
 
 import { parseCliArgs, printHelp, resolveServerConfig } from "./config.js";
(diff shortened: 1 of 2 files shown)


reachjalil:
Superseded by #3488 — same commit, head branch moved from the fork into this repo so checks and merge requirements run with in-repo permissions.

Issue #3533, opened 2026-08-04
Tracking: GitHub connector sync reliability


benjaminshafii:
Umbrella for making GitHub connector sync self-healing. Context: #3521 / #3522 (fixed in #3523 / #3524), end-to-end coverage + deterministic GitHub witness in #3529.

## Tier 1 — structural (in progress, demo-driven)
- [ ] #3530 — real queue: webhook acks instantly, background worker with per-target ordering + backoff retries; Retry button becomes real
- [ ] #3531 — periodic reconciliation sweep heals missed webhooks (shared code path with "Sync now")
- [ ] #3532 — repository rename/removal handling by repositoryId

## Tier 1.5 — fleet-scale reconciliation
- [ ] #3568 — deliveries-API reconciliation, failure-triggered probes, leader lease; interim mitigation: token cache + capped/jittered sweep

## Tier 2 — request hardening
- [ ] Retry + rate-limit awareness in `requestGithubJson` (honor `retry-after` / `x-ratelimit-reset`; jittered backoff on 429/secondary-limit/5xx/network; fail fast on semantic 4xx)
- [ ] Cache installation access tokens (~55 min TTL per installationId) instead of minting per operation
- [ ] Conditional requests (ETag/`If-None-Match`) for staleness probes and contents fetches

## Tier 3 — semantics + visibility
- [ ] Per-file apply resilience: itemized `partial` sync summaries (file + reason) instead of aborting the whole apply on one bad file (follow-up noted on #3521; strict `invalid_skill_name` also hard-fails applies today)
- [ ] Surface truncated `git/trees?recursive=1` responses as an explicit warning
- [ ] Connector health surface: last-sync age/status badge, failure-streak notification, webhook delivery health
- [ ] Fault-injection extensions to the `mock-github` witness (429s, timeouts, out-of-order pushes, renames, truncated trees) with a spec per failure mode

Every item is deterministically testable via the `mock-github` witness added in #3529.

pungsojun-cell:
มีพวกมึงมีความสุขมากไหมอีเหี้ยเชิญอยู่ได้ชวนอยู่ได้กูไปจีบเมียมึงบ้าง
จะลองหลอกเย็ดเมียมึงบ้างไอ้สัตว์มึงจะรู้สึกยังไง

ในวันที่ พฤ. 6 ส.ค. 2026 13:41 น. bauerpaul874801 ***@***.***>
เขียนว่า:

pungsojun-cell:
[...]

PR #3529, opened 2026-08-04
test(evals): cover GitHub connector repo pagination and marketplace skill import end-to-end

benjaminshafii:
Follow-up coverage for #3523 (fixes #3521) and #3524 (fixes #3522), per the repo validation standard (`evals/specs/**` + `@openwork/testkit`).

## What this adds

- **`evals/packages/labs/src/mock-github.ts`** — a deterministic GitHub provider witness (node:http), following the mock-google pattern: installation token minting, `/installation/repositories` with **real GitHub paging semantics (default per_page=30)** so the pagination regression stays observable, repo/branch/commit/tree/contents endpoints with deterministic content-hash shas, and a request log capturing method/path/query/ISO-time/authorization for per-identity witness assertions.
- **`evals/specs/github-connector-import.slow.test.ts`** — a stack spec that cold-boots a real Den (`server()`, ephemeral MySQL DB), points the connector at the witness, and drives the org API end-to-end: create connector account (installation 777, 130 repositories) → list repositories → GitHub setup on a **private** Claude marketplace repo (branch `develop`, mirroring the original user report) → discovery → apply → config-object versions → re-apply.
- **`ee/apps/den-api/.../github-app.ts`** — a lazy `GITHUB_CONNECTOR_API_BASE` env override (4 lines) so specs can aim the connector client at a witness. Read per-request, no env-module import (unit tests import this module bare), production default unchanged.

## Claims (all hold in the tape, every one with an observable assertion)

1. The repository picker returns **all 130** installation repositories in provider order across cursor pages, including the private repo with its marketplace manifest badge *(regression: only the first 30 before #3524)*.
2. Provider enumeration requests exactly pages 1 and 2 at `per_page=100` — no unpaged call, no page 3.
3. All repository/content traffic carries the **installation access token**, not the app JWT (witness per-request authorization attribution).
4. Discovery classifies the private marketplace repo, one supported plugin, skills componentPath, zero warnings.
5. **Apply returns 200** and materializes exactly one skill with frontmatter-derived `title: probe` / description *(regression: `invalid_skill_frontmatter` on every apply before #3523)*.
6. The companion markdown under the skill directory is **not** imported (negative half: exactly one materialized object, at the SKILL.md path).
[...]

Issue #3521, opened 2026-08-04
GitHub connector: apply always fails with invalid_skill_frontmatter, even for valid SKILL.md


benjaminshafii:
## Bug

Applying GitHub connector discovery (`/dashboard/integrations/github`) always fails with `invalid_skill_frontmatter` when the selected plugin contains skills — discovery itself succeeds (`claude_marketplace_repo`, plugin supported, no warnings) and `importPlansByPluginKey` lists the right `SKILL.md` paths with valid blob shas.

Reported via user feedback (openwork-web-app). The reporter reproduced it with a minimal plugin: one 8-line `skills/probe/SKILL.md` with clean `---` frontmatter (verified `---\n` at byte 0, no BOM, LF endings). Moving companion docs out of `skills/`, renaming directories, and shortening the description made no difference — nothing the user does to the file can fix it.

## Root cause

`materializeGithubImportedObject` in `ee/apps/den-api/src/routes/org/plugin-system/store.ts` strips the YAML frontmatter from the fetched file before deriving the projection:

```ts
const projectionRawSource = hasFrontmatter
  ? parseMarkdownFrontmatter(input.rawSourceText).body   // frontmatter removed
  : input.rawSourceText
const projection = deriveProjection({ objectType, value: { ..., rawSourceText: projectionRawSource } })
```

For `objectType === "skill"`, `deriveProjection` dispatches to `deriveSkillProjection`, which re-parses the text with `parseSkillMarkdown` and **requires** frontmatter — throwing `invalid_skill_frontmatter` (400) when it is absent. So:

- skill file **with** valid frontmatter → frontmatter stripped → `invalid_skill_frontmatter`
- skill file **without** frontmatter → `invalid_skill_frontmatter`

Every GitHub-connector skill materialization fails, on both first apply and webhook-driven re-sync. The stripping is intentional and harmless for the other object types (command/agent/hook/mcp), whose generic projection uses metadata-derived title/description.

## Fix

Keep the full `SKILL.md` text when deriving skill projections (`deriveSkillProjection` parses and validates the frontmatter itself); preserve the frontmatter-stripping behavior for all other object types. Regression tests cover the reporter's exact repro (incl. a CRLF variant).

Possibly related (different code path, similar smell): #2350 — marketplace install mangles SKILL.md frontmatter during local materialization.

## Follow-up worth splitting out

[...]

PR #3536, opened 2026-08-04
feat(den-api): self-healing GitHub connector sync — queue, retries, reconciliation, renames

### LINKED ISSUE #3530: GitHub webhook sync runs inline in the delivery request; queued/retried sync events are never processed
## Problem

`enqueueGithubWebhookSync` executes full discovery + materialization **synchronously inside the webhook HTTP handler** (`ee/apps/den-api/src/routes/webhooks/github.ts:84` → `store.ts` auto-import loop). Consequences:

- GitHub allows ~10s per delivery and **GitHub Apps do not auto-retry failed deliveries** — a slow tree/contents fetch means the push is silently lost.
- `ConnectorSyncEventTable` has a `queued` status, but **nothing consumes it**: `retryConnectorSyncEvent` (`store.ts`) flips a row back to `queued` forever — the UI Retry button is a no-op.
- Transient GitHub failures mark the event `failed` and that revision is never re-attempted.
- Concurrent deliveries for back-to-back pushes can process out of order (the per-file blob-sha guard prevents redundant fetches, not stale-overwrites-fresh).

## Fix

Ack-then-process:
1. Webhook ingress persists the event as `queued` and returns 202 immediately (no GitHub API calls in the request path).
2. A background worker claims events with **per-target serialization** (ordering), jittered exponential backoff retries, and a dead-letter/`failed` terminal state after N attempts.
3. `retryConnectorSyncEvent` re-enqueues into the same worker (becomes real).

The automations subsystem already implements this claim/retry queue shape (`ee/apps/den-api/src/automations/repository.ts:328-641`) — reuse the pattern/infra.

Deterministically testable with the `mock-github` witness from #3529 (inject latency/429s, assert ingress acks before any provider call, assert retry then completion).

Part of the GitHub sync reliability plan (see tracking issue).

### LINKED ISSUE #3531: GitHub sync has no reconciliation: a missed webhook causes permanent drift
## Problem

If a webhook delivery is lost (den downtime, secret rotation, GitHub incident, delivery timeout), the connector target drifts from the repository **forever**. The head-SHA staleness probe (#1871) only runs when a human opens the discovery UI; push-driven sync is the only automatic path.

## Fix

Periodic reconciliation sweep (~every 15 min, jittered):
- For each active GitHub connector target: one `GET /repos/{owner}/{repo}/commits/{branch}` (with ETag/`If-None-Match` so unchanged branches cost no rate limit).
- If head SHA ≠ last **completed** sync event's `sourceRevisionRef`, enqueue a sync event through the same worker queue as webhook events (see the inline-webhook issue) — never import inline in the sweep.
- Surface the sweep as the "Sync now" action so manual refresh and self-healing share one code path.

Deterministically testable with the `mock-github` witness (#3529): drop a webhook, advance the fixture head, assert the sweep enqueues and heals within one cycle.

Part of the GitHub sync reliability plan (see tracking issue).

### LINKED ISSUE #3532: Renaming or removing a GitHub repository silently kills connector sync
## Problem

Connector targets are keyed by `remoteId = repositoryFullName`. The webhook ingress accepts `repository` and `installation_repositories` events but `enqueueGithubWebhookSync` ignores everything except `push` and `installation.deleted` (`store.ts`). So:

- **Rename**: after a rename on GitHub, push payloads carry the new `full_name`, the target lookup by `remoteId` never matches again, and sync stops silently. (The original reporter of #3521 renamed their repo mid-debugging — this is a real path.)
- **Removal from the installation**: targets keep looking healthy while every sync fails or goes stale, with no user-facing signal.

## Fix

- Handle `repository.renamed` (and transfers): match the target by `repositoryId` (already stored in `targetConfigJson`), update `remoteId`/`repositoryFullName`, record a sync event documenting the rename, and continue syncing.
- Handle `installation_repositories.removed` / `repository.deleted`: mark affected targets disconnected and surface that state in the connector UI instead of failing silently.
- Reconciliation sweep (companion issue) should also detect renames via `GET /repos/{id}`-style lookup when the full-name fetch 404s.

Deterministically testable with the `mock-github` witness (#3529): rename the fixture repo, deliver the rename event, assert the target follows and the next push syncs.

Part of the GitHub sync reliability plan (see tracking issue).

benjaminshafii:
Fixes #3530, fixes #3531, fixes #3532 (Tier 1 of #3533). Demo-driven: the approved six-frame voice-over is encoded 1:1 in `evals/specs/github-sync-self-healing.slow.test.ts`.

> **Stacked on #3529** (scheduled to merge at 20:45 PT): this branch is based on `spec/github-connector-import` because the spec builds on the `mock-github` witness. Once #3529 lands I'll merge `dev` in and the diff collapses to just this feature.

## What changes

**Queue (ack-then-process) — #3530**
- Webhook ingress does zero GitHub API work in-request: push deliveries persist `queued` sync events and ack 202 immediately (GitHub's 10s delivery window can no longer eat pushes; GitHub Apps never redeliver on their own).
- New `workers/github-sync.ts` (started/stopped in `server.ts` like the other loops): claims due events oldest-first with **per-target serialization** and an optimistic `WHERE status='queued'` row guard (multi-instance safe); executes the existing auto-import; transient provider failures (429/5xx/network) retry with jittered exponential backoff (`attempt_count`, `next_attempt_at` — migration `0054`), dead-letter `failed` after `GITHUB_SYNC_MAX_ATTEMPTS`.
- `retryConnectorSyncEvent` now re-enqueues into the worker — the UI Retry button used to flip a row **nothing consumed**.

**Reconciliation — #3531**
- Sweep every `GITHUB_RECONCILE_INTERVAL_MS` (default 15 min): one head-SHA probe per active target; drift vs the last completed sync enqueues a `manual_resync`. Lost webhooks heal without a human.
- `POST /v1/connector-instances/:id/sync-now` + a den-web **Sync now** button run the same path on demand.

**Renames/removals — #3532**
- `repository.renamed` is matched by `repositoryId` and updates target+instance identity, records the rename event, and enqueues a continuity resync (previously: sync silently died forever — the #3521 reporter hit exactly this). `installation_repositories.removed` marks the instance `disabled` with a recorded event.

**Pre-existing prod bug found & fixed by the spec**
- Connector automation contexts (`session: null`) tripped `ensureFreshPluginArchAdmin`'s interactive step-up guard, so **webhook-driven auto-import always failed with a reauth error** on dev. Automation contexts now carry an explicit `automation: true` flag settable only by server-side builders (verified: `actorContext` cannot produce it); human reauth behavior unchanged and still covered by a negative test.

## Demo frames = spec claims (all green)
1. Push → skill auto-updates; ack precedes processing (witnessed via injected provider latency); event carries the exact commit.
2. 429s mid-sync → retries with backoff (`attemptCount ≥ 1`) → completes; never `failed`.
3. Back-to-back pushes → serialized; newest commit wins; no stale overwrite.
4. Repo renamed on GitHub → target follows; next push (new full name) syncs.
5. Webhook never delivered → reconciliation heals drift within one sweep cycle.
6. Sync now enqueues the same check; the sync history shows the whole story (pushes, retries, rename, resyncs).

## Validation
- `evals`: `pnpm exec vitest run --config vitest.config.ts --project stack specs/github-sync-self-healing.slow.test.ts` → **1 passed** (multiple cold boots incl. one at this exact HEAD); reference spec `github-connector-import.slow.test.ts` still green.
- `bun test` across 8 den-api suites (github-* + plugin-system access/create-bundle) → **36 pass, 0 fail** (new coverage: backoff bounds, transient classification, fast-ack ingress with cold fetch spy, rename/removal matching, automation-vs-human reauth).
- `tsc --noEmit` den-api clean; den-web `typecheck` clean; evals typecheck has only the pre-existing `connectors-quick-add` error.
- Migration: `pnpm --filter @openwork-ee/den-db db:generate` → `0054_mighty_scarecrow.sql` (two columns + one index).

Evidence tape published as the sticky comment (ambient testkit tape; API-driven spec — verdict comes from the tape facts).


benjaminshafii:
## Photo roll — github-sync-queues-work-retries-transient-faults-reconciles-drift-and-survives-r — 0/0 frames passed · 6 facts

✅ **0/0 frames passed · 6 facts** · 6 expectations passed · 0 failed

### ℹ️ FACT — 1. A push is acknowledged before processing and auto-updates the exact commit

Delivery self-healing-1 returned HTTP 202 with queued=true in 6ms, before the 1500ms content read; event cse_01kz845r88esjv4ft6bdsvq8h9 completed at 1634540a94eb978a591d2dc47582b013b1aac343 and stored v2.

- ✅ **PASS** A push is acknowledged before processing and auto-updates the exact commit — Delivery self-healing-1 returned HTTP 202 with queued=true in 6ms, before the 1500ms content read; event cse_01kz845r88esjv4ft6bdsvq8h9 completed at 1634540a94eb978a591d2dc47582b013b1aac343 and stored v2.

### ℹ️ FACT — 2. A rate-limited push retries itself instead of ending failed

The v3 event completed with attemptCount=2 after two provider 429s; final status=completed, never failed, and the latest source is v3.

- ✅ **PASS** A rate-limited push retries itself instead of ending failed — The v3 event completed with attemptCount=2 after two provider 429s; final status=completed, never failed, and the latest source is v3.

### ℹ️ FACT — 3. Back-to-back pushes are serialized so the newest branch head wins

The v4 event ended completed, the v5 event ended completed, and the newest version-list entry is v5; the older event did not overwrite it.

- ✅ **PASS** Back-to-back pushes are serialized so the newest branch head wins — The v4 event ended completed, the v5 event ended completed, and the newest version-list entry is v5; the older event did not overwrite it.

### ℹ️ FACT — 4. GitHub sync survives a repository rename and its next push

Stable repository id 42 moved the target remoteId to acme-eval/probe-plugins-renamed; the next event cse_01kz845wq7esjv4gnnevj50jyq completed and stored v6, covering the exact rename failure reported in #3521.

- ✅ **PASS** GitHub sync survives a repository rename and its next push — Stable repository id 42 moved the target remoteId to acme-eval/probe-plugins-renamed; the next event cse_01kz845wq7esjv4gnnevj50jyq completed and stored v6, covering the exact rename failure reported in #3521.

### ℹ️ FACT — 5. Reconciliation heals branch drift when the webhook is lost

No delivery was sent for v7; reconciliation event cse_01kz845x3kesjv4gw5ebdrc63c completed and stored v7 without human action.

- ✅ **PASS** Reconciliation heals branch drift when the webhook is lost — No delivery was sent for v7; reconciliation event cse_01kz845x3kesjv4gw5ebdrc63c completed and stored v7 without human action.

### ℹ️ FACT — 6. Sync now preserves an unchanged version and the event history tells the full self-healing story

Sync now enqueued 1 target(s), event cse_01kz845xafesjv4h1hc659htmf ended completed, versions stayed at 6, and history contains a completed push, a retried event, rename continuity, and manual reconciliation.

- ✅ **PASS** Sync now preserves an unchanged version and the event history tells the full self-healing story — Sync now enqueued 1 target(s), event cse_01kz845xafesjv4h1hc659htmf ended completed, versions stayed at 6, and history contains a completed push, a retried event, rename continuity, and manual reconciliation.

---
_Roll created 2026-08-05T04:50:52.443Z · Source: `evals/results/rolls/2026-08-05T04-50-52-443Z-github-sync-queues-work-retries-transient-faults-reconciles-drift-and-survives-r/roll.json` · Repro: `pnpm fraimz:publish -- --pr 3536 --roll 2026-08-05T04-50-52-443Z-github-sync-queues-work-retries-transient-faults-reconciles-drift-and-survives-r`_

PR #3602, opened 2026-08-07
feat(desktop): seed a chat-first default workspace when none exists

starsoul666:
## What
When a fresh desktop has no local workspace, the engine (and with it the model picker and provider list) never boots, leaving the app stuck on a "create workspace" empty state. This seeds a chat-first global default workspace — `OpenWork Chat` under the user's home — so models, providers, and ordinary chat are immediately usable without creating a workspace. Mirrors the chat-first onboarding folder in the renderer.

## How
- `apps/desktop/electron/workspace-store.mjs`: new `ensureDefaultLocalWorkspace({name, folderPath})` — returns `{state, seeded}`. Seeds only when the workspace list is **completely empty**; remote-only setups are left untouched (they boot their own engine remotely).
- `apps/desktop/electron/main.mjs`: `bootRuntimeForSelectedWorkspace` calls it before starting the server+engine and logs when a new default is seeded.
- No server-side change: the engine is started by the desktop main process, which is the branch point when no workspace is selected.

## Tests run
- `pnpm --dir apps/desktop test` → **185 pass / 1 skip (pre-existing macOS gate) / 0 fail**. 3 new cases: empty-list seeds, existing-local no-ops, remote-only no-ops.
- `pnpm --dir evals typecheck` → new `default-workspace` spec compiles clean; only pre-existing `connectors-quick-add.slow.test.ts:253` error remains (untouched).

## E2E note
Wrote `evals/specs/default-workspace.slow.test.ts` (stack lane, `OPENWORK_EVAL_APP_SPECS=1`). A local run booted the app, seeded the default workspace, and passed the frame-1 DOM assertions (no empty state, straight into the default workspace) before the screenshot-vision layer required a vision API key that isn't available in this sandbox. Full vision tape deferred until a key is available (CI/Daytona). Functional assertions are the gate; per AGENTS.md the tape is human-verification evidence.

### DIFF
--- a/apps/desktop/electron/main.mjs
+++ b/apps/desktop/electron/main.mjs
@@ -129,6 +129,11 @@ if (process.env.OPENWORK_ELECTRON_USE_MOCK_KEYCHAIN === "1") {
 const RELEASE_DOWNLOAD_BASE_URL = "https://github.com/different-ai/openwork/releases/latest/download";
 const RELEASE_PAGE_URL = "https://github.com/different-ai/openwork/releases/latest";
 const DOCS_PAGE_URL = "https://openworklabs.com/docs";
+// Chat-first global default workspace: when the user has not created or
+// connected any local workspace, the app seeds this one so the engine, model
+// picker, and provider list are available immediately (no "create workspace"
+// empty state). Mirrors the chat-first onboarding folder in the renderer.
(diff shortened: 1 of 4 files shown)


OmarMcAdam:
We had this
We removed it
Then we added it back
and removed it again

Thanks for contribution but it will stay removed for now

PR #3606, opened 2026-08-07
fix(ci): unbreak Den DB Migrate against pscale CLI v0.311+

reachjalil:
## Summary
- The Den DB Migrate workflow has failed on every run since pscale CLI **v0.311.0** shipped (2026-08-05 18:51Z — after the last green run that morning): `setup-pscale-action@v1` installs `latest`, and the new CLI changed `branch show --format json` output/exit codes, so `read_safe_migrations_state` exits 2 → the **Disable safe migrations** step hard-fails → **Apply migrations never runs**.
- Today that failure mode left migration `0056_oauth_provider_17_schema_sync` (#3603) unapplied while the code that needs it deployed — the current MCP outage window.
- Fix, three parts:
  1. **Pin the CLI to `v0.310.0`** (the version behind the last green run).
  2. **Accept both JSON shapes** in the state read (`.safe_migrations // .branch.safe_migrations // empty`).
  3. **Degrade gracefully**: if the state still can't be read, attempt the `safe-migrations disable` toggle anyway instead of aborting — the step's existing consecutive-DDL-probe loop remains the actual gate before `db:migrate` runs, so this does not weaken safety.

## Validation
- Failure signature from runs 31172119594 (both attempts): ~20× `pscale command failed (exit 2)` with `null` output in `read_safe_migrations_state`, while the `safe-migrations enable/disable` toggles in the same runs succeeded — confirming only the read regressed.
- Timeline: last green run Aug 5 09:40Z; v0.310.0 → v0.312.0 released Aug 5 16:38–18:51Z; every run since fails at the same step.

## Rollout
- After merge, trigger `workflow_dispatch` on Den DB Migrate to apply pending migration 0056 (unless it has been applied manually by then — the run is idempotent either way; drizzle skips applied migrations via the journal).

### DIFF
--- a/.github/workflows/den-db-migrate.yml
+++ b/.github/workflows/den-db-migrate.yml
@@ -100,6 +100,11 @@ jobs:
       - name: Setup pscale CLI
         id: setup-pscale
         uses: planetscale/setup-pscale-action@v1
+        with:
+          # v0.311.0+ (2026-08-05) changed `branch show --format json` output and
+          # exit codes, breaking the safe-migrations state reads below. Pin to the
+          # last version verified by a green run of this workflow.
+          version: v0.310.0
 
       - name: Disable safe migrations (allow DDL for this run only)
         run: |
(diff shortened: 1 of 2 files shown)


reachjalil:
Pushed a second commit fixing the failing **Schema and migrations in sync** check: the better-auth tables predate the migration chain, so the parity test seeds them from the current schema export before replay — and 0056 (#3603, already on dev) is the first migration to ALTER those seeded tables, causing duplicate-column errors on replay. The fix generalizes the existing `worker` column-strip into a per-table map covering 0056's additions. All 5 parity tests verified green locally against MySQL 8. Note: this check is currently red on *every* PR against dev for the same reason — merging this PR clears it repo-wide.

reachjalil:
Added the self-heal commit: the workflow now also runs twice-hourly on cron with a read-only pending gate (drizzle journal vs `__drizzle_migrations`). Idle ticks exit before touching pscale/safe-migrations; real drift re-applies within ~30 min. Combined with the earlier workflow-file path trigger, a merged-but-unapplied migration can no longer sit silently — the exact failure mode of today's outage.

OmarMcAdam (approved):

PR #3642, opened 2026-08-09
chore(agents): strict test-scenario protocol for orchestrator, trim executors

benjaminshafii:
## What

De-bloats the always-loaded agent context: AGENTS.md and the orchestrator/executor agent configs. Mechanics stay in skills; agents keep protocol and gates; AGENTS.md becomes the constitution.

### AGENTS.md (99 → 59 lines)

- Identity rewritten to the current product (per openworklabs.com and README): any-model desktop app, OpenWork MCP gateway (`ee/apps/den-api`, `api.openworklabs.com/mcp/agent`), and Den control plane — replacing the stale "control surface / Add a worker → Connect remote / messaging connectors" story. Each surface now names its repo path.
- *PR Expectations* + *Validate Every Experience* merged into one **Verification** section (they restated each other: tape decides, humans audit, Daytona-first/local-fallback, verdict rules).
- Demo-driven development compressed to two lines pointing at the `voiceover` skill.
- Coding guidelines: 5 subsections → 4 bullets, no rules lost (pnpm-only, no `any`/`as`, stack preferences, smallest diff, stop-when-overloaded).

### Agent configs

- `orchestrator.md` — removes sections that duplicated AGENTS.md, drops executor model names from the description, and adds a **Test scenarios (strict)** protocol: any "create a test scenario / test X / cover Y" request must start with a spec plan in chat (claims + negative halves, overlap check against `evals/specs/`, lane choice `*.test.ts` vs `*.slow.test.ts`, testkit resources, Daytona vs local environment, spec budget, exact run commands + verdict rules) before any file is written. Coverage-only requests stop for approval on the plan; scoping decisions are declared in the plan, not discovered later as code comments.
- `executor.md` / `executor-deep.md` — bodies were byte-identical and wordy; trimmed ~40%, `executor-deep` now states it is the escalation tier with the same contract.

## Validation

Docs + inert `.opencode/` agent config — no runtime-observable product change, so per AGENTS.md (old and new versions agree) this skips testkit runtime proof (stated explicitly here).

Behavioral check instead: ran `opencode run --agent orchestrator` from this branch with a coverage request to confirm the new protocol produces a spec plan and stops for approval without writing files. Transcript in the PR comment below.

### DIFF
--- a/.opencode/agents/executor-deep.md
+++ b/.opencode/agents/executor-deep.md
@@ -5,20 +5,16 @@ model: openai/gpt-5.6-sol-fast
 variant: xhigh
 ---
 
-You are the executor. You receive a concrete, well-specified coding task from an orchestrator and implement exactly it — no scope expansion.
+You are the escalation tier — harder tasks, same contract as `executor`.
 
-## Rules
+You implement exactly the brief you receive — no scope expansion, no re-exploration beyond what the task requires. If the brief seems wrong, ambiguous, or underspecified, say so and stop rather than improvising.
 
-- The brief gives you exact files and acceptance criteria; do not re-explore beyond what the task requires.
-- If the brief seems wrong, ambiguous, or underspecified, say so and stop rather than improvising.
-- Verify with the narrowest fast check that covers your change (e.g. `pnpm --filter <pkg> typecheck`, a targeted `test:*` script) — never repo-wide builds unless the brief asks for them.
+Verify with the narrowest fast check that covers your change (e.g. `pnpm --filter <pkg> typecheck`, a targeted `test:*` script) — never repo-wide builds unless the brief asks.
 
-## Reporting back
-
-Your final message is the only thing the orchestrator sees. Keep it under ~30 lines:
+Report back — your final message is all the orchestrator sees; keep it under 20 lines:
 
 1. Files changed: `path:line` + one-line summary each.
 2. Commands run, with exit codes.
 3. Anything skipped, assumed, or needing follow-up.
 
-On a resumed repair round, report only the delta since your last report.
+Resumed repair rounds report only the delta.
--- a/.opencode/agents/executor.md
+++ b/.opencode/agents/executor.md
@@ -5,20 +5,14 @@ model: openai/gpt-5.6-sol-fast
 variant: medium
 ---
 
-You are the executor. You receive a concrete, well-specified coding task from an orchestrator and implement exactly it — no scope expansion.
+You implement exactly the brief you receive — no scope expansion, no re-exploration beyond what the task requires. If the brief seems wrong, ambiguous, or underspecified, say so and stop rather than improvising.
 
-## Rules
+Verify with the narrowest fast check that covers your change (e.g. `pnpm --filter <pkg> typecheck`, a targeted `test:*` script) — never repo-wide builds unless the brief asks.
 
-- The brief gives you exact files and acceptance criteria; do not re-explore beyond what the task requires.
-- If the brief seems wrong, ambiguous, or underspecified, say so and stop rather than improvising.
-- Verify with the narrowest fast check that covers your change (e.g. `pnpm --filter <pkg> typecheck`, a targeted `test:*` script) — never repo-wide builds unless the brief asks for them.
-
-## Reporting back
-
-Your final message is the only thing the orchestrator sees. Keep it under ~30 lines:
+Report back — your final message is all the orchestrator sees; keep it under 20 lines:
 
 1. Files changed: `path:line` + one-line summary each.
 2. Commands run, with exit codes.
 3. Anything skipped, assumed, or needing follow-up.
 
-On a resumed repair round, report only the delta since your last report.
+Resumed repair rounds report only the delta.
--- a/.opencode/agents/orchestrator.md
+++ b/.opencode/agents/orchestrator.md
@@ -1,5 +1,5 @@
 ---
-description: Orchestrator agent. Thinks, plans, and verifies; delegates all actual coding to the executor subagents (GPT 5.6 Sol Fast, medium/xhigh tiers).
+description: Orchestrator. Plans, delegates, and verifies; never writes code. Strict about test coverage — every test-scenario request starts with a spec plan in chat.
 mode: primary
 model: anthropic/claude-fable-5
 variant: max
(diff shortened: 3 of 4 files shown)


benjaminshafii:
## Demo run — new orchestrator handling a raw coverage request

Behavioral check for this PR's protocol, run from this branch's worktree:

```
opencode run --agent orchestrator "Create a test scenario: an org admin adds an MCP connector for the whole org and a second member's desktop can use it without restarting; when the admin removes it, the member loses access too."
```

Result: the orchestrator searched `evals/specs/` for overlap, produced the spec plan below, and **stopped for approval — `git status` stayed clean** (no spec or source files written; the only diff was opencode's own plugin bootstrap touching `.opencode/package-lock.json`, reverted).

<details>
<summary>Full plan output (verbatim)</summary>

Overlap check is done and the ingredients are confirmed. Here is the spec plan — no file gets written until you approve it.

## Spec plan: org connector live add/remove propagation

**Scenario (one journey, one spec):** a member's desktop is already running → admin publishes an org-wide MCP connector server-side → that same desktop process can use it → admin deletes it → that same desktop process loses access.

### 1. Claims (each with its negative half)

| # | Must happen | Must NOT happen |
|---|---|---|
| C1 | Baseline: after the member's desktop boots and opens the connections surface, the connector name is absent from the DOM and the mock has served 0 tool calls in this run window | No pre-existing connection leaks in: `deleteConnectionsNamed(den.admin, …)` sweeps the name prefix first |
| C2 | Admin creates the connection via `createOrgConnection(den.admin, { authType: "none", credentialMode: "shared", access: { orgWide: true } })`; the running desktop shows the connection card and `readUsableConnection(member, id)` reports it usable | No restart/reload: a `window.__evalContinuity = nonce` marker set right after boot must still be present when the card appears and again at the end of the spec (a relaunch or page reload wipes it) |
| C3 | Member submits a composer task; the mock connector — the authority, not the UI — serves ≥1 `mock_echo` call whose `args.text` contains unique marker `use-<stamp>` (`toolCalls({ name, atLeast: 1, sinceIso })`) | No credit from residue: `sinceIso` + run-unique marker means calls from other runs/accounts can't satisfy the assertion |
| C4 | Admin calls `deleteConnection(den.admin, id)`; on the same still-running desktop: `readUsableConnection(member, id)` stops reporting it usable and the card leaves the connections surface | Member truly loses access: a second composer task with marker `post-<stamp>` is submitted (submission acknowledged in the UI), then a bounded 90s poll of the mock finds **zero** served calls containing `post-<stamp>`, and the count of `use-<stamp>` calls is unchanged. The mock stays alive throughout (spec-owned process), so silence proves revocation, not a dead witness |

### 2. Overlap — checked, none covers this

- `org-connector-two-members.slow.test.ts` — per-member OAuth credential isolation across two desktops; connection exists **before** any desktop launches; no removal. Not extendable: its ordering is the opposite of ours by design.
- `org-connection-lifecycle.slow.test.ts` — **member-initiated** connect/reconnect/disconnect; connection also pre-exists the desktop; admin never removes anything. Appending admin-removal would put a second scenario in one spec.
- `automations-den-hosted.slow.test.ts` — uses a no-auth org connection only as a witness for automations (this is where the `authType: "none"` precedent lives).
- `connectors-quick-add.slow.test.ts` — admin's quick-add **UI** in Den Web; no member desktop, no propagation.
- `slack-style-mcp-connector.slow.test.ts`, `library-view.slow.test.ts`, `cloud-mcp-provider-capability-submit.slow.test.ts` — API-level connector shapes / library rendering; none run the add→live-use→remove arc against a running desktop.

New spec justified: **`evals/specs/org-connector-live-propagation.slow.test.ts`**.

### 3. Lane

Stack lane, `.slow.test.ts` — the mechanism under test is cross-process propagation (Den → already-running Electron desktop → engine tool call). It cannot be exercised app-less.

### 4. Resources

- `needs()`: `model: "tool-capable"`, `env: ["OPENWORK_EVAL_DEN_API_URL"]`, `optIn: ["OPENWORK_EVAL_APP_SPECS"]` — same weight class as the automations spec, so no extra dedicated opt-in.
- `server({ place, mocks: { connector: mcpMock({ port: 3983 }) }, reuseMembers: { b: <member> } })` — port distinct from 3979/3981 used by sibling specs; executor verifies no collision.
- `app({ den, as: "b", place })` — **one** desktop, launched **before** the connection exists (this ordering is the point). Single desktop + single engine fits one resolved place, so no `OPENWORK_EVAL_DAYTONA_SANDBOX_A/B` split.
- Admin acts through `den.admin` API only — faithful to "admin uses the Den dashboard," whose UI-side add is already covered by `connectors-quick-add`.
- Not used: `chrome()`, `inviteMember()` (reuseMembers provisions the member), `faultProxy()`.

### 5. Environment

Daytona lane first when `OPENWORK_EVAL_DAYTONA=1` and credentials are available (testkit `place` resolves it). Local fallback: running Den at `OPENWORK_EVAL_DEN_API_URL` (MySQL prerequisite like sibling Den specs), `OPENWORK_EVAL_APP_SPECS=1`, and a tool-capable model key. I'll report which lane actually ran and why, per AGENTS.md.

### 6. Budget

One new spec, one scenario. The add→use→remove arc must stay in a single spec because desktop-process continuity (the `window.__evalContinuity` nonce) **is** the claim; splitting it destroys the evidence. Nothing pushes down to unit coverage — API-level add/remove visibility (`readUsableConnection`) is asserted inline as supporting checks, and pure CRUD on connections is already exercised by the API-lane siblings.

### 7. Run + verdict

```
pnpm -C evals spec:stack -- org-connector-live-propagation
```

One spec per run. Verdict rules: `Passed` only if every claim above has an observable assertion in the ambient tape; any `needs()` skip reports `Incomplete`, never passed; red goes through `diagnose-a-red-run` before any fix.

### Deliberately out of scope

- **Agent's wording after removal** — we assert the mock served nothing and the UI dropped the card, not how the model phrases "tool unavailable" (brittle).
- **Per-member OAuth dance** — `authType: "none"` keeps availability the only variable; credential isolation is `org-connector-two-members`' job.
- **Non-interference with a second connection** ("removal deletes only this one") — would double resources for a secondary claim; the id-scoped `deleteConnection` plus C4's scoped assertions carry the primary claim.
- **Admin adding via Den Web UI** — covered by `connectors-quick-add`; here the API stands in for the dashboard.

---

If this plan looks right, I'll load `write-a-spec`, delegate authoring to an executor, run it via `run-tests`, and come back with the verdict and evidence. Anything you want tightened first — e.g. a dedicated opt-in env var, OAuth instead of no-auth, or a second connection for the non-interference claim?

</details>

PR #3659, opened 2026-08-10
fix(desktop): mirror undici so the packaged embedded server boots

benjaminshafii:
## Symptom

Installed desktop app (dev tip) fails to boot the embedded server:

```
Cannot find package 'undici' imported from /Applications/OpenWork.app/Contents/Resources/app.asar/server/dist/local-managed-mcp-url-guard.js
```

The imports are static (`embedded.js → server.js → local-managed-mcp.js → local-managed-mcp-url-guard.js → "undici"`), so the entire embedded server fails to load in the packaged app — not just the managed OAuth gateway.

## Root cause

#3652 added `local-managed-mcp-url-guard.ts`, which imports the `undici` package (needed for `Agent({ connect: { lookup } })` — the DNS-rebinding-safe guarded fetch; neither Node's nor Bun's global fetch exposes that hook). It added `undici` to `apps/server/package.json`, and mirrored `@modelcontextprotocol/sdk` + `@openwork/enterprise-mcp-client` into `apps/desktop/package.json` — but missed `undici`.

The packaged app imports `server/dist` from inside `app.asar`, and electron-builder packs `node_modules` only from `apps/desktop/package.json` dependencies. Dev mode never sees this because the server resolves `undici` from the pnpm workspace, which is exactly why it slipped through.

## Fix

- `apps/desktop/package.json`: add `"undici": "6.28.0"` (identical pinned spec as `apps/server`).
- New guard test `apps/desktop/electron/server-dependency-mirror.test.mjs`: every runtime dependency of `apps/server/package.json` must appear in `apps/desktop/package.json` with the identical spec. The hand-mirroring convention already covered `better-sqlite3`, `drizzle-orm`, `jsonc-parser`, `minimatch`, `yaml`, `zod`, etc. — this pins it so the next server dep can't silently miss the asar.
- `pnpm-lock.yaml`: desktop importer entry only (undici@6.28.0 was already in the lockfile via apps/server).

## Validation

Commands run (in this branch's worktree):

1. Guard test **before** the dep fix (test written first) — **red on exactly `undici`**:
   ```
   AssertionError [ERR_ASSERTION]: Server runtime dependency "undici" must be mirrored...
   + undefined
   - '6.28.0'
   tests 1, pass 0, fail 1
   ```
2. `pnpm install` — exit 0.
3. `node --test electron/server-dependency-mirror.test.mjs` — pass (rerun independently by orchestrator).
4. `node -e "require('undici/package.json').version"` from `apps/desktop` → `6.28.0`; `apps/desktop/node_modules/undici` resolves, so electron-builder collects it.
5. `pnpm --filter @openwork/desktop test` — **187 tests, 186 pass, 1 platform skip, 0 fail** (run twice: executor + orchestrator).

## Why there is no testkit tape

The failure mode exists only inside the electron-builder-packed asar. `@openwork/testkit` specs run the server in dev-mode resolution (workspace `node_modules`), where `undici` always resolves — by construction, a dev-mode spec cannot observe this regression (that is exactly how it escaped #3652's e2e spec). The verdict here comes from the manifest guard test, red on the dev tip and green with the fix, plus the full desktop suite.

Packaged repro / final gate (release build):
```
pnpm --filter @openwork/desktop package:electron:dir
npx @electron/asar l dist-electron/mac-arm64/OpenWork.app/Contents/Resources/app.asar | grep node_modules/undici
```
Before this fix the packaged app throws `ERR_MODULE_NOT_FOUND` for `undici` at embedded-server import; after it, `node_modules/undici` ships in the asar.

### DIFF
--- a/apps/desktop/electron/server-dependency-mirror.test.mjs
+++ b/apps/desktop/electron/server-dependency-mirror.test.mjs
@@ -1,0 +1,23 @@
+// Regression: server/dist/local-managed-mcp-url-guard.js failed to resolve undici in the packaged app (PR #3652).
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { describe, it } from "node:test";
+
+const desktopPackage = JSON.parse(
+  readFileSync(new URL("../package.json", import.meta.url), "utf8"),
+);
+const serverPackage = JSON.parse(
+  readFileSync(new URL("../../server/package.json", import.meta.url), "utf8"),
+);
+
+describe("server dependency mirror", () => {
+  it("mirrors every server runtime dependency in the desktop package", () => {
+    for (const [name, spec] of Object.entries(serverPackage.dependencies)) {
+      assert.equal(
+        desktopPackage.dependencies[name],
+        spec,
+        `Server runtime dependency "${name}" must be mirrored with the same version in apps/desktop/package.json because the packaged app imports server/dist from app.asar and electron-builder only packs node_modules declared by apps/desktop/package.json.`,
+      );
+    }
+  });
+});
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -24,7 +24,7 @@
     "check:electron": "node ./scripts/check-electron-bridge.mjs",
     "typecheck:electron": "pnpm --dir ../server exec tsc -p ../desktop/tsconfig.electron.json --noEmit",
     "prepare:sidecar": "node ./scripts/prepare-sidecar.mjs",
-    "test": "node --test electron/agent-context-diagnostics-fetch.test.mjs electron/automation-runner.test.mjs electron/brand-app-name.test.mjs electron/brand-icon-darwin.test.mjs electron/brand-icon-windows.test.mjs electron/connect-link.test.mjs electron/desktop-distribution.test.mjs electron/dev-profile.test.mjs electron/electron-builder-config.test.mjs electron/linux-desktop-integration.test.mjs electron/no-bare-external-fetch.test.mjs electron/nuke.test.mjs electron/open-external.test.mjs electron/runtime.test.mjs electron/runtime-ca.test.mjs electron/runtime-chain-repair.test.mjs electron/secure-vault-key.test.mjs electron/ui-control-server.test.mjs electron/updater.test.mjs electron/remote-workspace.test.mjs electron/workspace-store.test.mjs"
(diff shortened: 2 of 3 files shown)


benjaminshafii:
### `openwork-tests` red here is pre-existing on `dev`, not from this PR

This PR touches only `apps/desktop/package.json`, a new desktop test, and `pnpm-lock.yaml` — no `apps/server` code. The failing job is **server** unit tests.

**Failing tests (both runners):**
- `runtime MCP engine sync > pushes toggled enabled state to the engine`
- `runtime MCP engine sync > disconnects a removed MCP from the engine`

**Clean control** (per `diagnose-a-red-run`) — detached worktree at unmodified `origin/dev` @ `175508512`, `bun --conditions=development test src/mcp.engine-sync.e2e.test.ts`:

| checkout | result |
|---|---|
| clean `origin/dev` @ `175508512` | **18 pass, 2 fail** |
| this branch | **18 pass, 2 fail** (identical test names) |

Same signature both sides, so this branch introduces zero new failures. `dev`'s own OpenWork Tests run at `175508512` failed; the parent commit `2db84f348` passed — so #3652 turned `dev` red.

### Root cause of that separate failure (likely a real regression)

`DELETE /workspace/:id/mcp/:name` and `POST /workspace/:id/mcp/:name/enabled` now call `deleteLocalManagedMcp` / `setLocalManagedMcpEnabled` **first and unconditionally**. Those read the encrypted vault, and `resolveVaultKey` throws when no key is available:

```
ApiError(503, "managed_mcp_secure_storage_unavailable",
  "Secure storage for OpenWork-managed MCP credentials is unavailable. Start through OpenWork Desktop or set OPENWORK_ENCRYPTION_KEY.")
```

Confirmed by toggling only that variable on clean `origin/dev`:

```
bun test src/mcp.engine-sync.e2e.test.ts                                  ->  18 pass, 2 fail   (503)
OPENWORK_ENCRYPTION_KEY=... bun test src/mcp.engine-sync.e2e.test.ts      ->  20 pass, 0 fail
```

So on any server without secure storage (the documented standalone case), **deleting or toggling any MCP — including ordinary non-managed ones — returns 503**. The managed-credential path should not be consulted for connections that aren't managed.

Tracking separately; this PR is the packaging fix only.

benjaminshafii:
The pre-existing `openwork-tests` failure described above is now fixed in #3663 (`fix(server): keep non-managed MCP delete and toggle off the credential vault`). Once that lands on `dev`, this PR's `openwork-tests` will go green. The two PRs touch disjoint files (`apps/desktop` here, `apps/server` there) and are independently mergeable.

PR #3664, opened 2026-08-10
feat(release): release:rollback stop-the-bleed script + post-publish rollback runbook

benjaminshafii:
## Why

`dev` requires reviewed PRs (SOC 2 four-eyes) — which is right — but our only documented recovery for a bad release covered the *pre-publish* case. Once a bad version was published and marked Latest, rollback ran through merge → review → rebuild, the slowest possible incident path. Rollback of already-reviewed artifacts should not require a code change at all.

## What

- **`pnpm release:rollback`** (`scripts/release/rollback.mjs`): stop-the-bleed automation. Re-points Latest to the last good published release, demotes the bad release to prerelease (so `releases/latest` and updater endpoints can never resolve it — assets and tag stay intact), and prepends a rollback banner to its notes. **Dry-run by default**, `--execute` to act, `--bad`/`--to` overrides, idempotent re-runs, never deletes anything.
- **`docs/RELEASING.md`**: new "Rolling back a published release" runbook. Key nuance it encodes: updaters never downgrade, so re-pointing Latest only protects clients that haven't updated yet; already-updated clients need a **reissue of the last-good code from its tag** (not from `dev`, which still contains the bad commit) at a higher version. Includes the npm deprecate step and the den-api install-door pin caveat.

Part of a release-resilience set with the clean-revert fast lane and the emergency-change audit trail PRs.

## Verification

- `node --test scripts/release/rollback.test.mjs` — 8/8 pass (target selection, refusals, plan construction, idempotency skips, banner build; fake `gh`, no network).
- Live **read-only** dry run against the real repo: correctly selected bad=v0.18.18 (current Latest), good=v0.18.17, printed the 3-step plan + reissue guidance, made no mutations.
- `node scripts/release/rollback.mjs --bad v0.0.1` → clean failure, exit 1.
- `pnpm release:rollback` resolves from the root (package.json wiring).

No testkit tape: release tooling + docs, not app-runtime observable. Verified via unit tests and the live read-only dry run above; the `--execute` path issues the same `gh` commands the dry run prints.

### DIFF
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -69,7 +69,59 @@ tag/dev divergence; the release waits on review latency.
 - **A tagged version turned out to be defective before publish**: leave the
   release as a draft or delete it (`gh release delete vX.Y.Z`), fix forward,
   and cut the next patch. If the bad version reached npm, deprecate it:
-  `npm deprecate openwork-server@X.Y.Z "<reason — use X.Y.Z+1>"`.
+  `npm deprecate openwork-server@X.Y.Z "<reason — use X.Y.Z+1>"`. If the
+  release was published, follow
+  [Rolling back a published release](#rolling-back-a-published-release).
+
+## Rolling back a published release
+
+Use this flow when a bad version is **published** and marked Latest.
+
+### 1. Stop the bleed
+
+Dry-run first, then execute:
+
+```bash
+pnpm release:rollback
+pnpm release:rollback --execute
+```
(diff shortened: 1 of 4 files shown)


benjaminshafii:
Release-resilience set: #3664 (rollback script + runbook) · #3665 (clean-revert fast lane) · #3666 (emergency-change policy + audit trail). Independent — mergeable in any order.

PR #3707, opened 2026-08-12
fix(server): keep running sessions alive across workspace switches

### LINKED ISSUE #3708: Workspace switch disposes the engine instance and aborts a running session
## Summary

Switching to another local workspace while a task is running **aborts the in-flight session** with an "Aborted" badge. A plain sidebar switch (no settings change) kills the running task.

## Repro

1. Start a workspace, begin a task that takes a minute or two (e.g. `sleep 60`).
2. While it is running, create/activate a second local workspace (sidebar → new workspace, or switch to an existing other local workspace).
3. Wait a few seconds; the original session now shows **Aborted** and the task is interrupted.

## Root cause

`POST /workspaces/:id/activate` (`apps/server/src/routes/workspaces.ts`) calls `reloadOpencodeEngine` whenever the *activated* workspace isn't already active. The in-place reload does `POST /instance/dispose` on the currently serving engine instance, so the engine aborts every message it is still running — including the session in the workspace the user just left. The only existing guard (`wasActive`) prevents *re-activating the same workspace*; nothing checks for live sessions.

Observed engine timeline from a real abort:

```
04:19:26.748 POST /workspaces/:id/activate        (workspace.activate audit)
04:19:26.759 engine: disposing instance
04:19:26.762 engine: error=Aborted on in-flight message
04:19:27.608 instance re-created + bootstrapped
```

Note: with `engineRollover` enabled (engine pool, #3696) live runs already survive reloads on a draining generation; the default in-place path is the one that aborts them.

## Related

- #3696 — engine rollover keeps live sessions through reloads (default off)
- #1252 — prior workspace-switch reload behavior
- #870 — why re-activating the active workspace must not dispose

ImBIOS:
## Problem

Activating a different local workspace calls `reloadOpencodeEngine`, which disposes the currently serving engine instance (`/instance/dispose`) and re-boots the engine for the newly activated workspace. Any session still running in the workspace being left flips to **Aborted** — the user loses the in-flight task on a plain sidebar switch.

Evidence from a real aborted run:

```
04:19:26.748 POST /workspaces/:id/activate        (workspace.activate audit)
04:19:26.759 engine: disposing instance
04:19:26.762 engine: error=Aborted on in-flight message
04:19:27.608 instance re-created + bootstrapped
```

The existing guard in `routes/workspaces.ts` only skips the reload when the *already-active* workspace is re-activated (`wasActive`); it never checks for running sessions.

## Fix

Fix #3708

- New `shouldDeferInPlaceEngineReload(config, workspace, hasActiveSessions)` in `apps/server/src/engine-reload-defer.ts`: with engine rollover **off** (the default), an activation that would reload the engine in place is deferred while the workspace being activated still has an active session. The idle path keeps reloading exactly as before, so switching between idle workspaces is unchanged.
- Rollover engines (`engineRollover` + engine pool) are untouched — they already keep live runs on a draining generation.
- Applies to both switch reloads (workspace activate) and the providers config reload, reusing the established `engineHasActiveSessions` probe.

## Verification

- `bun test apps/server/src/engine-reload-defer.test.ts` — 5 passing unit tests covering: busy+no-pool defers; idle reloads; pool exists → never defers; pool exists → probe not called; rollover flag alone without a pool still defers.
- Acceptance tape: `evals/specs/workspace-switch-keeps-live-sessions.slow.test.ts` — starts a long task in workspace A, creates + activates workspace B, switches back to A, and asserts the task completes with its marker and no interrupted-message error / Aborted badge.

> Note: the stack-lane tape is gated on `ANTHROPIC_API_KEY` + `OPENWORK_EVAL_APP_SPECS` and runs on Daytona; if it cannot run here the focused server suite above is the green path.

### DIFF
--- a/apps/server/src/engine-reload-defer.test.ts
+++ b/apps/server/src/engine-reload-defer.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, describe, expect, test } from "bun:test";
+
+import { shouldDeferInPlaceEngineReload } from "./engine-reload-defer.js";
+import { clearEnginePoolForConfig, setEnginePoolForConfig, type EnginePool } from "./engine-pool.js";
+import type { ServerConfig, WorkspaceInfo } from "./types.js";
+
+function fixtureConfig(engineRollover: boolean): ServerConfig {
+  return {
+    engineRollover,
+  } as ServerConfig;
+}
+
+const workspace: WorkspaceInfo = { id: "ws_test", path: "/tmp/workspace", workspaceType: "local" } as WorkspaceInfo;
+
+describe("shouldDeferInPlaceEngineReload", () => {
+  afterEach(() => {
+    clearEnginePoolForConfig(fixtureConfig(true));
+  });
+
+  test("defers when the engine has active sessions and no rollover pool", async () => {
+    const config = fixtureConfig(false);
+    const hasActiveSessions = async () => true;
+    expect(await shouldDeferInPlaceEngineReload(config, workspace, hasActiveSessions)).toBe(true);
+  });
+
+  test("does not defer when the engine is idle", async () => {
+    const config = fixtureConfig(false);
+    const hasActiveSessions = async () => false;
+    expect(await shouldDeferInPlaceEngineReload(config, workspace, hasActiveSessions)).toBe(false);
+  });
+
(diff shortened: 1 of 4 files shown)


benjaminshafii:

PR #3771, opened 2026-08-13
Keep the workspace New task plus clickable over long names

reachjalil:
## Summary

- The per-workspace New task plus is an overlay on a full-width title that is also a drag handle. On long truncated names, clicks at the plus often hit the title instead, so the row selects or starts a drag instead of creating a task.
- Raise the action overlay above the title, keep pointer events off until hover or keyboard focus, and reserve enough title padding to cover plus, overflow menu, and chevron.
- Stop gating that plus on the selected workspace's global New task disabled state. Creating a task in another workspace uses that workspace's own client; a loading or model-unavailable selected workspace should not swallow clicks on other rows.

## What changed

- Workspace action overlay: `z-10`, `pointer-events-none` until hover, focus-within, or an open menu, matching the existing session-row hover actions.
- Title padding: `pr-8` at rest for the always-visible chevron, `pr-20` on hover/focus without the 75ms padding race.
- Per-workspace plus is no longer `disabled={newTaskDisabled}`. The top-level sidebar New task row is unchanged.
- Added `data-workspace-new-task` and `evals/specs/workspace-new-task-hit-target.slow.test.ts`. The spec hovers a long workspace name, asserts `elementFromPoint` at the plus center is the plus rather than the title, then clicks it and expects a new session without collapsing the workspace.

## Why

A visible plus that does not receive the click is worse than a hidden control. Session-row hover actions already used stacking and pointer-event gating; the workspace row did not.

## Verification

- `git diff --check` — passed.
- Exact-head Electron spec: `pnpm evals workspace-new-task-hit-target` — **Incomplete**. Local spawn died before CDP: `electron-rebuild` / `node-gyp` could not reach `104.18.17.205:443`, and a retry with `OPENWORK_ELECTRON_SKIP_NATIVE_REBUILD=1` then failed because Corepack could not fetch `pnpm@11.4.0` from the npm registry. No feature assertion ran. Lane: local. Daytona was not used.

Repro for the missing tape:

```bash
OPENWORK_EVAL_APP_SPECS=1 pnpm evals workspace-new-task-hit-target
```

## Exact candidate

- Base: `84debb7d541e9db2a5bdaedee7c9f441202f4758`
- Head: `3e1b42808e446cc6b43bec7e468788b87c2c9cb4`

### DIFF
--- a/apps/app/src/react-app/domains/session/sidebar/app-sidebar.tsx
+++ b/apps/app/src/react-app/domains/session/sidebar/app-sidebar.tsx
@@ -1481,7 +1481,7 @@ function WorkspaceHeader({
         )}
       </SidebarGlyphSlot>
       <div
-        className="min-w-0 flex-1 cursor-grab touch-none transition-[padding] duration-75 active:cursor-grabbing group-hover/workspace-header:pr-14 group-has-[[data-workspace-actions]:focus-within]/workspace-header:pr-14 group-has-data-popup-open/workspace-header:pr-10 group-hover/workspace-header:group-has-data-popup-open/workspace-header:pr-14 pr-2"
+        className="min-w-0 flex-1 cursor-grab touch-none active:cursor-grabbing pr-8 group-hover/workspace-header:pr-20 group-has-[[data-workspace-actions]:focus-within]/workspace-header:pr-20 group-has-data-popup-open/workspace-header:pr-20"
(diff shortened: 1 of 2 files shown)


reachjalil:
Superseded by https://github.com/different-ai/openwork/pull/3772 on `cursor/workspace-new-task-hit-target`, rebased onto current `dev`.

PR #3802, opened 2026-08-15
feat(ci): prune old Daytona dev snapshots after each publish

benjaminshafii:
## Why
Every push to `dev` publishes a Daytona snapshot (`openwork-dev-<sha>`) and nothing ever deleted them — 18 live snapshots and counting, one per merge. (50 older ones had already been deleted by hand, which also proved deletion is safe while stopped/archived sandboxes still reference a snapshot: sandboxes keep their own disks.)

## What
- `scripts/prune-daytona-dev-snapshots.mjs` — mirrors the `update-daytona-snapshot-pin.mjs` conventions (pure exported selection fn, injectable fetch/log, argv parsing, USAGE, main-guard). Scope is strictly `<base>-dev-*`; releases and base images are structurally out of scope.
  - Protects: explicit `--keep` names (the just-pinned snapshot), the `--keep-count 5` newest, in-flight states (pulling/building/…), and snapshots referenced by any sandbox in an active state.
  - Deletes the rest oldest-first; a failed DELETE continues the batch and exits non-zero.
  - Paginates `/snapshots` (page/total) and `/sandbox` (cursor) with a hard page cap.
- `.github/workflows/dev-daytona-snapshot.yml` — final step runs the pruner after pin+redeploy on pushes, the nightly backstop, and dispatch, reporting into the step summary.

## Verification
- `node --test scripts/prune-daytona-dev-snapshots.test.mjs` → 6 pass / 0 fail (selection incl. negative halves, dry-run issues zero DELETEs, exact DELETE targeting, cross-page keep-count protection with real cursor shape, delete-failure exits non-zero).
- `node --test scripts/update-daytona-snapshot-pin.test.mjs` → 13 pass (sibling untouched).
- Live read-only dry-run against the prod Daytona org (keep=`openwork-dev-2633397`, keep-count 5):
```
summary: pruned=13 kept=5 out-of-scope=41 failed=0
```
  Keeps = pin + 4 newest; prunes exactly the 13 stale dev snapshots; all 41 release/base images untouched. The first unit-test iteration assumed `/sandbox` returned a bare array — the live dry-run caught the real `{items, nextCursor}` cursor shape, fixed in the second commit.
- Workflow YAML is inert config: validated by the live dry-run above and observable on the next dev merge / nightly run. `actionlint` flags only the pre-existing Blacksmith runner-label warning, reproduced identically on clean `origin/dev`.

## Not in scope (tracked follow-ups)
- den-api deletion of superseded per-worker sandboxes after successful recycle (~30 accumulated for one worker).
- Release snapshot (`openwork-X.Y.Z`) retention policy — left manual for rollback value.

### DIFF
--- a/.github/workflows/dev-daytona-snapshot.yml
+++ b/.github/workflows/dev-daytona-snapshot.yml
@@ -235,3 +235,32 @@ jobs:
             echo
             echo "Triggered a den-api deploy so the new pin takes effect."
           } >> "$GITHUB_STEP_SUMMARY"
+
+      # Per-push dev snapshots accumulate forever. Keep the pinned snapshot plus
+      # the 5 newest and skip snapshots referenced by active sandboxes. Old
+      # stopped/archived sandboxes do not block pruning because they keep their own disks.
+      # This also runs nightly as a pruning backstop when the snapshot already exists.
+      - name: Prune old dev snapshots
+        if: steps.channel.outputs.enabled == 'true' && steps.resolve.outcome == 'success'
+        shell: bash
+        env:
+          DAYTONA_API_KEY: ${{ secrets.DAYTONA_API_KEY }}
+          DAYTONA_API_URL: ${{ vars.DAYTONA_API_URL }}
+          DAYTONA_SNAPSHOT_NAME_BASE: ${{ vars.DAYTONA_SNAPSHOT_NAME_BASE }}
+          DAYTONA_SNAPSHOT_NAME: ${{ steps.resolve.outputs.snapshot_name }}
+        run: |
+          set -euo pipefail
+
+          result="$(node scripts/prune-daytona-dev-snapshots.mjs \
+            ${DAYTONA_API_URL:+--api-url "$DAYTONA_API_URL"} \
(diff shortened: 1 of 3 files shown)


benjaminshafii:
**Review requested — this is the one gate a bot can't clear.** This PR touches `.github/workflows/`, which warden-clearance deliberately refuses to self-approve (`guarded` verdict), so it needs an approval from someone other than the pusher.

What to look at (5 min):
- **Workflow step** (29 lines, end of `dev-daytona-snapshot.yml`): runs the pruner after pin+redeploy; key stays in env; output goes to the step summary.
- **Pruner script**: only ever deletes `openwork-dev-*` names; keeps the pinned snapshot + 5 newest + in-flight builds + anything referenced by an active sandbox. Releases/base images are structurally out of scope.
- **Evidence in the PR body**: 19/19 unit tests, plus a live read-only `--dry-run` against prod: `summary: pruned=13 kept=5 out-of-scope=41 failed=0` — exactly the 13 stale dev snapshots.

Once approved it merges clean (all other checks green).

OmarMcAdam (approved):

PR #3825, opened 2026-08-16
feat(release): one-click release cuts — diff-warden creates the tag

benjaminshafii:
## What

Completes the commit-free release flow: `pnpm release:cut` is now truly one-click. Fresh dispatch cuts create their `v*` tag as the org-owned **diff-warden** GitHub App instead of requiring a manual admin tag push.

Reuses existing infrastructure — no new apps, no new secrets:
- `WARDEN_APP_ID` (repo secret) + `WARDEN_PRIVATE_KEY` (`warden-clearance` environment) via the same pinned `actions/create-github-app-token` used by warden-clearance.
- `resolve-release` runs in the `warden-clearance` environment; its deployment policy (`dev` + `v*` tags — tag policy added) means the key is only reachable from reviewed code. A feature-branch workflow edit cannot mint the token (the smoke proved this the hard way: the first run failed with an empty key until the branch was temporarily allowed).

### Mechanics

1. `Mint tag-push token` (fresh dispatch only, `continue-on-error`) mints a diff-warden installation token.
2. The tag ref is created **via REST** (`POST /git/refs`). Empirical finding encoded in comments: ruleset bypass attributes REST ref creation to the app, but a git-over-https push with the same installation token is **not** attributed and gets rejected (GH013).
3. App-created tags retrigger the workflow via `push` (unlike GITHUB_TOKEN) — a job-level actor guard on `resolve-release` skips the duplicate run, and every downstream job skips with it.
4. No token → warning + GITHUB_TOKEN fallback + a clear error pointing at the manual admin tag push.

### Repo settings changed (already applied, admin API)

- Ruleset `Protect v* release tags`: added Integration `diff-warden` (4454147) as an always-bypass actor (the built-in GitHub Actions app is rejected by GitHub as a bypass actor — 422).
- `warden-clearance` environment: added `v*` **tag** deployment policy so tag-triggered release runs can enter the environment; the temporary smoke-branch allowance was removed after the proof.

### Trade-off stated plainly

This couples review (warden clearance) and release (tag creation) power in one app credential. Acceptable at current team size since both live in this repo's secrets; the clean end state remains a dedicated `openwork-release` app if separation of duties is ever wanted.

## Verification

| Claim | Proof |
| --- | --- |
| diff-warden token can create a `v*` tag through the ruleset (REST) and delete it | ✅ smoke run [31939099715](https://github.com/different-ai/openwork/actions/runs/31939099715) (temporary workflow, removed before review) |
| git-over-https push with the same token is rejected → REST is load-bearing | ✅ smoke run [31939024171](https://github.com/different-ai/openwork/actions/runs/31939024171) (GH013, `Cannot create ref due to creations being restricted`) |
| Duplicate Release App run from the app's tag creation is skipped by the actor guard | ✅ run [31939116994](https://github.com/different-ai/openwork/actions/runs/31939116994) — `conclusion: skipped` on `v0.0.0-tagsmoke-*` push |
| Environment gate blocks non-reviewed branches from the key | ✅ smoke failed with empty `private-key` until the branch was explicitly allowed |
| Workflow YAML | ✅ parses |

**Remaining live proof**: the first `pnpm release:cut` after merge exercises the full fresh-dispatch path end to end (recommend cutting the next patch this way).


benjaminshafii:
@OmarMcAdam — the den-api double-check from the thread, verified against live prod for this review:

- **Max version updates itself**: `latestAppVersion: "0.18.25"` live on `api.openworklabs.com/v1/app-version` — appeared within the 5-min cache of the release publishing, zero commits/deploys.
- **Array updates itself and reflects reality**: `0.18.24` correctly absent (its release is demoted to prerelease), `0.18.22` correctly absent (tag exists, no published release). The old committed inventory could not do either.
- **Org limits**: the Den Web pin dropdown reads this same array; install-door precedence unchanged (org pin max → explicit env → runtime latest), covered by green den-api suites.
- **Air-gap**: `DEN_DESKTOP_RELEASES_MODE=static` never fetches (the whole test suite runs in this mode); `github` mode degrades to the committed snapshot on any fetch failure; explicit `OPENWORK_INSTALLER_RELEASE_TAG` overrides everything.
- **Tag timing**: tags exist at the *start* of a run, published releases at the *end* — which is exactly why den keys on published releases, never raw tags (in-flight/failed cuts are invisible to users).

This PR itself is the last mile: diff-warden creates the tag via REST (bypass proven in [31939099715](https://github.com/different-ai/openwork/actions/runs/31939099715), git-push path rejected as expected in [31939024171](https://github.com/different-ai/openwork/actions/runs/31939024171), duplicate-run guard proven in [31939116994](https://github.com/different-ai/openwork/actions/runs/31939116994) → skipped). All checks green; auto-merge is armed on your approval.

OmarMcAdam (dismissed):

benjaminshafii:
Finding is correct and now addressed in-workflow: `resolve-release` begins with a **Require release permission** gate — dispatch is rejected unless the actor is a repo **admin** or listed in the `RELEASE_MANAGERS` repo variable. This restores exactly the pre-app-token bar (tag creation was admin-only via the ruleset; manual tag pushes remain ruleset-gated and skip this check).

Why this can't be sidestepped:
- the gate runs from the workflow on the dispatched ref, but a feature-branch copy with the gate removed **cannot mint the token** — `WARDEN_PRIVATE_KEY` is environment-scoped with a deployment policy of `dev` + `v*` refs only (empirically verified earlier in this PR: the smoke failed with an empty key until the branch was temporarily allowed);
- changing the gate on `dev` requires a reviewed PR;
- the gate **fails closed**: any error resolving the actor's permission aborts the run.

Verified: the permission endpoint returns `admin` for the release actor; the rejection path is structurally unreachable to prove without a write-only account dispatching, but the failure mode is a plain non-zero exit before any token exists.

OmarMcAdam (approved):

PR #3894, opened 2026-08-18
fix(updates): follow latest stable desktop releases

arikmeyer:
## Summary
- make the unrestricted desktop policy an explicit Follow latest stable mode
- keep Pin approved versions as the finite allowlist mode
- check for updates at startup and every 24 hours while the app remains open
- default automatic downloads on for new installs while preserving explicit opt-outs

## Why
A finite list becomes stale as soon as the next stable release is published. The previous checkbox-only UI made that easy to configure accidentally, and a long-running desktop checked only once per lifecycle key.

## Verification
- app updater/settings/version-gate: 34 tests passed
- Den web policy helper: 4 tests passed
- app typecheck passed
- Den web typecheck passed


OmarMcAdam:
If all are checked, it allows latest stable
If only a few are checked, then it requires admin approval for the new version
We do not have a checkbox for "Allow latest" as an option
This is a product choice.
Its eiither allow Latest & all supported
Or
restrict versions

If this is not working as described, please open an issue

PR #3895, opened 2026-08-18
fix(connect): make routine Calendar reads direct and reliable

arikmeyer:
## Summary
- deny only the observed unrelated `using-superpowers` native skill in the managed runtime
- load each remote skill at most once per task and reuse its complete body
- add a native Google Workspace agenda capability for today, tomorrow, or one local date
- use the signed-in member’s primary Google Calendar timezone by default; allow an explicit override only when requested
- advertise the exact capability in both desktop/runtime and Den MCP steering so routine agenda reads skip capability search and Calendar skill loading

## Why
A live agenda trace spent less than one second in provider tools but required three to four model rounds. A later direct-call trace proved that making timezone mandatory caused the model to guess the wrong zone. This PR removes both model decisions: one exact capability call now owns day-boundary resolution and its authoritative timezone.

## Verification
- server runtime/skill/steering tests: 35 passed
- Google Workspace capability tests: 35 passed
- Den MCP-agent tests: 17 passed
- native capability tests: 7 passed
- server and Den API typechecks passed
- OpenAPI snapshot refreshed
- `git diff --check` passed


arikmeyer:
Follow-up from live acceptance: bounded first-time Gmail and Drive reads still spent a discovery round, and one Drive turn stopped after capability search without returning data. Commit `db9156c0f` adds direct steering for the existing first-party Gmail messages and Drive files capabilities alongside Calendar; generic discovery remains unchanged for every other service. Focused server verification remains green: 35 tests plus `openwork-server` typecheck.

arikmeyer:
Delivery-boundary correction added in 6ac57a6c8: the live Den MCP descriptions now advertise the exact bounded Gmail and Drive read shortcuts too. The previous desktop-local steering did not reach Den-connected clients. Regression coverage now verifies both the execute-tool description and server initialization instructions.

arikmeyer:
Live acceptance on `v0.18.28-swup.5` showed Gmail and Drive still searched once: execute-tool and MCP initialization instructions arrive too late for the initial tool choice. Commit 7b96d825d moves the same exact shortcuts into the search tool description, the decision surface the model is selecting. Added a regression test for that surface; 82 focused Den tests and the Den build pass.

arikmeyer:
Correction: the upstream branch commit for the search-description change is `d1401e98b`.

arikmeyer:
Correction after a true OpenCode sidecar restart: the existing direct Gmail/Drive execute guidance works; the earlier search calls came from stale in-memory MCP metadata. I reverted the extra search-tool shortcut to keep this PR minimal. Fresh live checks now use exactly one native Gmail/Drive capability call.

PR #3926, opened 2026-08-19
feat(den): serve /mcp/agent from a standalone cloud-mcp app on a read replica

OmarMcAdam:
## What

Moves the `/mcp/agent` MCP gateway onto its own service, `ee/apps/cloud-mcp`, reading from a MySQL **read replica** and writing to the primary.

`/mcp/agent` is the one-URL MCP gateway that external clients (Claude Code, Codex, Cursor, ChatGPT, Zed) connect to. It exposes two tools — `search_capabilities` and `execute_capability` — plus skill/automation/connect-server resources, and resolves capabilities across six sources (Den REST catalog, native Google Workspace, external MCP connections, marketplace plugins/Programs, built-in skills, admin tools).

**den-api keeps serving `/mcp/agent` unchanged.** This is a deliberate dual-run: the advertised OAuth resource string is derived from the `/mcp/agent` path (`den-core/src/mcp/resource.ts`), and for that route the only accepted token audience is `DEN_MCP_OAUTH_RESOURCE`. Dual-running lets traffic be flipped at the ingress with **no token invalidation**. Removing it from den-api is a follow-up.

## Why it isn't a simple move

The transitive relative-import closure of `mcp/agent.ts` was **149 files** — including `auth.ts` (44 KB Better Auth config), `orgs.ts` (59 KB), 26 `capability-sources/*`, SCIM, Stripe, workers. A naive move would have relocated nearly all of den-api. Two other things also had to be solved: `getCatalog` derived the MCP catalog by introspecting den-api's **own** OpenAPI document, and `execute_capability` **writes** to the DB, which a read-only replica cannot serve.

## How

**`ee/packages/den-core` (new)** — 153 modules moved out of den-api (the closure, plus 4 more needed for middleware parity). Layout preserved, so relative imports inside the moved set are untouched. Import specifiers rewritten to `@openwork-ee/den-core/*`; **no compatibility shims** — an early shim attempt broke ~72 `mock.module("../src/db.js", …)` call sites in den-api's tests, because the shim and den-core's internal import resolved to two different module records. Rewriting the specifiers makes test mocks and den-core internals resolve to the same file, restoring mocking.

**Read/write split** — `den-db` gains an optional replica pool and an opt-in routing facade. Reads go to the replica; writes, transactions, `getConnection()`, locking reads (`FOR UPDATE`, `LOCK IN SHARE MODE`) and anything unrecognised go to the **primary** — when in doubt, primary, since a false-replica is a correctness bug while a false-primary is only a perf cost. `runWithDbRoutingContext` (AsyncLocalStorage) pins a request to the primary once it has written, so `execute_capability`'s read-after-write is safe under replication lag. This keeps the 153 den-core call sites untouched. Routing is off unless `DEN_DB_ROUTING_SERVICE=cloud-mcp`, so den-api cannot enable it by accident even if `DATABASE_REPLICA_URL` is set.

**No runtime dependency on den-api** — the catalog comes from a committed, pruned OpenAPI snapshot in den-core (416 KB), emitted by the existing `openapi:snapshot` script in the same pass as `packages/docs/openapi.json` so the two cannot diverge. A drift check gates it in CI, and a test asserts the snapshot-derived catalog is deep-equal to the live-introspected one. `registerAgentMcpRoutes(app, { catalogSource })` defaults to today's live behaviour, so den-api's call site is unchanged.

**Auth constants deduplicated** — the 8 `DEN_MCP_*` claim/audience constants now have a single declaration in `mcp/auth-constants.ts`, re-exported from `auth.ts`. Both hosts verify JWTs in-process via Better Auth; cloud-mcp does **not** fetch JWKS over the network.

`cloud-mcp` mirrors den-api's global middleware stack (observability, request-id, header sanitiser, HSTS, telemetry sanitiser, operational-error envelope, access log, CORS), reusing the implementations rather than copying them. It omits `sessionMiddleware` (neither route reads `c.get("user")`/`c.get("session")`) and runs **none** of den-api's six background loops. `/ready` probes both primary and replica and reports which failed.

## New env vars

| var | purpose |
|---|---|
| `DATABASE_REPLICA_URL` | read-replica connection string (MySQL only) |
| `DEN_DB_ROUTING_SERVICE` | must be exactly `cloud-mcp` to enable replica routing |
| `DEN_MCP_OAUTH_RESOURCE_SEEDING` | `disabled` on resource-server hosts |

`DEN_MCP_RESOURCE_URL` must stay pinned to `https://api.openworklabs.com/mcp/agent` on cloud-mcp or every existing token breaks. Documented with a warning in `ee/apps/cloud-mcp/.env.example`.

## Verification

Verdict: **Incomplete** — the code is verified against baseline, but there is no `evals/specs/**` testkit tape yet, so per `AGENTS.md` this stays a draft.

Green:

| check | result |
|---|---|
| `pnpm --filter @openwork-ee/cloud-mcp run test` | 6 pass / 0 fail |
| `pnpm --filter @openwork-ee/den-core run test` | 1 pass / 0 fail |
| `pnpm --filter @openwork-ee/den-db run test` | 23 pass / 0 fail / 3 skip |
| `pnpm --filter @openwork-ee/den-api exec tsc -p tsconfig.json --noEmit` | exit 0 |
| `pnpm --filter @openwork-ee/den-api run typecheck:automations` | exit 0 |
| den-core / cloud-mcp / den-db builds | exit 0 |
| OpenAPI drift check | passes; **proven to fail** on a perturbed snapshot |
| cloud-mcp booted, `curl`'d | `/health` 200; well-known metadata byte-identical to den-api |

den-api's 213-file bun suite, measured **per file** against baseline `5e537edee` (aggregate counts are not comparable — files that failed to load at baseline now load):

| | baseline | this PR |
|---|---|---|
| pass | 1136 | **1155** |
| fail | 218 | **199** |
| red files | 55/213 | **54/213** |

**0 regressed files.** `mcp-membership-revocation` went `0 pass / 1 fail` → **19 pass / 0 fail** (it was failing at load on a genuinely broken module boundary); it covers the audience isolation that matters most here — that public `/mcp/agent` JWTs are rejected on `/mcp` and `/mcp/admin`. All 12 OAuth-resource/audience-sensitive files are byte-identical to baseline or improved.

The 54 remaining red files are pre-existing: 40 need local `openwork_test*` MySQL schemas, 14 are pre-existing assertion failures. den-api has no `test` script and no CI leg, so this suite is a local-only lane; it was **not** wired into root `pnpm test` here, which would have newly broken it.

## Known issue to fix before review

**Import-time DB query.** Better Auth's OAuth plugin seeds the `oauthResource` table during construction, and the refactor changed load order so `auth.ts` is now imported in contexts where it previously wasn't. 45 den-api test files now perform a DB query at import time (baseline: 0). With MySQL down, `external-connection-proxy` silently goes `11 pass` → `Ran 0 tests` while bun still exits 0 — a top-level `await import` never resolves. cloud-mcp already guards this via `DEN_MCP_OAUTH_RESOURCE_SEEDING=disabled`; den-api does not. Not a test regression under the baseline environment, but it is load-order fragility this PR introduced and it should be fixed.

## Not done yet

- **Stage E — deployment.** No `packaging/docker/Dockerfile.cloud-mcp`, no Helm template, no entry in the `publish-ee-images.yml` image matrix, no Render service, `warden.toml` `desktop-den-sync-review` paths not extended to `ee/apps/cloud-mcp/**`, `turbo.json` `globalEnv` and `scripts/dev-local.mjs` not updated.
- **Stage F — testkit proof.** No `evals/specs/**` spec driving the real surface.
- The 12 `/mcp/agent` consumers that hardcode the URL (`apps/app/src/app/constants.ts:164`, `apps/server/src/cloud-mcp-health.ts`, the probe/diagnostics modules, `evals/runner/journeys/mcp.ts`, the installer-parity script, 10 docs pages) are untouched — correct for dual-run, required for cutover.
- The 149 den-core modules still import the `db` singleton directly; `readDb` exists but no call site opts into it explicitly. Routing is transparent by design.


benjaminshafii:
don't feel we should go ahead with this

PR #4009, opened 2026-08-22
Export and import sessions

### LINKED ISSUE #4008: Export and import chat sessions (per session and per workspace)
## Summary

There is currently no way to get a chat session out of OpenWork, or to bring one back in. Sessions live only in the local OpenCode SQLite DB, so a conversation cannot be shared with a teammate, attached to a bug report, moved between machines, or kept as a record after a workspace is deleted.

This proposes **session export and import**, at two levels:

- **Per session** - export a single session.
- **Per workspace** - export every session in a workspace in one bundle.

## Motivation

- **Sharing.** "Here is the session where the agent did X" is currently impossible without screenshots.
- **Portability.** Moving to a new machine or a new workspace loses all session history.
- **Bug reports.** Maintainers frequently need a full transcript, including tool calls, to diagnose agent behaviour.
- **Retention.** Sessions are only as durable as the local OpenCode DB.

## Proposed formats

Two formats, one canonical:

| Format | Purpose | Round-trips |
|---|---|---|
| **JSON** | Canonical bundle. What import consumes. Lossless: session metadata, messages, parts, todos. | Yes |
| **Markdown** | Human-readable transcript for sharing, pasting into issues/PRs, or review. | No (export only) |

A single bundle shape covers both levels - per-workspace export is the same envelope with more entries in `sessions` - so there is exactly one schema and one import path:

```jsonc
{
  "format": "openwork.session-export",
  "version": 1,
  "exportedAt": "2026-01-01T00:00:00.000Z",
  "workspace": { "id": "..." },
  "sessions": [
    { "session": { "id": "...", "title": "..." }, "messages": [ { "info": {}, "parts": [] } ], "todos": [] }
  ]
}
```

## Proposed implementation

This follows existing precedent in the repo rather than inventing a parallel mechanism.

**Export**
- Source the data from the existing snapshot read model, `buildSessionSnapshot` (`apps/server/src/session-read-model.ts`), already surfaced by `GET /workspace/:id/sessions/:sessionId/snapshot` (`apps/server/src/routes/sessions.ts`).
- New routes alongside it:
  - `GET /workspace/:id/sessions/:sessionId/export`
  - `GET /workspace/:id/sessions/export` (whole workspace)
- Both go through `requireWorkspaceSession` / workspace directory fencing so a session in another workspace can never be read.

**Import**
- `POST /workspace/:id/sessions/import`
- Creates each session via the OpenCode SDK, then seeds messages using the existing `seedOpencodeSessionMessages` primitive (`apps/server/src/opencode-db.ts`) - the same mechanism `materializeBlueprintSessions` already uses.
- Import is **additive**: it always creates new sessions and never overwrites an existing one.

**Secret safety**

Transcripts routinely contain file contents, tool output, and credentials. Export must reuse the existing workspace-export safety contract (`apps/server/src/workspace-export-safety.ts`) rather than dumping raw snapshots:
- a `sensitive` mode (`auto` / `include` / `exclude`),
- redaction of detected secrets by default,
- warnings surfaced to the user before the file is written.

**Surfaces**
- Session context menu → *Export session*.
- Workspace menu → *Export sessions* / *Import sessions*.
- Matching `session.export` / `session.import` control actions so the flow is drivable from automations and E2E specs.

## Out of scope

- Merging into or overwriting an existing session.
- Cross-version migration of bundles (bundle carries `version` for future use).
- Publishing sessions to a hosted/share URL.

g3saranath:
Closes #4008

## Why

Sessions live only in the local OpenCode database. A conversation could not be shared with a teammate, attached to a bug report, moved between machines, or kept after a workspace was removed.

## What

Export and import at two levels — **per session** and **per workspace** — using one bundle envelope, so there is a single schema and a single import path.

| Format | Purpose | Round-trips |
|---|---|---|
| JSON | Canonical bundle. The only thing import consumes. | Yes |
| Markdown | Read-only transcript for sharing. | No |

**Routes** (`apps/server/src/routes/sessions.ts`)
- `GET /workspace/:id/sessions/:sessionId/export`
- `GET /workspace/:id/sessions/export` — whole workspace
- `POST /workspace/:id/sessions/import`
- `GET /workspace/:id/session-imports` — provenance for imported sessions

**In the app**
- Every session offers *Export session* → JSON or Markdown, from both the dropdown and right-click menus.
- Workspace menu gains *Export all sessions* and *Import sessions*.
- `session.export` / `session.import` control actions, so the flow is drivable from automations and specs.

## Design decisions worth reviewing

**Secret safety.** Transcripts routinely carry tool output and credentials, so export reuses the existing workspace-export contract: `auto` refuses to guess (409 naming the offending session), `exclude` redacts, `include` is an explicit opt-in. Redaction is deliberately **value-level** — a token is replaced while the surrounding sentence survives, unlike config export which drops whole values. The app always requests `exclude`.

**Import is additive and lossy by choice.** It always creates new sessions and never overwrites one. It replays the transcript as text through the existing `seedOpencodeSessionMessages` primitive rather than reconstructing raw part payloads from an untrusted file, which would risk writing a session the UI cannot render. The JSON bundle remains the lossless record.

**Imported sessions are permanently read-only, enforced on the server.** Sessions belong to OpenCode, so provenance is recorded in the per-workspace runtime store (the same `workspace-kv-store` pattern behind session groups). The rule is enforced in `assertSessionNotReadOnly` on the OpenCode proxy that every writer passes through — so automations, control actions, and any other MCP/OpenCode client are held to it, not just the composer. A badge with a disabled composer alone would have been cosmetic.

Management actions stay available on purpose: rename, archive, group, fork and delete still work, and deleting an imported session drops its provenance so marks cannot outlive their sessions.

**Route ordering.** `GET /sessions/export` is registered before `GET /sessions/:sessionId` so the literal segment is not captured as a session id — the same ordering `/session-groups/reorder` already relies on.

## Verification

Lane: **local**. Daytona was not used — no Daytona credentials available in this environment. This spec is app-less (`pr` lane) and needs neither Daytona nor a Den/MySQL stack.

**Feature spec — `evals/specs/session-export-import.test.ts`**

```
$ pnpm evals:pr specs/session-export-import.test.ts
 ✓ |pr| specs/session-export-import.test.ts (8 tests) 254ms
 Test Files  1 passed (1)
      Tests  8 passed (8)
```

8 passed, 0 failed, **0 skipped**. Each claim has an observable assertion plus its negative half:

| Claim | Negative half |
|---|---|
| A session exports as a versioned bundle, and as readable Markdown | Markdown carries no raw JSON part scaffolding |
| A transcript holding a token is refused by default (409, names the session) | `exclude` removes the token but keeps "Deploy finished."; `include` returns it untouched |
| A workspace export carries every session | A session in another directory is 404, not exported |
| A bundle imports back, persisted in order in the engine DB | The written session id differs from the source, so import cannot clobber history |
| An imported session records the workspace it came from | — |
| Writing to an imported session is refused (409 `session_read_only`, engine saw zero prompts) | The same prompt on a normal session succeeds and reaches the engine once |
| Imported sessions are partitioned out of the working list | A workspace with no marks keeps every session visible |
| Export filenames derive from the title | `../../etc/passwd` yields no path separator |

**Regression suites**

```
$ pnpm --filter openwork-server test     1590 pass, 3 fail
$ pnpm --filter @openwork/app test        850 pass, 0 fail
$ pnpm --filter openwork-server typecheck  clean
$ pnpm --filter @openwork/app typecheck    clean
$ pnpm evals:pr specs/test-terminology.test.ts   1 passed
$ pnpm evals:pr specs/spec-impact.test.ts        3 passed
```

The 3 server failures are **pre-existing on `dev`**, confirmed by stashing this branch and re-running on a clean tree: baseline is `1557 pass / 3 fail` with the identical set — `serve > handles a malformed raw Node TRACE request` (x2) and `ensureWorkspaceFiles > uses shipped extension preview plugin`. Neither touches session transfer. This branch adds 33 passing server tests and introduces no new failures.

`pnpm evals:typecheck` is already red on `dev` for unrelated specs; the new spec contributes no errors to it.

## Notes for reviewers

- Tests isolate `OPENWORK_RUNTIME_DB` and `OPENCODE_DB` to temp files, so no test writes to a real profile.
- I could not run the Electron E2E lane here; the sidebar/composer surfaces carry stable hooks (`data-session-imported-tag`, `data-session-imported-source`, `data-sidebar-imported-section`, `data-session-read-only-notice`, `data-session-menu-export-json`, `data-session-menu-export-markdown`) so an `.e2e.test.ts` can drive them if you'd like that added.
- Only `en.ts` has the new strings; other locales fall back, matching how this repo already handles new keys.


g3saranath:
## Verified in the running desktop app

Exercised the full flow against a real workspace and a real OpenCode database, not just the test harness.

**Export** — all routes 200 with real data:

| Call | Result |
|---|---|
| `GET /sessions/:id/export` | `200` — bundle with `workspaceName: "ws1"` |
| `GET /sessions/:id/export?format=markdown` | `200` — readable transcript |
| `GET /sessions/export` | `200` — whole workspace in one envelope |
| `GET /session-imports` | `200` |

**Import round trip** — bundle exported, then imported back:

```json
{ "importStatus": 201,
  "imported": [{ "sourceSessionId": "ses_fd683e256ffe…", "sessionId": "ses_fd62f7402ffe…",
                 "title": "Greeting message", "messages": 2 }],
  "mark": { "sourceWorkspaceId": "ws_027ef420cfd5", "sourceWorkspaceName": "ws1", … } }
```

The imported messages read back with their original content, so the transcript is genuinely persisted in the engine database rather than just echoed in the response.

**Read-only enforcement** — prompting the imported session through the OpenCode proxy:

```
409  {"code":"session_read_only","message":"This session was imported and is read-only. …"}
```

**UI** — verified via CDP against the live window:
- `[data-sidebar-imported-section]` renders as an `Imported 1` section at the bottom.
- The imported session is excluded from the working list (only the two original sessions remain there).
- `[data-session-imported-tag]` renders `Imported ws1` with `title="Imported from ws1"`.
- `[data-session-read-only-notice]` renders in the composer, which is disabled.

## One fix from this pass (915fbee)

The read-only notice concatenated the explanation and the hint into a single inline string, which truncated mid-sentence in the composer status row — `"…and is read-only. Start a ne…"`. The explanation now renders inline (`truncated: false` measured via `scrollWidth`/`clientWidth`) and the longer guidance moved to the tooltip.

Re-ran after the fix: `pnpm --filter @openwork/app typecheck` clean, `pnpm --filter @openwork/app test` 850 pass / 0 fail.

g3saranath:
## Update: an imported session now reads exactly like the original

Three gaps surfaced while driving this in the desktop app.

### 1. Import lost the shape of the conversation

Import replayed transcripts as flattened text, so reasoning was folded into the reply and tool calls became `[tool: name]` placeholders. Message and part payloads now cross over as exported, via a new `writeOpencodeSessionMessages`. Ids are reassigned and `parentID` links rewritten onto the new ids so a message never points at an unrelated message. Parts describing a live run (`step-start`, `step-finish`, snapshots, patches) are dropped, since re-inserting them would describe work that never happened in the destination.

`seedOpencodeSessionMessages` is untouched, so blueprint sessions keep their behaviour.

Measured on a real transcript with tool calls:

```
source:   text, step-start, reasoning, text, step-finish, text, step-start,
          reasoning, tool, step-finish, step-start, reasoning, tool, ...
imported: text, reasoning, text, text, reasoning, tool, reasoning, tool, ...
```

The spec now asserts this: part kinds round-trip as `["text", "reasoning", "text"]` and `step-start` is absent.

*Known difference:* the tool-run summary reads **"5 steps"** where the original said **"Worked for 48s"**. Step-finish timing is dropped, so duration can't be computed — arguably correct, since that work didn't take 48s in the destination workspace.

### 2. The sidebar named two different workspaces on one row

The heading named the workspace a session now lives in; a tag named the one it came from. The Imported section is now grouped by **source** workspace, keyed by id so two workspaces sharing a name stay distinct, and the redundant tag is gone.

Each group uses the same `WorkspaceAvatar` as the Workspaces list, so a workspace keeps one colour everywhere — verified `ws1` is `rgb(226, 59, 76)` in both places — and collapses like a normal workspace (2 visible → 0 collapsed → 2 expanded).

The section renders even when empty, because importing is only offered there; hiding it with no imports would leave no way to make the first one.

### 3. Read-only was only skin deep

Only the text input and send button were disabled. Everything else was live:

| Control | Before | Now |
|---|---|---|
| Tools menu, attachments, agent picker, model picker | enabled | disabled |
| Placeholder | "Describe your task…" | "This imported session is read-only" |
| **Revert** | **offered** | hidden |
| Edit message | offered | hidden |
| Copy, Branch in new chat | offered | unchanged |

Revert mutates a transcript the server already refuses (`409 session_read_only`), so there was no data risk — but it was a button guaranteed to fail. Copy and Branch stay: neither touches the transcript, and branching is the intended way to carry an imported conversation forward.

## Verification

```
pnpm --filter openwork-server test    1627 pass, 3 fail (pre-existing)
pnpm --filter @openwork/app test        850 pass, 0 fail
pnpm evals:pr specs/session-export-import.test.ts   8 passed, 0 skipped
server + app typecheck                clean
```

The 3 server failures are the same ones present on `dev` (`serve > malformed raw Node TRACE` x2, `ensureWorkspaceFiles > shipped extension preview plugin`). A 4th, `mcp.authorization-link`, appeared in one run; I re-ran it against a stashed clean tree and it timed out identically there, so it is pre-existing and environment-dependent, not from this branch.

### One test was updated, deliberately

`composer-model-controls.test.ts` pinned the literal `disabled={props.steering}`, which now reads `disabled={props.steering || props.readOnly}`. The assertion was updated but its real guarantee was kept and widened: `expect(modelSelect).not.toContain("props.busy")` (previously `"disabled={props.busy}"`), so "model controls must not be disabled during ordinary generation" is still enforced and harder to sidestep.

PR #4017, opened 2026-08-23
fix(den-web): call Den API origin directly

OmarMcAdam:
## Summary
- route den-web browser API/auth calls to the matching api.* origin instead of the same-origin proxy
- convert legacy /api/den/* compatibility route to 307 redirects that preserve method/body
- keep /api/auth/* proxy in place for auth callback compatibility

## Tests
- pnpm --filter @openwork-ee/den-web exec bun test --conditions development tests/den-api-origin.test.ts app/api/_lib/den-api-redirect.test.mjs tests/mcp-tool-error-attribution.test.ts
- pnpm --filter @openwork-ee/den-web typecheck
- pnpm --filter @openwork-ee/den-web test


OmarMcAdam:
Addressed the Warden finding: the /api/den redirect now constructs the URL from the fixed API origin first and assigns pathname separately, so a double-slash suffix stays on api.app.openworklabs.com instead of becoming a network-path authority. Added regression coverage for /api/den//evil.example/path.\n\nValidation:\n- pnpm --filter @openwork-ee/den-web exec bun test --conditions development app/api/_lib/den-api-redirect.test.mjs tests/den-api-origin.test.ts\n- pnpm --filter @openwork-ee/den-web typecheck

PR #4069, opened 2026-08-25
fix(den-api): enforce production build contract

benjaminshafii:
## Summary

- replace Den API hand-maintained production dependency builds with pnpm graph traversal
- fail builds when a production workspace export target is missing
- make Docker invoke the same Den API build and expand image-smoke path coverage
- preserve the externally consumed focused MCP Apps build utility
- add production-condition startup proof and the deployment incident runbook

Partially addresses #4064. External Render notification and on-call configuration is tracked in #4070.

## Verification

Local fallback lane; these app-less artifact contracts do not require Daytona services. All commands ran on clean PR head `f7eb2d4d8`.

- `pnpm --filter @openwork-ee/den-api exec bun test test/sentry-sourcemaps-build.test.ts` (4 passed, 0 failed, 0 skipped)
- `pnpm --filter @openwork-ee/den-db exec tsx --test test/migration-readiness.test.ts` (8 passed, 0 failed, 0 skipped)
- `pnpm evals:pr specs/den-api-production-package.test.ts` (1 passed, 0 failed, 0 skipped)
- `pnpm evals:pr specs/mcp-agent-standalone-get.test.ts` (1 passed, 0 failed, 0 skipped)
- `pnpm evals:pr specs/connect-flow-mcp-apps.test.ts` (1 passed, 0 failed, 0 skipped)
- `pnpm evals:pr specs/skill-created-mcp-app.test.ts` (2 passed, 0 failed, 0 skipped)

Verdict: Passed for repository-owned build, startup, and focused MCP build compatibility claims. Testkit evidence is published on this PR. The image workflow remains the MySQL-backed `/health` proof.

## Rollback

Revert this PR to restore the explicit production package build commands. No persisted data or schema changes are included.

PR #4079, opened 2026-08-25
test(evals): resolve grant-liveness witnesses from source exports

reachjalil:
## Summary

- fix the `openwork-tests` failure on the `dev` push after #4055 merged: `specs/mcp-grant-scoped-liveness.test.ts` failed with `FileNotFound while resolving package '@openwork-ee/utils/den-urls'` in its final witness
- the witnesses resolved `@openwork-ee/utils` and `@openwork-ee/den-db` through built dist output and prebuilt it at spec start, but other specs in the parallel eval lane (notably #4069's den-api production-package build) rebuild that shared output concurrently, and tsup removes dist before rewriting it — a witness that resolves during that window fails with a missing module
- run the witnesses with `bun --conditions=development` so both packages resolve their sources directly, removing the dependency on generated build output entirely (the same approach `specs/mcp-stateless-2026-transport.test.ts` already uses); drop the now-unnecessary prebuild step

## Verification

- deleted `ee/packages/utils/dist` and `ee/packages/den-db/dist`, then ran `pnpm evals:pr specs/mcp-grant-scoped-liveness.test.ts` — passed 1/1 with all five witnesses green and no dist present, proving the spec no longer depends on shared build output

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -137,6 +137,12 @@ jobs:
         run: pnpm --dir evals install --frozen-lockfile --prefer-offline
 
       - name: Run eval spec lane
+        # The 3-core macOS runner cannot host three parallel eval workers plus
+        # their build subprocesses: worker RPC misses vitest's deadline and the
+        # lane fails with "Timeout calling onTaskUpdate" after every test
+        # passed. Cap it like other constrained environments.
+        env:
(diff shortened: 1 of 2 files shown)


reachjalil:
Superseded by #4078, which landed the same grant-liveness source-resolution fix and resolved the macOS lane's onTaskUpdate hang at its root by making the den-api production build an async spawn. The dev push for 148c7a37d is green on all lanes. Keeping the macOS two-worker cap from this branch in reserve if the timeout signature ever reappears.

PR #4112, opened 2026-08-26
fix(server): fetch live models.dev catalog so freshly released models are usable

### LINKED ISSUE #4035: [Bug]: Opencode go models are outdated (do not update automatically)
### Summary

Freshly installed openwork (providers automatically detected from opencode config). There are no recently released models (such as ox-alpha or GLM-5.3). These models are visible in opencode itself (opencode desktop). I've tried disconnecting and reconnecting opencode go provider, but it doesn't work

### To Reproduce

1. Connect opencode go
2. There is neither ox-alpha, nor GLM-5.3 

### Expected behavior

All new models would be fetched instantly and be available to use

### Actual behavior

There are no new models (such as GLM-5.2 or ox-alpha)

### Screenshots (optional)

Image

_No response_

_No response_

lucad87:
## Summary

The managed engine was pinned to the first-party model catalog mirror (`https://models.openworklabs.com/`, passed as `OPENCODE_MODELS_URL`), which is a static snapshot of models.dev rebuilt only when someone runs the `update-models.yml` workflow by hand (it has no schedule). Any model released after the last manual run never reaches the engine's provider list and therefore cannot be picked or used — for example `openrouter/stealth/ox-alpha`, or `opencode/x-preview-f-free` on OpenCode Zen. This is the stale-catalog behavior reported in #4035 ("Opencode go models are outdated (do not update automatically)"): vanilla OpenCode shows those models because it reads models.dev live.

## Changes

- `apps/server/src/opencode-models-url.ts`: outside development, resolve the engine catalog from live `https://models.dev/` using a 1s reachability probe, falling back to the first-party mirror when models.dev is unreachable (corporate proxies / outages). The explicit `OPENCODE_MODELS_URL` override and the dev-mode local inference stack probe keep their current priority.
- `.github/workflows/update-models.yml`: add a daily `schedule` so the fallback mirror stays fresh instead of advancing only on manual dispatch. `BASE_BRANCH` now falls back to `dev` because `inputs.base_branch` is empty on `schedule` events.
- `apps/server/src/opencode-models-url.test.ts`: cover the new production resolution order.

Resolution order after this change:

1. explicit `OPENCODE_MODELS_URL` override (unchanged; air-gapped escape hatch)
2. dev mode: local inference stack probe at `localhost:8791` (unchanged)
3. production: live `https://models.dev/api.json` probe (1s timeout)
4. fallback: first-party mirror `https://models.openworklabs.com/`

## Test evidence

```
bun test apps/server/src/opencode-models-url.test.ts

(pass) resolveOpencodeModelsUrl > honors an explicit catalog URL
(pass) resolveOpencodeModelsUrl > uses the live models.dev catalog outside development
(pass) resolveOpencodeModelsUrl > falls back to the mirror when models.dev is unreachable
(pass) resolveOpencodeModelsUrl > falls back to the mirror when models.dev errors
(pass) resolveOpencodeModelsUrl > uses the local catalog when the development server is available
(pass) resolveOpencodeModelsUrl > falls back to the mirror when the development server is unavailable

6 pass, 1 fail, 7 tests
```

The single failure is the `startEmbeddedServer managed OpenCode models URL` case: it spawns a fake engine child process and fails on Windows with "Managed OpenCode process did not exit after SIGKILL" (`apps/server/src/managed-opencode.ts:83`). That failure is environment-specific and pre-existing: stashing this branch and re-running on pristine HEAD reproduces the same `4 pass / 1 fail` result, so it is unrelated to these changes.

Fixes #4035

### DIFF
--- a/.github/workflows/update-models.yml
+++ b/.github/workflows/update-models.yml
@@ -1,6 +1,10 @@
 name: Update Models
 
 on:
+  schedule:
+    # Keep the first-party catalog mirror fresh; without this the mirror only
+    # updates when someone runs the workflow by hand.
+    - cron: "17 6 * * *"
   workflow_dispatch:
     inputs:
       base_branch:
@@ -20,7 +24,7 @@ concurrency:
 env:
   MODELS_PATH: ee/apps/inference/src/models/base.json
   MODELS_URL: https://models.dev/api.json
-  BASE_BRANCH: ${{ inputs.base_branch }}
+  BASE_BRANCH: ${{ inputs.base_branch || 'dev' }}
   BRANCH_NAME: automation/update-models-${{ github.run_id }}-${{ github.run_attempt }}
 
 jobs:
--- a/apps/server/src/opencode-models-url.test.ts
+++ b/apps/server/src/opencode-models-url.test.ts
@@ -39,8 +39,27 @@ describe("resolveOpencodeModelsUrl", () => {
     })).toBe("https://catalog.example.test/models");
   });
 
-  test("uses the production catalog outside development", async () => {
-    expect(await resolveOpencodeModelsUrl({ env: {} })).toBe("https://models.openworklabs.com/");
+  test("uses the live models.dev catalog outside development", async () => {
+    expect(await resolveOpencodeModelsUrl({
+      env: {},
+      fetchModels: async () => new Response(null, { status: 200 }),
+    })).toBe("https://models.dev/");
+  });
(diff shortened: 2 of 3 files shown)


OmarMcAdam:
Appreciate the PR, this is by choice for caching of large files - both so we dont kill OC's caches and we dont kill our own.
Will implement automated update flow instead
Thanks for the contribution

PR #4119, opened 2026-08-26
fix(chat): keep interleaved thoughts chronological between tool aggregates

reachjalil:
## What changes for users

During a long agent run, the chat used to show one "Thought" line at the top and then an ever-growing pile of "Ran N commands" — any thinking the model did *between* tool calls was shuffled below the pile and squashed into a single collapsed line. A 5-minute run looked like the agent had stopped thinking and was blindly running commands.

Now the transcript reads like a work log, in the order things actually happened:

> Thought ▸
> Ran 2 commands
> Thought ▸
> Ran 3 commands
> Thought ▸
> …

Each mid-run thought appears right where it happened, between the tool batches it separates.

- **Trust:** long runs stop feeling like a black box — you can see the agent deliberating, not looping.
- **Debuggability:** when a run goes sideways, click the exact Thought before the bad command batch to see why it did that.
- **No regression for the quiet mode:** with "Show model reasoning" off, the old compact view is unchanged — one tidy aggregate line.

Out of scope here: thoughts stay collapsed by default (auto-expand / live-thought ticker is a follow-up), and this cannot show thinking a model never emitted between calls (interleaved-thinking model config is a separate track).

## Problem

When a thinking model alternates thoughts and tool calls (R T T R T T…), `getAssistantRenderGroups` deliberately let tool calls skip past reasoning groups (`groups.findLast(kind !== "reasoning")`). Two artifacts followed:

- every later call was absorbed into the aggregate **above** the first thought, and
- all interleaved thoughts merged into **one** collapsed "Thought" line sitting **below** the growing aggregate.

A long run with plenty of thinking rendered as a couple of tool groups and nothing else — out of chronological order, with the thinking effectively invisible.

## Change

- A visible (non-empty) reasoning part now ends the current tool aggregate; the next aggregatable calls start a new aggregate after it. The transcript reads in the order the model produced it: thought, calls, thought, calls.
- Hidden reasoning (Show model reasoning off) and whitespace-only reasoning still form no group, so those runs keep one compact aggregate exactly as before.
- The grouping import chain (`chat/utils` → `lib/tool-aggregate` → `lib/tool-activity` → `components/tools/path`) switched from `@/` aliases to relative imports so the evals spec can exercise the real function from the alias-free test workspace. No behavior change.

## Proof

- `evals/specs/chat-thought-chronology.test.ts` (PR lane, app-less): asserts the R T T R T T shape renders `reasoning, tool-aggregate(c1,c2), reasoning, tool-aggregate(c3,c4)` with thoughts in model order; negative halves assert hidden thinking keeps one four-call aggregate and whitespace-only reasoning does not fragment the run.
  - `pnpm evals:pr specs/chat-thought-chronology.test.ts` → 1 passed / 0 failed / 0 skipped (local PR lane; spec is app-less, no Den or app surface required).
- Unit tests updated to the new contract: `apps/app/tests/chat-message-grouping.test.ts`, `apps/app/tests/message-list-step-fold.test.tsx`.
  - `bun test tests/chat-message-grouping.test.ts tests/message-list-step-fold.test.tsx` → 13 pass / 0 fail.
- `pnpm --dir apps/app typecheck` → clean.
- Full `bun test --isolate tests/`: 991 pass / 5 fail; the same 5 failures reproduce on pristine `origin/dev` (989 pass / 5 fail, same command) in unrelated areas (connect capability inventory, cloud workspace overlay, session provider auth sync, extensions sidebar) — pre-existing suite interference, not introduced here.

### DIFF
--- a/apps/app/src/components/chat/utils.ts
+++ b/apps/app/src/components/chat/utils.ts
@@ -1,6 +1,8 @@
 import { isReasoningUIPart, isToolUIPart, type DynamicToolUIPart, type FileUIPart, type ToolUIPart, type UIMessage } from "ai"
 import type { ThreadStatus } from "@/lib/messages"
-import { isAggregatableToolPart } from "@/lib/tool-aggregate"
+// Relative so the pure grouping contract stays importable from alias-free
+// test workspaces (evals specs import this module directly).
+import { isAggregatableToolPart } from "../../lib/tool-aggregate"
 
 const DOCX_MIME = "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
 const PPTX_MIME = "application/vnd.openxmlformats-officedocument.presentationml.presentation"
@@ -295,12 +297,15 @@ export function getAssistantRenderGroups(
 
     if (isToolUIPart(part)) {
       // Paper aggregation rule: consecutive command/edit/read/search calls
-      // collapse into one aggregate group. Prose, files, and other tools
-      // break the run; reasoning does not — thinking models emit a
-      // reasoning part before nearly every call, and letting it split the
-      // run degrades every aggregate to a single call.
+      // collapse into one aggregate group. Prose, files, other tools, and
+      // visible reasoning all break the run: a thought stays between the
+      // calls it separated, so the transcript reads chronologically instead
+      // of absorbing later calls into the aggregate above the thought.
+      // Hidden reasoning (showThinking off) and whitespace-only reasoning
+      // never form a group, so they keep the run — and its compact
+      // aggregate — intact.
       if (isAggregatableToolPart(part)) {
-        const previous = groups.findLast((group) => group.kind !== "reasoning")
+        const previous = groups.at(-1)
         if (previous?.kind === "tool-aggregate") {
           previous.parts.push(part)
         } else {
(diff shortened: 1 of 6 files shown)


reachjalil:
## Test evidence — interleaved-thoughts-render-chronologically-between-tool-aggregates — 0/0 screenshots passed · 1 assertion

### ℹ️ ASSERTION — 1. Interleaved thoughts stay chronological between tool aggregates

An R T T R T T turn rendered as thought, two-call aggregate, thought, two-call aggregate in model order; hiding thinking kept one four-call aggregate; whitespace-only reasoning did not fragment the run.

- ✅ **PASS** Interleaved thoughts stay chronological between tool aggregates — An R T T R T T turn rendered as thought, two-call aggregate, thought, two-call aggregate in model order; hiding thinking kept one four-call aggregate; whitespace-only reasoning did not fragment the run.

---
_Test run created 2026-08-26T18:35:19.560Z · Source: `evals/results/test-runs/2026-08-26T18-35-19-559Z-interleaved-thoughts-render-chronologically-between-tool-aggregates/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4119 --test-run 2026-08-26T18-35-19-559Z-interleaved-thoughts-render-chronologically-between-tool-aggregates`_

PR #4145, opened 2026-08-26
Make Cloud extension status reflect real connectivity

reachjalil:
## What I noticed

The Extensions and Connect surfaces could claim an MCP connection that the engine never verified. Two independent problems combined into that lie: first, a successful-looking OAuth transition set "connected" optimistically even though the engine's status snapshot is eventually consistent and can key servers differently than the UI configured them; second, `refreshMcpServers` had no ordering — a slow refresh, a post-heal status read, or a logout-time status read could complete late and overwrite a newer verified result with stale data.

## What I changed

All MCP status publication now flows through one small, ordered synchronizer owned by the connections store:

- Every refresh takes a monotonic per-workspace-context token. A completion whose token is no longer current is dropped entirely — a stale asynchronous result can never overwrite newer verified state. The self-heal path re-enters the ordered refresh instead of writing its own late status, and the add/logout paths stop doing their own unordered status writes.
- A verified OAuth completion (the modal observed the engine report `connected`) is recorded as a bounded grace: it survives the engine's delayed or missing snapshot for a few refreshes and converges to engine truth, then expires. It is never created from the OAuth redirect alone.
- Real terminal engine verdicts — `needs_auth`, `failed`, `reconnect_required`, `needs_client_registration` — always override the grace immediately, so reauthentication and failure states stay honest.
- Logout, workspace switches, and entry removal clear the grace; a foreign workspace can never inherit another workspace's green status.
- Status lookup resolves identity aliases (id, configured name, canonical server name) conservatively: an ambiguous normalized collision resolves to nothing rather than borrowing a colliding foreign server's status. The modal and the store now share one canonical name derivation instead of two divergent slug functions.

## Why this matters

Connection state is the thing users act on: a false "connected" hides a broken integration until an agent call fails, and a stale "disconnected" pushes users into needless re-auth loops. This makes the displayed state derive from verified engine connectivity with explicit ownership and ordering, covering OAuth completion, delayed snapshots, differently keyed snapshots, reauthentication, logout, and out-of-order concurrent refreshes.

## How I validated it

- `pnpm evals:pr specs/mcp-oauth-status-truthfulness.test.ts` — 1 passed, 0 failed, 0 skipped. Asserts the positive claim (verified success survives delayed/missing/differently keyed snapshots until convergence) and five negative halves: unverified servers never render connected, stale refreshes are rejected, terminal verdicts override the grace, foreign workspaces inherit nothing, and the grace expires when the engine never confirms.
- `bun test --isolate tests/mcp-oauth-status-synchronization.test.ts` — 11 pass, 0 fail (identity resolution, workspace scoping, disabled/removed entries, bounded grace).
- `pnpm --filter @openwork/app typecheck` — clean.
- `pnpm --filter @openwork/app test` — 1034 pass, 5 fail; the same 5 failures reproduce with the identical command on a clean `dev` control checkout at `cdc18ac52` (the only commit since is docs/licensing-only), so they are pre-existing and unrelated.
- `pnpm --filter @openwork/app build` — production bundle builds clean.
- Final validated commit: `b9107a14cd82eeb104ae3d75903ab7af5f3871c9` on `dev` at `ddf85156ff963602ec8095ccbaeebd10d5f0afa6`.

## Review focus

1. The bounded optimistic grace (default 6 refreshes) after a verified OAuth completion — confirm the tradeoff between briefly trusting a verified handshake and strict engine-only truth is right for the Extensions surface.
2. `refreshMcpServers` re-entry from `healUnhealthyMcpEntries` — check the token discipline removes the old late-write path without starving refreshes under repeated heals.
3. The conservative alias resolution in `resolveMcpStatusByIdentity` — verify returning `undefined` on ambiguous normalized collisions is the behavior you want rather than first-match.

## Risk and rollback

Bounded to the connections domain (store, auth modal, status resolution) plus one new leaf module. The riskiest surface is status display for servers whose engine snapshot keys differ from their configured names — previously those silently showed nothing; now they resolve through aliases, with ambiguity handled by showing nothing rather than something wrong. No persistence or migration is involved; reverting the single commit restores the previous behavior exactly.

### DIFF
--- a/apps/app/src/app/constants.ts
+++ b/apps/app/src/app/constants.ts
@@ -1,6 +1,7 @@
 import type { ModelRef, SuggestedPlugin } from "./types";
 import { t } from "../i18n";
 import { getDenMcpUrl } from "./lib/den";
+import { canonicalMcpServerName } from "./mcp";
 import {
   BUILT_IN_OPENWORK_EXTENSION_MANIFESTS,
   extensionContribution,
@@ -92,11 +93,7 @@ export function isBuiltInOpenWorkExtension(entry: Pick<McpDirectoryInfo, "kind" 
 /** Derive a safe MCP server name from a display name or explicit serverName. */
 export function getMcpServerName(entry: McpDirectoryInfo): string {
   if (entry.serverName) return entry.serverName;
-  return entry.name
-    .toLowerCase()
-    .replace(/[^a-z0-9_-]/g, "-")
-    .replace(/-+/g, "-")
-    .replace(/^-|-$/g, "") || "mcp";
+  return canonicalMcpServerName(entry.name);
 }
 
 export const MCP_QUICK_CONNECT: McpDirectoryInfo[] = [
--- a/apps/app/src/app/mcp.ts
+++ b/apps/app/src/app/mcp.ts
@@ -14,8 +14,16 @@ export function normalizeMcpSlug(name: string): string {
   return name.toLowerCase().replace(/[^a-z0-9]+/g, "-");
 }
 
+export function canonicalMcpServerName(name: string): string {
+  return name
+    .toLowerCase()
+    .replace(/[^a-z0-9_-]/g, "-")
+    .replace(/-+/g, "-")
+    .replace(/^-|-$/g, "") || "mcp";
+}
+
 export function getMcpIdentityKey(entry: McpIdentity): string {
-  return entry.id ?? entry.serverName ?? normalizeMcpSlug(entry.name);
+  return entry.id ?? entry.serverName ?? canonicalMcpServerName(entry.name);
 }
 
 export function validateMcpServerName(name: string): string {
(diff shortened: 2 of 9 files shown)


reachjalil:
## Test evidence — extension-connectivity-reflects-verified-engine-state-not-oauth-optimism — 0/0 screenshots passed · 1 assertion

### ℹ️ ASSERTION — 1. Extension connectivity is truthful

Verified OAuth success survives delayed and differently keyed engine snapshots until convergence; unverified, stale, terminal, foreign-workspace, and expired states can never render as connected.

- ✅ **PASS** Extension connectivity is truthful — Verified OAuth success survives delayed and differently keyed engine snapshots until convergence; unverified, stale, terminal, foreign-workspace, and expired states can never render as connected.

---
_Test run created 2026-08-26T23:46:57.186Z · Source: `evals/results/test-runs/2026-08-26T23-46-57-186Z-extension-connectivity-reflects-verified-engine-state-not-oauth-optimism/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4145 --test-run 2026-08-26T23-46-57-186Z-extension-connectivity-reflects-verified-engine-state-not-oauth-optimism`_

PR #4146, opened 2026-08-26
Keep Web composer drafts private and conflict-safe

reachjalil:
## What I noticed

Composer drafts persisted into one shared localStorage document with no identity boundary and no write discipline. On a shared browser that has real consequences: sign out, and the next signed-in account could hydrate the previous person's unfinished message; switch organizations, and drafts written in one org context stayed readable in the other. Separately, two tabs editing the same session could silently overwrite each other — the tab with the older snapshot won whenever it wrote last — and a corrupted or quota-limited storage could throw inside the composer path.

## What I changed

The persisted draft store is now a versioned (v2) document keyed by an explicit scope plus workspace and session:

- **Scope ownership is explicit.** The unsigned local scope is `local`; a cloud scope exists only for a *verified* principal and organization reported by the Den session provider (a new `verifiedIdentity` value that is set strictly when the fetched user matches the exact credential and org that initiated the refresh, and cleared on logout, token change, or org change). A credential that has not been re-verified resolves to no scope at all, so drafts become inaccessible the instant identity verification lapses — not after the next round trip.
- **Stale writes conflict instead of winning.** Every save/clear merges against a fresh storage read and uses the per-draft revision as a compare-and-swap guard. A tab holding an old revision gets `conflict` plus the newer snapshot back; a stale clear cannot delete or resurrect a newer draft. After a conflict the tab is resynchronized, so its next operation is intentional.
- **Storage events reconcile live tabs**, scoped to the draft document key.
- **Legacy storage fails closed.** v1 drafts carry no identity attribution, so they are removed rather than migrated into any scope — no guessing about who wrote them.
- **Untrusted storage degrades safely.** Malformed documents parse as empty and repair on the next valid save; storage that throws (privacy modes, quota) reports `unavailable` and keeps drafts in memory instead of crashing.
- **The in-memory composer store participates.** Each session claims a draft scope; when the scope key changes at an account/org boundary, a layout effect replaces the previous scope's text and revokes its attachment previews before the browser can paint them.

## Why this matters

Drafts are unsent, often sensitive text. This makes their ownership match the verified cloud identity and organization, keeps concurrent tabs from destroying newer work, and removes a whole class of cross-identity leakage on shared machines — while unsigned local use keeps working exactly as before.

## How I validated it

- `pnpm evals:pr specs/session-draft-scope-integrity.test.ts` — 1 passed, 0 failed, 0 skipped. Covers reload continuity within one identity; unreadability across account, organization, local, and unverified scopes; two-tab reconciliation via storage events; stale save and stale clear conflicts plus post-conflict recovery; fail-closed legacy v1 removal; malformed-storage repair and denied-storage degradation.
- `bun test --isolate tests/draft-store.test.ts tests/composer-state-store.test.ts` — 16 pass, 0 fail (scope resolution, cross-scope isolation, eviction order, hydration-scope claims for attachment state).
- `pnpm --filter @openwork/app typecheck` — clean.
- `pnpm --filter @openwork/app test` — 1034 pass, 5 fail; the identical 5 failures reproduce with the same command on a clean `dev` control checkout at `cdc18ac52` (only a docs/licensing commit separates it from this base), so they are pre-existing and unrelated.
- `pnpm --filter @openwork/app build` — production bundle builds clean.
- Final validated commit: `9c6113016dbdb20c0ddeb07a0495770530232acd` on `dev` at `ddf85156ff963602ec8095ccbaeebd10d5f0afa6`.

## Review focus

1. The `verifiedIdentity` contract in `den-auth-provider.tsx` — it must only be set when the fetched user corresponds to the exact token and organization that initiated the refresh, and must clear on logout, credential change, org change, and stale-refresh rejection. This is the single source the draft scope trusts.
2. The fail-closed v1 removal — existing users lose currently persisted drafts once (in-memory drafts in open tabs are unaffected). I chose deletion over migration because v1 data cannot be attributed to an identity; confirm you agree.
3. The `useLayoutEffect` hydration boundary in `session-surface.tsx` — verify the scope-claim logic cannot fight the queued-draft or revert flows in the composer store.

## Risk and rollback

The blast radius is draft persistence and hydration; message sending, transcripts, and attachments-in-flight are untouched. The riskiest change is scope resolution: if identity verification misbehaved, drafts would err on the side of being invisible (no scope) rather than leaking — the failure mode is losing a persisted draft, never exposing one. Reverting the single commit restores v1 behavior; the v2 storage key is separate, so a revert cannot mis-read v2 documents.

### DIFF
--- a/apps/app/src/react-app/domains/cloud/den-auth-provider.tsx
+++ b/apps/app/src/react-app/domains/cloud/den-auth-provider.tsx
@@ -117,6 +117,7 @@ export function shouldRetryDenAuthOnSignal(input: {
 export type DenAuthStore = {
   status: DenAuthStatus;
   user: DenUser | null;
+  verifiedIdentity: { principalId: string; organizationId: string } | null;
   error: string | null;
   isSignedIn: boolean;
   refresh: () => Promise<void>;
@@ -187,6 +188,11 @@ function pendingServerSwitchForDeepLink(input: {
 export function DenAuthProvider({ children }: DenAuthProviderProps) {
   const [status, setStatus] = useState<DenAuthStatus>("checking");
   const [user, setUser] = useState<DenUser | null>(null);
+  const [verifiedIdentity, setVerifiedIdentity] = useState<{
+    principalId: string;
+    organizationId: string;
+  } | null>(null);
+  const verifiedCredentialRef = useRef<{ token: string; organizationId: string } | null>(null);
   const [error, setError] = useState<string | null>(null);
   // Monotonic token so stale async refreshes can't clobber a newer result.
   const refreshTokenRef = useRef(0);
@@ -219,9 +225,21 @@ export function DenAuthProvider({ children }: DenAuthProviderProps) {
     const currentRun = ++refreshTokenRef.current;
     const settings = readDenSettings();
     const token = settings.authToken?.trim() ?? "";
+    const organizationId = settings.activeOrgId?.trim() ?? "";
+    const verifiedCredential = verifiedCredentialRef.current;
+
+    if (
+      !verifiedCredential
+      || verifiedCredential.token !== token
+      || verifiedCredential.organizationId !== organizationId
+    ) {
+      verifiedCredentialRef.current = null;
+      setVerifiedIdentity(null);
+    }
 
     if (!token) {
       setUser(null);
+      setVerifiedIdentity(null);
       setError(null);
       lastSignalRetryAtRef.current = null;
       updateStatus("signed_out");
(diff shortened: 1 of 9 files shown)


reachjalil:
## Test evidence — composer-drafts-stay-private-per-identity-and-safe-under-concurrent-tabs — 0/0 screenshots passed · 1 assertion

### ℹ️ ASSERTION — 1. Composer drafts are private and conflict-safe

Drafts survive reloads within one verified identity scope; other accounts, organizations, the local scope, and unverified credentials read nothing; concurrent tabs reconcile via storage events; stale writers and clears conflict instead of overwriting; legacy v1 data is dropped fail-closed; corrupt or denied storage degrades safely.

- ✅ **PASS** Composer drafts are private and conflict-safe — Drafts survive reloads within one verified identity scope; other accounts, organizations, the local scope, and unverified credentials read nothing; concurrent tabs reconcile via storage events; stale writers and clears conflict instead of overwriting; legacy v1 data is dropped fail-closed; corrupt or denied storage degrades safely.

---
_Test run created 2026-08-26T23:52:52.870Z · Source: `evals/results/test-runs/2026-08-26T23-52-52-869Z-composer-drafts-stay-private-per-identity-and-safe-under-concurrent-tabs/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4146 --test-run 2026-08-26T23-52-52-869Z-composer-drafts-stay-private-per-identity-and-safe-under-concurrent-tabs`_

PR #4147, opened 2026-08-26
Make per-turn instruction assembly latency-safe

reachjalil:
## What I noticed

Instruction assembly runs on every single agent turn, and it was doing its deduplication work twice. The extensions plugin first combined all system-prompt sections into a deduplicated, ordered list, and then `composeAgentInstructions` ran the exact same combination pass again over the already-combined output before extracting the bodies. That is redundant per-turn work on the hot path that shapes every prompt.

## What I changed

`composeAgentInstructions` now accepts section groups directly (single sections, arrays, or sparse `null`/`undefined` entries — the same shape `combineInstructionSections` takes) and composes in one pass. Duplicate ids are skipped **before** their bodies are read, empty bodies never claim an id so first-non-empty semantics are preserved, and ordering is exactly what the two-pass pipeline produced. The extensions plugin drops its redundant pre-combination and feeds sections straight into the composer. Callers that still pass an already-combined array get identical output, and `combineInstructionSections` remains available as a primitive for section-list manipulation.

I kept this deliberately small and localized — the current architecture's section primitives (`create`/`combine`/`delete`/`expand`) are unchanged; only composition gained a direct group-accepting form.

## Why this matters

Per-turn latency budget is spent before the model sees a token. Removing a full redundant iteration, set construction, and body materialization from every turn is a small but strictly free win, and skipping body reads for duplicate ids means future sections with computed bodies can never pay for text that will be discarded.

## How I validated it

- `pnpm evals:pr specs/instruction-assembly-equivalence.test.ts` — 1 passed, 0 failed, 0 skipped. Asserts byte-equivalence between the legacy combine-then-compose pipeline and the single-pass composer for grouped, flat, sparse, duplicate, and empty inputs, plus the negative halves: each retained body is read exactly once, and a duplicate id's body is never read at all.
- `bun --conditions=development test src/opencode-plugins/agent-instruction-compose.test.ts` — 3 pass, 0 fail.
- `bun --conditions=development test src/opencode-plugins/` (owning module) — 80 pass, 0 fail across 8 files.
- `pnpm --filter openwork-server typecheck` — clean.
- `pnpm --filter openwork-server build` — the production plugin bundle (which ships `openwork-extensions-preview`) builds clean.
- Final validated commit: `d8710a9dc8e64db2e8a6fae8915375f9e9ea7cb1` on `dev` at `ddf85156ff963602ec8095ccbaeebd10d5f0afa6`.

## Review focus

1. The first-non-empty semantics: in both old and new code an empty body does not claim its id. The equivalence spec pins this; please confirm it matches your intent for section overlap control.
2. Whether you want `combineInstructionSections` kept exported as a primitive (I kept it — `delete`/`expand` operate on combined lists) or folded away in a follow-up.

## Risk and rollback

The change is confined to one pure module and its single caller; there is no state, I/O, or ordering dependency beyond the function itself, and the equivalence spec locks the observable output. Reverting the single commit restores the two-pass pipeline with no data implications.

### DIFF
--- a/apps/server/src/opencode-plugins/agent-instruction-compose.test.ts
+++ b/apps/server/src/opencode-plugins/agent-instruction-compose.test.ts
@@ -31,10 +31,25 @@ describe("agent instruction compose primitives", () => {
     ]);
   });
 
-  test("composeAgentInstructions returns ordered bodies only", () => {
-    expect(composeAgentInstructions([
+  test("composeAgentInstructions combines section groups once and returns ordered bodies", () => {
+    let observedBodyReads = 0;
+    const observedSection = {
+      id: "observed",
+      get body() {
+        observedBodyReads += 1;
+        return "three";
+      },
+    };
+
+    expect(composeAgentInstructions(
       createInstructionSection("a", "one"),
-      createInstructionSection("b", "two"),
-    ])).toEqual(["one", "two"]);
+      [
+        createInstructionSection("a", "ignored duplicate"),
+        createInstructionSection("empty", "  "),
+        createInstructionSection("b", "two"),
+      ],
+      observedSection,
+    )).toEqual(["one", "two", "three"]);
+    expect(observedBodyReads).toBe(1);
   });
 });
--- a/apps/server/src/opencode-plugins/agent-instruction-compose.ts
+++ b/apps/server/src/opencode-plugins/agent-instruction-compose.ts
@@ -5,6 +5,7 @@
  * combine — merge sections in order, one id wins (first non-empty)
  * delete  — drop a section by id
  * expand  — replace a section body with derived text when present
+ * compose — combine sections and return their prompt bodies
  *
  * Sections are the unit of overlap control: routing/tools/skills/session each
  * own one id so transforms stop stacking contradictory brochure text.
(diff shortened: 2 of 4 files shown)


reachjalil:
## Test evidence — per-turn-instruction-assembly-dedupes-once-and-stays-behaviorally-equivalent — 0/0 screenshots passed · 1 assertion

### ℹ️ ASSERTION — 1. Instruction assembly is single-pass and equivalent

The composed system prompt matches the legacy combine-then-compose output for grouped, flat, sparse, duplicate, and empty inputs; each retained body is read once and duplicate ids are never read.

- ✅ **PASS** Instruction assembly is single-pass and equivalent — The composed system prompt matches the legacy combine-then-compose output for grouped, flat, sparse, duplicate, and empty inputs; each retained body is read once and duplicate ids are never read.

---
_Test run created 2026-08-26T23:55:43.105Z · Source: `evals/results/test-runs/2026-08-26T23-55-43-105Z-per-turn-instruction-assembly-dedupes-once-and-stays-behaviorally-equivalent/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4147 --test-run 2026-08-26T23-55-43-105Z-per-turn-instruction-assembly-dedupes-once-and-stays-behaviorally-equivalent`_

PR #4173, opened 2026-08-27
fix(app): restart a dead workspace event stream when the sync token generation changes

reachjalil:
## Problem

`shouldRetrySyncSubscribe` treated 401/403/404 subscribe failures as terminal: the reconnect loop stopped scheduling retries entirely. That is reasonable for a permanently invalid token, but the same statuses occur transiently while the local server restarts or the runtime generation rotates. Later reattachment updated `entry.openworkToken`, but a stream that had already terminated was never restarted — the watchdog only acts while an active connection controller exists. The task kept running on the server while Desktop received no further events until a full reload.

## Change

- Extracted the reconnect loop from `session-sync.ts` into a dependency-free, generation-aware lifecycle (`sync-stream-lifecycle.ts`). There is no terminal dead state anymore:
  - Auth-shaped failures (401/403/404) retry on a slower **bounded exponential backoff** (5s doubling to a 60s cap) instead of dying — and instead of tight-looping on the fast network-error schedule.
  - Reattachment with a rotated token advances the connection **generation**: a stream parked in auth backoff restarts immediately with fresh backoff and the new credential. An auth failure from an attempt that raced the rotation is treated as stale evidence and retried promptly.
  - A healthy stream is left alone on rotation; if the server cuts it, the ordinary retry path reconnects with the new generation.
- On every successful (re)connect, run statuses are **level-reconciled** (existing `reconcileSessionRunStatuses`) before the stream is trusted again, so a cached idle recorded while the stream was down cannot mask work that continued on the server.
- The stream phase is **exposed** to surfaces through `useWorkspaceSyncStreamStore` / `getWorkspaceSessionSyncStreamPhase`: `connecting`, `live`, `reconnecting`, `auth-blocked`, `stale`.

## Proof (local lane)

`evals/specs/session-sync-stream-generation-restart.test.ts` (`@openwork/testkit`, PR lane):

- **Resume after rotation**: a long task streams, the connection drops, stale-token reconnects park the stream in `auth-blocked`; rotating the token restarts the stream immediately (no timer wait, no remount), reconciliation runs before the first resumed event, and the same task's events resume and complete. Phase history asserts every state transition.
- **Negative proof**: a permanently invalid token over 10 simulated minutes yields a bounded number of attempts (≤15), every gap ≥5s and ≤60s, exponential (5s, 10s, 20s, 40s) up to the 60s cap — no infinite tight reconnect loop, and the stream still probes (never terminal).
- **Watchdog**: a silent live stream is exposed as `stale` and reconnected.

`apps/app/tests/session-sync-token-rotation.test.ts` (wiring through the real `session-sync` module): a stale token parks the real sync entry in `auth-blocked`; `ensureWorkspaceSessionSync` reattachment with a rotated token restarts the stream on the same entry (no remount), the fresh-token subscription reconciles a cached idle back to the server's busy truth, a resumed assistant message reaches the transcript cache, and `session.idle` completes the task. Negative half: exactly one live subscription, never with the stale token; an unchanged token does not churn a healthy stream.

```
pnpm --dir evals exec vitest run --project pr specs/session-sync-stream-generation-restart.test.ts
  → 3 passed (3), 0 failed, 0 skipped
cd apps/app && bun test --isolate tests/session-sync-token-rotation.test.ts \
  tests/session-sync-run-status.test.ts tests/session-sync-lifecycle.test.ts \
  tests/session-sync-permissions.test.ts tests/session-sync-tool-parts.test.ts
  → 51 pass, 0 fail
pnpm --dir apps/app typecheck → clean
```

Full `bun test --isolate tests/` shows 5 failures that reproduce identically on a clean `origin/dev` (4e9b6eac6) control run in the same environment (1047 pass/5 fail clean vs 1049 pass/5 fail with this change — the +2 are this PR's new tests): `connect-capability-inventory` ×2, `cloud workspace boot takeover`, `session-route cloud provider sync wiring`, `extensions-sidebar-header` — pre-existing, unrelated.

Verdict: **Passed** (local lane; Daytona not used for this renderer-unit-provable change).

### DIFF
--- a/apps/app/src/react-app/domains/session/sync/session-sync.ts
+++ b/apps/app/src/react-app/domains/session/sync/session-sync.ts
@@ -1,4 +1,5 @@
 import type { UIMessage } from "ai";
+import { create } from "zustand";
 import type { FilePart, Part, PermissionRequest, PermissionV2Request, QuestionRequest, Session, SessionStatus, Todo } from "@opencode-ai/sdk/v2/client";
 
 import { getReactQueryClient } from "../../../infra/query-client";
@@ -44,7 +45,9 @@ import {
   type DeltaFlushLane,
   type PendingDelta,
 } from "./session-transcript-deltas";
+import { startSyncStreamLifecycle, type SyncStreamPhase } from "./sync-stream-lifecycle";
 
+export { type SyncStreamPhase } from "./sync-stream-lifecycle";
 export {
   applyPendingDeltasToTranscript,
   coalescePendingDeltas,
@@ -68,6 +71,10 @@ type ListenerRegistry<Listener> = Map<Listener, number>;
 type SyncEntry = {
   input: SyncOptions;
   openworkToken: string;
+  // Reattachment can rotate the token after the stream already failed. This
+  // hook advances the stream lifecycle's connection generation so a stream
+  // parked in auth backoff restarts immediately with the new credential.
+  notifyStreamGenerationChanged: (() => void) | null;
   refs: number;
   dispose: () => void;
   disposeTimer: ReturnType<typeof setTimeout> | null;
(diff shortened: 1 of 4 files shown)


reachjalil:
## Test evidence — a-silent-live-stream-is-exposed-as-stale-and-reconnected-by-the-watchdog — 0/0 screenshots passed

⚪ **0/0 screenshots passed** · 0 expectations passed · 0 failed · 0 pending

---
_Test run created 2026-08-27T19:04:35.901Z · Source: `evals/results/test-runs/2026-08-27T19-04-35-901Z-a-silent-live-stream-is-exposed-as-stale-and-reconnected-by-the-watchdog/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4173 --test-run 2026-08-27T19-04-35-901Z-a-silent-live-stream-is-exposed-as-stale-and-reconnected-by-the-watchdog`_

PR #4199, opened 2026-08-28
fix(app): stop the Library page from refetching opencode-config on every settings re-render

benjaminshafii:
## Problem

`McpView` reloads `opencode.json` whenever its `readConfigFile` prop changes identity. `settings-route.tsx` passed an inline lambda, so every settings re-render — which store snapshots trigger every few seconds while idle (health-poll timestamps like `openworkServerCheckedAt`, MCP status refresh `mcpLastUpdatedAt`) — refired the config-load effect and issued a project+global read pair.

Measured live against an installed-production workspace (headless-prod-live world, CDP-instrumented `window.fetch`): **37 `GET /workspace/:id/opencode-config` requests in a 30s idle window** on the Library page (~1.2 req/s of pure refetch churn).

The unstable prop dates to #1470 (April); recent settings-route store churn made it fire more often.

## Fix

Memoize the prop on the stable connections store (`useCallback`), so the config effect only refires when the workspace actually changes. One line at the call site plus the memoized callback.

## Proof

New spec `evals/specs/library-config-read-budget.e2e.test.ts`:
- Positive half: the Library page still performs the initial opencode-config read (counted at both seams — openwork-server fetch and Electron `readOpencodeConfig` IPC).
- Negative half: during a 20s idle window on Library, reads stay within a budget of 2.

Discrimination check: with the fix stashed (unfixed tree), the same spec fails with **7 reads in 20s idle**; with the fix it passes.

Lane: Daytona (`OPENWORK_EVAL_DAYTONA=1`), evidence to be published on the PR head.

### DIFF
--- a/apps/app/src/react-app/shell/settings-route.tsx
+++ b/apps/app/src/react-app/shell/settings-route.tsx
@@ -759,6 +759,13 @@ function SettingsRouteContent(props: SettingsSurfaceProps = {}) {
   refreshMcpServersRef.current = connectionsStore.refreshMcpServers;
   notifyMcpReloadingRef.current = connectionsStore.notifyMcpReloading;
   pollMcpServersAfterReloadRef.current = connectionsStore.pollMcpServersAfterReload;
+  // Stable identity: McpView reloads opencode.json whenever this prop changes,
+  // so an inline lambda would refetch the config on every settings re-render
+  // (store snapshots tick every few seconds while idle on the Library page).
+  const readMcpConfigFile = useCallback(
+    (scope: "project" | "global") => connectionsStore.readMcpConfigFile(scope),
+    [connectionsStore],
+  );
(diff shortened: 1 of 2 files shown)


benjaminshafii:
## Test evidence — an-idle-library-page-keeps-opencode-config-reads-bounded — 0/0 screenshots passed

⚪ **0/0 screenshots passed** · 0 expectations passed · 0 failed · 0 pending

---
_Test run created 2026-08-28T14:58:32.314Z · Source: `evals/results/test-runs/2026-08-28T14-58-32-313Z-an-idle-library-page-keeps-opencode-config-reads-bounded/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4199 --test-run 2026-08-28T14-58-32-313Z-an-idle-library-page-keeps-opencode-config-reads-bounded`_

PR #4273, opened 2026-09-01
fix(app,server): stop Working timers from ticking unvalidated when the engine is unreachable

reachjalil:
## Problem

A network drop, laptop sleep, or dead engine leaves the transcript's live **"Working Xm Ys"** row ticking forever (users have watched it pass 38 minutes with nothing behind it). Four verified gaps compound:

- Busy status is latched on positive evidence only: `session.status`/`session.idle`/`session.error` edges or a **successful** `/session/status` reconcile. No layer downgrades confidence when evidence stops arriving.
- Failed reconciles are swallowed (`catch { return; }`) and rescheduled forever, invisibly.
- The engine event-stream keepalive is an SSE **comment** (`: ping`), which SSE parsers drop before liveness tracking — so the renderer's 30s stale-stream watchdog can't tell a healthy quiet stream from a dead socket and churns healthy connections through reconnects.
- Nothing revalidates immediately when connectivity returns; the first paint after sleep shows the whole gap as confident "worked" time.

## Change

Treat busy as a **validated lease**, reusing the machinery that already exists (the 250ms active-status reconcile loop is the continuous health check — it was just blind and unconsumed):

- **Failure ledger:** consecutive failed status revalidations (and the frozen last-confirmed time) are recorded per workspace stream in the existing sync stream store. Aborted fetches don't count. A failure never touches run state itself — absence of evidence never fabricates idle, so legitimately long runs are safe.
- **Honest transcript state:** past `reconcileFailureDegradedThreshold` (3 — ~1s for refused connections, ~3 request timeouts for blackholes), the ticking Working row is replaced by `Connection lost — reconnecting…` with the timer frozen (`data-loading-message="reconnecting"`), plus a "last update HH:MM" hint once the outage exceeds two minutes. The sub-agent Working line freezes the same way. Recovery settles only from authoritative status and resumes the true task age.
- **Visible heartbeat:** the merged engine event stream now emits `data: {"type":"server.heartbeat"}` (default 15s, env-tunable) instead of the `: ping` comment, so two beats fit inside the renderer's 30s stale window; quiet-but-healthy streams stop being aborted as stale, and real half-open sockets become detectable.
- **Online nudge:** on the window `online` event every workspace sync reconnects parked streams with fresh backoff and revalidates run statuses immediately, so post-sleep/offline state converges within one fetch instead of waiting out backoff.

## Proof

- `pnpm evals:e2e working-status-liveness --local` at this exact head — **passed (1/0/0)**: a real OpenCode engine runs a genuinely busy session (held-open provider stream), is SIGKILLed mid-run, the failure ledger crosses the degraded threshold in bounded time while run state stays busy (no fabricated idle), then an engine restarted on the same endpoint settles the run to idle and clears degradation within one revalidation. Runs cold-boot per invocation; also passed twice pre-rebase. Daytona credentials are unavailable in this environment and the spec family (like `session-completion-reconciliation`) declares `placement: "local"`, so local is the eligible lane.
- New spec registered under the `desktop.sessions-engine` contract; `vitest run specs/spec-impact.test.ts` — passed (6/6).
- `bun test --isolate` on the six owning app suites (message-list loading/reconnecting, session-sync run-status/lifecycle/token-rotation/permissions/tool-parts) — **71 pass, 0 fail** (three consecutive runs).
- `bun --conditions=development test src/engine-event-heartbeat.e2e.test.ts` — passed: a quiet proxied stream emits ≥2 data-bearing heartbeats and no comment frames (through the real server + engine pool with a mock engine).
- `apps/app` and `apps/server` `tsc --noEmit` — clean. `git diff --check` — clean.
- Full `apps/app` suite: 1121 pass, 5 fail — the identical five failures reproduce with the same command on a clean `dev` control worktree at the same base (all five pass in isolation in both trees), so they are pre-existing full-suite ordering flakes unrelated to this change (`connect-capability-inventory`, `cloud-workspace-overlay`, `session-provider-auth-server-sync`, `extensions-sidebar-header`).

## Notes for review

- Store updates are transition-only (healthy validations don't notify subscribers every 250ms; the failure counter caps at 99), so there is no render churn from the ledger.
- UI strings follow the existing hardcoded-English pattern of the Working row itself.
- The `evals` workspace `tsc -p .` lane is broadly red on clean `dev` (582 errors on a control worktree, same import-resolution classes on the sibling spec); the new spec matches sibling conventions and runs green under the Vitest transform.

### DIFF
--- a/apps/app/src/components/chat/message-list-provider.tsx
+++ b/apps/app/src/components/chat/message-list-provider.tsx
@@ -18,6 +18,11 @@ interface MessageListContextValue {
   displaySuggestions: boolean
   providerConnectedCount: number
   connectorIdentities: ConnectorToolIdentity[]
+  /**
+   * True while the workspace sync layer cannot validate live run status
+   * (failed status revalidation). Working indicators must stop ticking.
+   */
+  syncDegraded: boolean
   dispatchAction: (action: DispatchAction) => void
   setPrompt: (prompt: string) => void
   onRevertToUserMessage: (messageId: string) => void
@@ -58,6 +63,7 @@ interface MessageListProviderProps {
   displaySuggestions: boolean
   providerConnectedCount: number
   connectorIdentities?: ConnectorToolIdentity[]
+  syncDegraded?: boolean
   dispatchAction: (action: DispatchAction) => void
   setPrompt: (prompt: string) => void
 }
@@ -78,6 +84,7 @@ export function MessageListProvider({
   displaySuggestions,
   providerConnectedCount,
   connectorIdentities = [],
+  syncDegraded = false,
   dispatchAction,
   setPrompt,
   onRevertToUserMessage,
@@ -155,6 +162,7 @@ export function MessageListProvider({
       displaySuggestions,
       providerConnectedCount,
       connectorIdentities,
+      syncDegraded,
       ...stableHandlers,
       onOpenSubagentSession: canOpenSubagentSession
         ? stableHandlers.onOpenSubagentSession
@@ -172,6 +180,7 @@ export function MessageListProvider({
       displaySuggestions,
       providerConnectedCount,
       connectorIdentities,
+      syncDegraded,
       stableHandlers,
       canOpenSubagentSession,
       canResumeInterrupted,
--- a/apps/app/src/components/chat/message-list.tsx
+++ b/apps/app/src/components/chat/message-list.tsx
@@ -15,6 +15,7 @@ import {
   Pencil,
   Split,
   Undo2,
+  WifiOff,
 } from "lucide-react"
 import {
   DynamicToolUIPart,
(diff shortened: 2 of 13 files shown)


reachjalil:
## Test evidence — a-live-run-whose-engine-becomes-unreachable-network-drop-sleep-dead-process-is-c — 0/0 screenshots passed · 4 assertions

✅ **0/0 screenshots passed · 4 assertions** · 4 expectations passed · 0 failed · 0 pending

### ℹ️ ASSERTION — 1. continuousValidation: while a run is believed live, the sync layer keeps revalidating against /session/status and records consecutive failures once the engine is unreachable, crossing the degraded threshold within seconds — never: swallow failed revalidations silently while a Working indicator keeps ticking

After the engine was killed mid-run, the live reconcile loop kept attempting /session/status (3 attempts) and recorded 3 consecutive failures, crossing the degraded threshold 627ms after the kill.

- ✅ **PASS** continuousValidation: while a run is believed live, the sync layer keeps revalidating against /session/status and records consecutive failures once the engine is unreachable, crossing the degraded threshold within seconds — never: swallow failed revalidations silently while a Working indicator keeps ticking — After the engine was killed mid-run, the live reconcile loop kept attempting /session/status (3 attempts) and recorded 3 consecutive failures, crossing the degraded threshold 627ms after the kill.

### ℹ️ ASSERTION — 2. noFabricatedIdle: an unreachable engine leaves the busy run state and status cache untouched — never: invent idle or completion from elapsed time or from failed validation

While the engine was unreachable, the activity record stayed runActive and the status cache stayed busy: failed validation was recorded as degraded health, never converted into a fabricated idle.

- ✅ **PASS** noFabricatedIdle: an unreachable engine leaves the busy run state and status cache untouched — never: invent idle or completion from elapsed time or from failed validation — While the engine was unreachable, the activity record stayed runActive and the status cache stayed busy: failed validation was recorded as degraded health, never converted into a fabricated idle.

### ℹ️ ASSERTION — 3. degradedPresentation: the recorded failure count crosses the exported threshold that flips the transcript from the ticking Working row to its reconnecting presentation, and the frozen last-confirmed time predates the outage — never: present unvalidated elapsed time as confident progress

The recorded failure count (3) meets reconcileFailureDegradedThreshold (3) — the exact store input the session surface maps to the reconnecting row with its timer frozen — and the frozen last-confirmed time (1788287106797) predates the outage.

- ✅ **PASS** degradedPresentation: the recorded failure count crosses the exported threshold that flips the transcript from the ticking Working row to its reconnecting presentation, and the frozen last-confirmed time predates the outage — never: present unvalidated elapsed time as confident progress — The recorded failure count (3) meets reconcileFailureDegradedThreshold (3) — the exact store input the session surface maps to the reconnecting row with its timer frozen — and the frozen last-confirmed time (1788287106797) predates the outage.

### ℹ️ ASSERTION — 4. authoritativeRecovery: once an engine answers on the same endpoint again, one revalidation settles the run from authoritative status and clears the degradation without a reload or timer expiry — never: require user action or stream reconnect backoff to settle a reachable run

After an engine answered on the same endpoint again, the run settled from authoritative status within 560ms: runActive cleared, the status cache read idle, and consecutive failures reset to 0.

- ✅ **PASS** authoritativeRecovery: once an engine answers on the same endpoint again, one revalidation settles the run from authoritative status and clears the degradation without a reload or timer expiry — never: require user action or stream reconnect backoff to settle a reachable run — After an engine answered on the same endpoint again, the run settled from authoritative status within 560ms: runActive cleared, the status cache read idle, and consecutive failures reset to 0.

---
_Test run created 2026-09-01T18:25:02.491Z · Source: `evals/results/test-runs/2026-09-01T18-25-02-491Z-a-live-run-whose-engine-becomes-unreachable-network-drop-sleep-dead-process-is-c/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4273 --test-run 2026-09-01T18-25-02-491Z-a-live-run-whose-engine-becomes-unreachable-network-drop-sleep-dead-process-is-c`_

PR #4284, opened 2026-09-01
fix(hosts): never prune a live sibling desktop's profile when spawning another

benjaminshafii:
## Symptom
`pnpm world up acme-demo` launches two desktops from one process (alex, then jordan). Alex showed **"OpenCode is unavailable for this workspace"** while jordan worked. Alex's profile directory was a hollow skeleton — no `electron.log`, no engine `opencode.log`.

## Root cause
`clearStaleSurfaces(rootDir)` in `evals/packages/hosts/src/local.ts` ran before **every** non-caller-owned spawn and `rm -rf`'d every entry under the surfaces root — which is scoped by the spawning *process* pid, so it is shared by every desktop that process launches. The second spawn deleted the first desktop's live userdata, `openwork-dev-data` and engine dirs from under the running Electron; the engine could no longer start or write, and every proxied `/opencode/*` call failed. On Linux the accompanying `pkill --full <rootDir>` would additionally kill the live sibling outright (BSD `pkill` on macOS has no `--full`, so that step silently no-op'd here). The pruner's one-desktop-per-process assumption is violated by any two-desktop world or spec.

## Fix
- Module-level registry of live profile roots (`resolveHost()` builds a **new** local host per `desktop()`, so a per-host set cannot see siblings). Registered at spawn (Electron and Chrome), unregistered on `disposeSurface`, so the original disk-leak protection (the reason the pruner exists) still applies to dead profiles.
- `pruneStaleSurfaceProfiles(rootDir, { live, kill, log })` replaces `clearStaleSurfaces`: prunes only untracked entries; kills per stale path with `pkill -f <path>` (BSD + GNU), never the root.
- `handle.meta.profileRoot` carries the registered path so disposal can unregister precisely.

## Verification (PR head `ef82ee66d`, local lane — no Daytona credentials in this environment; the spec is app-less)

| Check | Command | Result |
|---|---|---|
| **New spec** — live sibling kept byte-identical, stale removed, kill invoked once with the stale path only; real child on the stale path dies while the live-path child survives; unregister → later prune removes it | `pnpm evals:pr specs/local-host-live-profile-prune.test.ts` | 1/1 |
| Hosts unit tests (3 new) | `node --test evals/packages/hosts/test/local-host.test.ts` | 10/10 |
| World engine neighbors | `pnpm evals:pr specs/shared-world-engine.test.ts specs/world-crash-reap.test.ts` | 4/4 |
| **Runtime proof of the original symptom** | `node evals/bin/world.mjs up acme-demo --detach` → both `testkit-admin-*` and `testkit-fresh-*` profiles keep `electron.log` and `…/opencode/log/*.log`; alex's log shows `GET …/opencode/session 200`, `POST …/opencode/session 200`, `GET …/opencode/session/status 200`; `world down` clean | manual transcript |

Not in this PR (separate, product-side): the 5s first-load engine deadline vs. 15s engine-start deadline race (`apps/server/src/server.ts:1570,1712` vs `managed-opencode.ts:188`) that produces two transient 502s on a cold start under load — that one self-heals and is not what broke alex.

### DIFF
--- a/evals/packages/hosts/src/local.ts
+++ b/evals/packages/hosts/src/local.ts
@@ -2,7 +2,7 @@ import { execFile, spawn } from "node:child_process";
 import { constants, existsSync, openSync } from "node:fs";
 import { access, mkdir, readdir, readFile, rm, stat, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
-import { delimiter, dirname, join } from "node:path";
+import { delimiter, dirname, join, resolve } from "node:path";
 import { allocateFreePort, allocateFreePorts, listTargets, waitForCdp } from "@openwork/cdp";
 import {
   desktopBootstrapPath,
@@ -75,6 +75,16 @@ interface SpawnDetachedOptions {
   cwd: string;
   env: NodeJS.ProcessEnv;
   logPath: string;
+}
+
+const liveProfileRoots = new Set<string>();
+
+export function registerLiveProfileRoot(profileRoot: string): void {
+  liveProfileRoots.add(resolve(profileRoot));
+}
+
+export function unregisterLiveProfileRoot(profileRoot: string): void {
+  liveProfileRoots.delete(resolve(profileRoot));
 }
 
 interface KillLocalPidOptions {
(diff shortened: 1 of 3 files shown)


benjaminshafii:
## Test evidence — pruning-stale-desktop-profiles-never-touches-a-live-sibling-desktop — 0/0 screenshots passed · 3 assertions

✅ **0/0 screenshots passed · 3 assertions** · 3 expectations passed · 0 failed · 0 pending

### ℹ️ ASSERTION — 1. Live sibling profile survives stale-profile pruning

The registered live profile stayed byte-identical, the stale profile was removed, and kill received only the stale absolute path—not the root or live path.

- ✅ **PASS** Live sibling profile survives stale-profile pruning — The registered live profile stayed byte-identical, the stale profile was removed, and kill received only the stale absolute path—not the root or live path.

### ℹ️ ASSERTION — 2. Default pruning kills only the stale profile process

A detached process whose argv named the stale profile exited within five seconds, while the registered live sibling process remained alive.

- ✅ **PASS** Default pruning kills only the stale profile process — A detached process whose argv named the stale profile exited within five seconds, while the registered live sibling process remained alive.

### ℹ️ ASSERTION — 3. Unregistered profiles become pruneable

After unregistering the formerly live profile, the next prune removed its directory and kept no live entries.

- ✅ **PASS** Unregistered profiles become pruneable — After unregistering the formerly live profile, the next prune removed its directory and kept no live entries.

---
_Test run created 2026-09-01T21:56:18.157Z · Source: `evals/results/test-runs/2026-09-01T21-56-18-157Z-pruning-stale-desktop-profiles-never-touches-a-live-sibling-desktop/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4284 --test-run 2026-09-01T21-56-18-157Z-pruning-stale-desktop-profiles-never-touches-a-live-sibling-desktop`_

PR #4307, opened 2026-09-02
perf(vercel): skip unaffected builds and serve /api/den redirects at the edge

reachjalil:
## Why

Two things made every push expensive on the Vercel side of this repo:

1. **Build fan-out.** Every push to any branch built all four Vercel projects (`ee/apps/den-web`, `ee/apps/landing`, `apps/app`, `ee/apps/diagnostics`) even when the commit touched none of them. None of the projects had an ignore step, and the repo's `.vercelignore` stripped `.git` from the build checkout, which would have made any git-aware ignore step impossible anyway.
2. **`/api/den/*` compatibility redirects ran as a function.** The route exists only so desktop builds that still call the web origin reach the Den API. Every one of those polls invoked the Next.js route handler just to emit a 307 to a target that is already known at build time.

## What changed

- **Graph-aware Ignored Build Step** in each project's `vercel.json`:
  `turbo query affected --base=$VERCEL_GIT_PREVIOUS_SHA --packages <own package> --exit-code || exit 1`
  - Exit 0 (skip) only when neither the package nor any dependency in the turbo graph changed since the last successful deployment of that project on the branch.
  - Everything else exits 1 and the build proceeds: affected, empty/unknown base ref (first deploy on a branch, or a base older than Vercel's `--depth=10` clone), git or turbo failure. Vercel treats other exit codes as a deployment *error*, which was observed on this branch before the guard was added.
  - `turbo-ignore` was tried first and is deprecated upstream (it also downloads itself and a turbo binary per deploy); `turbo query affected` uses the turbo already in Vercel's build image and is Vercel's documented replacement for graph-aware skipping.
  - Unlike Vercel's built-in "Skip deployment" toggle, the turbo graph also skips commits that only touch paths outside the pnpm workspace (`evals/`, `docs/`, `.github/`, `scripts/`), which is a meaningful share of `dev` commits. The toggle is complementary and can additionally be enabled in project settings.
- **`.vercelignore` keeps `.git`.** Only Git-integration builds were affected by that entry; the CLI already excludes `.git` from uploads by default.
- **Build-time `/api/den/:path*` redirect** in `ee/apps/den-web/next.config.js` via `next-config-den-api-redirects.cjs`. It resolves the same target as the route handler (`DEN_API_BASE`, else the `api.`-prefixed host of `DEN_BASE_URL` / `DEN_WEB_PUBLIC_ORIGIN`) and emits a temporary (307) Next.js redirect, which Vercel serves from its routing layer without invoking a function. Query strings are forwarded by default. When the origin is only known per request (prebuilt images without `DEN_*` at build time) no rule is emitted and the existing route handler keeps resolving it, unchanged.

Response bytes for callers are identical (307, same `Location`, same query); only the layer serving them changes.

## Verification

Lane: local (app-less PR lane; no Den or app needed). Daytona not attempted for these package-level checks.

| Check | Command | Result |
|---|---|---|
| Spec (exact head `7204dc404`) | `pnpm evals:pr specs/vercel-build-skips-and-den-api-redirect.test.ts` | 1 passed, 0 failed, 0 skipped |
| den-web redirect tests | `bun test --conditions development tests/den-api-redirect-rules.test.ts app/api/_lib/den-api-redirect.test.mjs` | 10 pass, 0 fail (build-time origin equals the route handler's `Location` for every supported env shape) |
| den-web typecheck | `pnpm --filter @openwork-ee/den-web typecheck` | exit 0 |
| den-web build | `DEN_BASE_URL=https://app.openworklabs.com DEN_API_BASE=https://api.openworklabs.com pnpm build` | compiled; `.next/routes-manifest.json` contains the single `/api/den/:path*` rule with `statusCode: 307` |
| Ignore step against `dev` history (local turbo 2.10.7) | `turbo query affected --base=<parent> --packages <pkg> --exit-code` | docs-only and evals-only commits: all four skip; landing-only commit: landing builds, others skip; `packages/ui` change: landing, den-web, app build, diagnostics skips; empty or unknown base: non-zero (build) |

On Vercel, for this branch:
- `729fde1b0` (touches every `vercel.json`): all four projects ran the ignore step with git available, reported the package affected, and built.
- `7204dc404` (evals-only): all four deployments were canceled by the Ignored Build Step in ~1.5 s each ("not affected").
- Preview deployments are behind SSO protection, so the routing-layer redirect could not be probed over HTTP before merge. After the production deploy, `curl -sI https://app.openworklabs.com/api/den/v1/me?include=org` should still return `307` with `location: https://api.openworklabs.com/v1/me?include=org`, and no longer carry `x-matched-path: /api/den/[...path]`.

Classification: deployment/build configuration plus a build-time routing rule with byte-identical responses; no user-visible product behavior changes. Evidence above is the exact-head proof for the claims in the spec.


reachjalil:
## Test evidence — vercel-builds-only-the-apps-a-commit-affects-and-hosted-api-den-callers-are-redi — 0/0 screenshots passed · 2 assertions

✅ **0/0 screenshots passed · 2 assertions** · 2 expectations passed · 0 failed · 0 pending

### ℹ️ ASSERTION — 1. skipUnaffectedBuilds: every Vercel-deployed app gates its build on turbo&#39;s affected graph for its own package — never: gate it on another package&#39;s changes, error the deployment when the query cannot run, or strip .git from the build checkout

4 Vercel projects run turbo query affected for their own workspace package before installing, failing open to a build

- ✅ **PASS** skipUnaffectedBuilds: every Vercel-deployed app gates its build on turbo&#39;s affected graph for its own package — never: gate it on another package&#39;s changes, error the deployment when the query cannot run, or strip .git from the build checkout — 4 Vercel projects run turbo query affected for their own workspace package before installing, failing open to a build

### ℹ️ ASSERTION — 2. routingLayerRedirect: Den Web emits the /api/den 307 as a build-time redirect once the API origin is configured — never: emit a redirect rule when the API origin is only known per request, or drop the per-request fallback route

next.config.js emits one temporary /api/den/:path* redirect only when DEN_API_BASE or DEN_BASE_URL is set; the route handler remains for per-request resolution

- ✅ **PASS** routingLayerRedirect: Den Web emits the /api/den 307 as a build-time redirect once the API origin is configured — never: emit a redirect rule when the API origin is only known per request, or drop the per-request fallback route — next.config.js emits one temporary /api/den/:path* redirect only when DEN_API_BASE or DEN_BASE_URL is set; the route handler remains for per-request resolution

---
_Test run created 2026-09-02T05:36:34.626Z · Source: `evals/results/test-runs/2026-09-02T05-36-34-626Z-vercel-builds-only-the-apps-a-commit-affects-and-hosted-api-den-callers-are-redi/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4307 --test-run 2026-09-02T05-36-34-626Z-vercel-builds-only-the-apps-a-commit-affects-and-hosted-api-den-callers-are-redi`_

PR #4313, opened 2026-09-02
fix(app): hide subagent sessions from the sidebar when their parent is archived or unloaded

yomgui:
## Problem

The sidebar fills up with `(@explore subagent)` rows. Sub-agent child sessions are meant to be reached only from the task card in their parent transcript (`sidebar/utils.ts` says so explicitly), but `getRootSessions` only hid a child when its parent was *also* in the loaded list:

```ts
return !parentID || !byID.has(parentID);
```

So a child was promoted to a root row whenever its parent was:
- **archived** — archived rows are partitioned out *before* the root check, so every child of an archived parent reappears;
- **outside the `limit: 200` page** — each subagent run is its own session, so they crowd parents out of the window;
- **deleted**.

The fallback is a leftover from the nested-tree sidebar (an orphan had nothing to nest under, so it became a root). #4088 removed nesting but kept the fallback and its test.

## Fix

`getRootSessions` now drops any session with a `parentID`, regardless of whether the parent is loaded. The `session.list` request is intentionally left unchanged (no `roots: true`): the "back to parent" control in a child session header looks the child up in that same list.

## Tests

- New `evals/specs/sidebar-hides-subagent-sessions.test.ts` (PR lane): archived-parent and out-of-window children produce no rows; real roots (`parentID` undefined or null) still render and a child of a loaded parent stays hidden. Red before the fix (both children leaked), green after.
- Updated `apps/app/tests/session-sidebar-utils.test.ts`: the old "keeps orphan child as root" case asserted the buggy behavior and now asserts the opposite, including the archived-parent case.

## Verification (local lane — Daytona not used; pure helper spec, no app/Den needed)

```
pnpm evals:pr specs/sidebar-hides-subagent-sessions.test.ts   # 2 passed, 0 skipped
cd apps/app && bun test tests/session-sidebar-utils.test.ts tests/sidebar-tree-nesting.test.ts   # 14 passed
pnpm --filter @openwork/app exec tsc --noEmit -p tsconfig.json   # exit 0
```

Verdict: **Passed**.

### DIFF
--- a/apps/app/src/react-app/domains/session/sidebar/utils.ts
+++ b/apps/app/src/react-app/domains/session/sidebar/utils.ts
@@ -45,13 +45,13 @@ const normalizeSessionParentID = (session: SessionListItem) => {
   return parentID || "";
 };
 
-export const getRootSessions = (sessions: WorkspaceSessionGroup["sessions"]) => {
-  const byID = new Set(sessions.map((session) => session.id));
-  return sessions.filter((session) => {
-    const parentID = normalizeSessionParentID(session);
-    return !parentID || !byID.has(parentID);
-  });
-};
+/**
+ * A session with a parentID is a sub-agent child, whether or not its parent is
+ * in the loaded page (the parent may be archived, deleted, or beyond the list
+ * limit). Children are only reached from the task card in their parent.
+ */
+export const getRootSessions = (sessions: WorkspaceSessionGroup["sessions"]) =>
+  sessions.filter((session) => !normalizeSessionParentID(session));
 
 /**
  * Return every descendant of a session in stable session-list order. The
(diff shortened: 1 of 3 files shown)


yomgui:
## Test evidence — children-of-a-loaded-active-parent-are-still-hidden-and-the-parent-still-renders — 0/0 screenshots passed · 1 assertion

### ℹ️ ASSERTION — 1. Root sessions keep rendering after the subagent filter change

Roots with no parentID (undefined or null) render in order, and a child of a loaded parent stays hidden as before.

- ✅ **PASS** Root sessions keep rendering after the subagent filter change — Roots with no parentID (undefined or null) render in order, and a child of a loaded parent stays hidden as before.

---
_Test run created 2026-09-02T08:35:23.286Z · Source: `evals/results/test-runs/2026-09-02T08-35-23-286Z-children-of-a-loaded-active-parent-are-still-hidden-and-the-parent-still-renders/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4313 --test-run 2026-09-02T08-35-23-286Z-children-of-a-loaded-active-parent-are-still-hidden-and-the-parent-still-renders`_

PR #4366, opened 2026-09-03
chore(evals): keep the config-provider disconnect spec comments reporter-neutral

yomgui:
Comment-only change: the spec added in #4357 described its scenario by naming the reporting organization. Field reports in a public repository should stay reporter-neutral, so the comments now describe the generic scenario (a custom OpenAI-compatible provider such as a LiteLLM proxy defined in the user-level opencode.json). No code, assertion, or behavior changes — the spec still passes: `pnpm --dir evals exec vitest run --config vitest.config.ts --project pr specs/config-provider-disconnect.test.ts` → 1 passed, 0 skipped.


yomgui:
Duplicate of #4363 (same comment-only neutral-wording change, opened independently); #4363 merged first.

PR #4372, opened 2026-09-03
fix(world): give isolated headless worlds their own OpenCode sessions database

yomgui:
## Why

`pnpm world up dev-headless` (state `isolated`) spawned its managed `opencode serve` with the person's `HOME` and no `OPENCODE_DB`, so it opened the **same** `~/.local/share/opencode/opencode.db` as the installed desktop app. Two long-lived engines on one SQLite file means every journal write in one is a `busy_timeout` wait in the other, and world sessions land in the person's history. On the reporting machine a headless engine from Sep 1 was still holding the desktop's 1.5 GB database open while the desktop engine was timing out. Only `installed-production` worlds isolated engine data (`installedProductionHeadlessEnv`).

## What

- `resolveHeadlessWorldRuntimePaths` exposes `opencodeDbPath`: `tmp/dev-headless-opencode.db` for the compatibility world, `tmp/worlds/runtime/<name>/opencode.db` for named worlds.
- New `isolatedHeadlessEngineEnv(runtimePaths, env)` sets **exactly** `OPENCODE_DB` for the isolated branch. `HOME`/`XDG_*` are deliberately untouched so the world keeps the person's provider `auth.json` and engine log; an explicit `OPENCODE_DB` still wins.
- The managed engine inherits the server's `process.env` (`managed-opencode.ts`), so the variable reaches `opencode serve` without further plumbing.
- AGENTS.md documents the location.

Behavior change: the isolated world no longer sees desktop sessions — which is what "isolated" means. `installed-production` is unchanged.

## Verification

Lane: **local** (pure path/env resolution; no Daytona prerequisite applies).

```
pnpm evals:pr specs/headless-world-engine-isolation.test.ts
  ✓ an isolated headless world gives its engine a private sessions database  (1 passed, 0 failed, 0 skipped) exit 0
pnpm --filter @openwork/world exec node --test test/headless-isolation.test.ts test/headless-production.test.ts   9 pass / 0 fail
pnpm --filter @openwork/world typecheck                                                                            exit 0
```

Verdict: **Passed**.

Related: #4371 (server log file). Follow-ups: engine-rollover fingerprint, engine-stall UX.

PR #4371, opened 2026-09-03
feat(server): persist openwork-server logs to a rotating file

yomgui:
## Why

The desktop runs openwork-server in-process, so its structured logger only reached Electron's stdout — invisible in packaged builds. Every `Engine rollover requested. {engine.rollover.reason}`, reload-watcher trigger, and MCP re-sync error was emitted and lost. Diagnosing today's engine churn (8 engine spawns in 25 min, each 1.1–1.6 GB) had to be reconstructed from sidecar run IDs and file mtimes; the actual rollover reasons are still unknown because nothing persisted them.

## What

- **apps/server** — `OPENWORK_SERVER_LOG_FILE` enables an append-only file sink (`server-log-file.ts`). It always receives structured JSON lines regardless of the stdout format, rotates once at 20 MB (`file` + `file.1`), disables itself on I/O errors, and redacts credential-looking attribute keys (`authorization|token|secret|password|api[-_]?key|cookie|credential`) before any line is written. Writes are synchronous appends on one descriptor; a stream opens its fd asynchronously and races the rotation rename.
- **apps/desktop** — defaults the sink to `<userData>/logs/openwork-server.log` before the embedded bundle loads (the server memoizes the sink from `process.env` on first logger creation) and exposes `logFilePath` in the server snapshot.
- **apps/app** — Settings › Debug shows `Server log: <path>` (en + ru).
- **packages/types** — `OpenworkServerInfo.logFilePath`.

No behavior change for the CLI unless the variable is set.

## Verification

Lane: **local** (no Daytona prerequisites are needed for this server-only spec; nothing app-driving).

```
pnpm evals:pr specs/server-log-file.test.ts
  ✓ openwork-server persists structured, credential-free logs when OPENWORK_SERVER_LOG_FILE is set  (1 passed, 0 failed, 0 skipped) exit 0
```
Boots the real `src/cli.ts` twice: with the variable → JSON lines including the listening line (`service.name=openwork-server`, `run.id`, `process.pid`) and a `GET /health 200` line with `status`/`durationMs`, explicit `--token`/`--host-token` values absent, stdout still pretty; without the variable → no file, no directory.

```
pnpm --filter openwork-server test src/server-log-file.test.ts src/server-logger.test.ts   8 pass / 0 fail
pnpm --filter openwork-server typecheck                                                  exit 0
pnpm --filter @openwork/desktop exec node --test electron/runtime.test.mjs               29 pass / 0 fail
pnpm --filter @openwork/app typecheck                                                    exit 0
[...]

PR #4373, opened 2026-09-03
fix(server): seed standby engines with provider auth and stop re-seeds from forcing rollovers

yomgui:
## Why

Engine churn, not token rotation. A spawned engine's environment carries no provider keys (`engineEnv` in `embedded.ts`); credentials reach it only through `PUT /auth` from `syncManagedProviderAuth`. The pool never seeded a standby, so after every rollover the next provider-sync pass (5-minute interval or any gesture such as New task) resolved a new `generation:` scope, re-delivered the same key, reported it as `authChanged`, and called `reloadEngine({ forceStandby: true })` — one fresh engine per pass, bounded only by the 30 s spawn floor. On the reporting machine: 8 engines in 25 minutes, two generations resident at ~2.7 GB, and the 10 s request timeouts that surfaced as "OpenCode unavailable — Request timed out". The loop only stopped there by accident when #4281 renamed the provider env var (`no_stored_credential`).

## What

- **engine-pool** — new optional `prepareStandby(config, { workspace, generationId, baseUrl, username, password })` hook, run after the standby is healthy and *before* the flip. A throwing seed logs `Engine standby preparation failed; flipping anyway.` and still flips; the primary is never stranded.
- **managed-provider-auth** — `syncManagedProviderAuth` accepts an explicit standby `target` that uses the **same** generation scope the promoted primary will resolve, so the post-flip sync finds its fingerprints applied. The result now reports `rotated` (value changed since the last delivery anywhere) separately from `delivered`.
- **cloud-provider-sync** — `authChanged = rotated || removed`. Re-seeding a replaced engine with an unchanged key is delivered but can no longer force a standby. Genuine rotation still reloads exactly once (SDK clients cache config+auth together).
- **server** — `envServiceForConfig()` exposes the running env store to the pool hook; every reload caller passes a distinct rollover reason (`cloud_provider_sync`, `managed_provider_reload`, `operation_route`, `workspace_bootstrap`) so the persisted log from #4371 names the trigger instead of a uniform `engine_reload`.

Not changed on purpose: `forceStandby` on the provider-sync path (provider config is global; an in-place dispose refreshes one workspace instance only) and the 15-minute drain inactivity window (long silent tool calls such as 4-minute test runs must not be aborted).

## Verification

Lane: **local** (server-only witnesses; no Daytona prerequisite applies).

```
pnpm evals:pr specs/engine-rollover-auth-reseed.test.ts
  ✓ an engine rollover seeds the standby and re-seeding never forces another rollover  (1 passed, 0 failed, 0 skipped) exit 0
```
Wraps three focused witnesses:
- `engine-pool.test.ts` › seeds the healthy standby before flipping and flips anyway when seeding fails — `prepareStandby` runs once with the standby's URL + basic-auth while the old engine is still primary; a throwing seed still rolls over.
- `managed-provider-auth.test.ts` › seeding a standby generation leaves the promoted primary unchanged on the next sync · re-seeding a replaced engine with the same key is delivered but not rotated — 2 PUTs total, `unchanged` after the flip; rotation → `rotated`.
- `cloud-provider-sync.e2e.test.ts` › re-seeding an unchanged credential to a replaced engine generation does not reload again — fake pool changes generation on every reload; passes 2 and 3 deliver but keep `reloads` at 1; a rotated key reloads once more. **Fails on the previous code**: `Expected: 1, Received: 2`.

```
pnpm --filter openwork-server test src/managed-provider-auth.test.ts src/engine-pool.test.ts src/cloud-provider-sync.e2e.test.ts src/server-logger.test.ts   53 pass / 0 fail
pnpm --filter openwork-server typecheck   exit 0
```

Verdict: **Passed**.

Related: #4371 (persisted server log), #4372 (headless engine isolation). Follow-up: engine-stall UX.


yomgui:
## Test evidence — an-engine-rollover-seeds-the-standby-and-re-seeding-never-forces-another-rollove — ✅ passed

✅ **passed** · 3 expectations passed · 0 failed · 0 pending

### ℹ️ ASSERTION — 1. The pool seeds a healthy standby before it becomes primary

prepareStandby runs once per rollover with the standby&#39;s own URL and basic-auth while the old engine is still primary; a throwing seed still flips.

- ✅ **PASS** The pool seeds a healthy standby before it becomes primary — prepareStandby runs once per rollover with the standby&#39;s own URL and basic-auth while the old engine is still primary; a throwing seed still flips.

### ℹ️ ASSERTION — 2. Seeding uses the same generation scope the promoted primary will use

After the flip the primary-scoped sync reports the provider unchanged with no further PUT /auth; the total stays at two deliveries.

- ✅ **PASS** Seeding uses the same generation scope the promoted primary will use — After the flip the primary-scoped sync reports the provider unchanged with no further PUT /auth; the total stays at two deliveries.

### ℹ️ ASSERTION — 3. Re-delivering an unchanged key to a replaced engine does not reload

With a fake pool whose generation changes on every reload, the second and third sync passes deliver but keep reloads at 1; a rotated key reloads exactly once more. The same test fails on the previous authChanged definition (Expected 1, Received 2).

- ✅ **PASS** Re-delivering an unchanged key to a replaced engine does not reload — With a fake pool whose generation changes on every reload, the second and third sync passes deliver but keep reloads at 1; a rotated key reloads exactly once more. The same test fails on the previous authChanged definition (Expected 1, Received 2).

---
_Test run created 2026-09-03T23:47:31.667Z · Source: `evals/results/test-runs/2026-09-03T23-47-31-667Z-an-engine-rollover-seeds-the-standby-and-re-seeding-never-forces-another-rollove/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4373 --test-run 2026-09-03T23-47-31-667Z-an-engine-rollover-seeds-the-standby-and-re-seeding-never-forces-another-rollove`_

yomgui:
**Red `openwork-tests` is pre-existing on `dev`, not this PR.**

Both runners fail the same two governance specs that reference a file this PR does not touch:

- `spec-impact.test.ts` › every E2E spec is mapped… → `evals/specs/cross-server-handoff-atomic-commit.e2e.test.ts` has no contract
- `spec-quarantine.test.ts` › the enabled E2E set equals the evidence-backed inventory → 38 vs 37

Control on a clean `origin/dev` checkout (3e425fa53, which merged #4054 after this branch was cut):

```
git checkout --detach origin/dev && pnpm evals:pr specs/spec-impact.test.ts specs/spec-quarantine.test.ts
  × the enabled E2E set equals the evidence-backed inventory  → expected [ …(38) ] to deeply equal [ …(37) ]
  × every E2E spec is mapped to an implementation contract or explicitly allowlisted
      → E2E specs missing a contract or unmapped reason: evals/specs/cross-server-handoff-atomic-commit.e2e.test.ts
```

Fix for `dev`: #4375 registers the spec (2 files, +3 lines). Once it lands, this PR's check goes green on re-run; the PR's own witnesses are unaffected (see the test-evidence comment above).

yomgui:
Updated onto current dev (32963279a) without force-push. The stale governance failures are fixed on dev by #4375. Final-head proof: 53/53 focused server tests, server typecheck exit 0, and 1/1 testkit rollover-auth witness; evidence republished above.

PR #4108, opened 2026-08-26
Preserve composer focus during same-task refreshes

reachjalil:
## Summary

- keep an already-rendered current task interactive while its snapshot refreshes in the background
- preserve the existing switching state for initial loads and actual task changes
- add transition-level regression coverage plus a desktop scenario that continues typing without refocusing

## Root cause

Every snapshot refetch was classified as a task switch, including completion updates for the task already on screen. That temporarily made the composer non-editable, blurred the editor, and then restored editability without restoring focus.

## Verification

- `bun test --isolate tests/session-transition-controller.test.ts` — 3 passed
- app TypeScript check — passed
- exact-head desktop E2E at `19fecb38428ce5f3383bf076e9f21d291b779070` — 1 passed, 0 failed, 0 skipped; 2 assertion claims passed

PR #4392, opened 2026-09-04
fix(composer): keep the prompt box editable when the session snapshot fails

benjaminshafii:
## Problem

Reported in #bug: "Under certain circumstances, the OpenWork prompt box does not allow me to type into it." Screenshot showed an empty composer with placeholder, a wide selection-style caret, "Jump to latest", idle session, no error.

Root cause: `session-surface.tsx` passed `disabled={model.transitionState !== "idle" || sessionModelUnavailable}` to the Lexical editor. `transitionState` becomes `"failed"` on *any* `snapshotQuery.isError` (or local `error` state — including a failed clipboard copy). Two things made it invisible:

- Lexical 0.35 renders a non-editable editor identically to an editable empty one (placeholder still shows, no styling), and clicking a `contenteditable="false"` `<p><br></p>` draws exactly the wide highlight in the screenshot.
- The "Failed to load session" UI only renders when there is no cached transcript, so a failed *background* refetch (sleep/wake, engine restart, `refetchOnWindowFocus`) locked the composer with nothing on screen.

The gate was inherited wholesale from the Solid→React port (#1470) with no rationale; #4108 already removed it for the `isFetching` branch for exactly this reason.

## Change

- `editor.tsx` / `composer.tsx`: the editor is always editable. `disabled` is renamed `submitDisabled` on the editor and only gates Enter-submit. The send button and `handleSend` guards are unchanged.
- `session-surface.tsx`: dev-only `eval.session_snapshot.fail` control action (same pattern as the other `eval.*` hooks) that forces the snapshot query into error state and returns `{ ok, isError }`. `retry` now reads the eval flag at fetch time; default behaviour (3 retries) is preserved.
- `new-task-composer.tsx` needs no change: its `disabled` now only blocks send, as intended.

## Verification

**Unit (bun)** — `apps/app/tests/composer-editable-on-snapshot-error.test.tsx` renders the real `SessionSurface`, fails the snapshot refetch, and asserts the editor stays `contenteditable="true"`, typed text lands, the send button stays disabled, `onSendDraft` is never called, and the cached transcript is preserved; also covers the model-unavailable case.

```
pnpm --filter @openwork/app exec bun test --isolate tests/composer-editable-on-snapshot-error.test.tsx tests/composer-focus-continuity.test.tsx tests/session-transition-controller.test.ts
# 5 pass, 0 fail, 12 expect() calls
pnpm --filter @openwork/app typecheck   # exit 0
```
Confirmed red without the fix: with `editor.tsx`/`composer.tsx` stashed the new test fails (0 pass, 1 fail).

**E2E (testkit)** — `evals/specs/composer-editable-during-snapshot-error.e2e.test.ts` with world `snapshotFailure` in `evals/worlds/chat.ts` (seeds a transcript, executes `eval.session_snapshot.fail`, throws unless `isError: true`). Claims: composer editable + typed text lands; negative: text is not in the transcript (no send) and the seeded transcript literal is still visible. Channel ratchet: 0 raw escapes; `pnpm --dir evals typecheck:spec-layer` exit 0.

```
node evals/bin/evals.mjs composer-editable-during-snapshot-error --daytona
# verdict: passed (1/1, 0 skipped)
```

**Verdict: Passed** on Daytona against the PR head (see the test-evidence comment). The earlier red runs were the engine-bootstrap hang diagnosed in #4423; #4397 pre-seeds the engine cache in eval sandboxes, which is why this now runs.

Pre-existing unrelated bun failures on `origin/dev` (not touched here): `connect-capability-inventory`, `app-inspector`, `cloud-workspace-overlay`, `session-provider-auth-server-sync`.

## Follow-ups (out of scope)
- A failed clipboard copy or stop request sets `error`, which still disables *send* until dismissed; the `SessionErrorCard` renders at the top of the transcript where it can be off-screen.
- Consider visible disabled styling on the composer shell so a non-editable state can never be silent again.

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/composer/composer.tsx
+++ b/apps/app/src/react-app/domains/session/surface/composer/composer.tsx
@@ -1295,7 +1295,7 @@ export const ReactSessionComposer = memo(function ReactSessionComposer(props: Co
                 kind: isImageAttachment(attachment) ? "image" : "file",
                 previewUrl: attachment.previewUrl,
               }))}
-              disabled={props.disabled}
+              submitDisabled={props.disabled}
               placeholder={t("composer.placeholder")}
               onChange={props.onDraftChange}
               onSubmit={handleEditorSubmit}
--- a/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
+++ b/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
@@ -53,7 +53,7 @@ type EditorProps = {
   mentions: Record<string, ComposerMentionKind>;
   pastedText?: PastedTextToken[];
   attachments?: ComposerAttachmentToken[];
-  disabled: boolean;
+  submitDisabled: boolean;
   placeholder: string;
   onChange: (value: string) => void;
   onSubmit: (options: { queue: boolean }) => void | Promise<void>;
@@ -818,14 +818,9 @@ function SyncPlugin(props: {
   mentions: Record<string, ComposerMentionKind>;
   pastedText?: PastedTextToken[];
   attachments?: ComposerAttachmentToken[];
-  disabled: boolean;
 }) {
   const [editor] = useLexicalComposerContext();
   const valueRef = useRef(props.value);
-
-  useEffect(() => {
-    editor.setEditable(!props.disabled);
-  }, [editor, props.disabled]);
 
   useEffect(() => {
     // When the external value is cleared (e.g. after sending a message),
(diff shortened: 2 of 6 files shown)


benjaminshafii:
## Test evidence — the-composer-stays-editable-while-the-session-snapshot-query-is-in-error — ✅ passed

✅ **passed** · 0 expectations passed · 0 failed · 0 pending

**[world]** desktop(daytona) · workspace(/tmp/openwork-composer-snapshot-failure-1788542765796) · session("Composer snapshot failure proof") · [seed:raw] evalIn(<expression>) ×2
**[user]** see(composer, editable, timeoutMs=30000) · type(composer, "Draft survives the broken snapshot without being sent — composer proof 9472.") · see(composer, editable, text="Draft survives the broken snapshot without being sent — composer proof 9472.", timeoutMs=30000)
**[agent]** run(session.read_transcript)
**[user]** see(text=/chat-transcript-proof/, timeoutMs=30000)
**steps** 1 ✅ the composer remains editable after the snapshot fails (0.2s) · 2 ✅ a distinctive draft can be typed and remains editable (0.9s) · 3 ✅ typing does not send and the cached transcript remains visible (0.7s)
**verdict** passed · 3 user observations (see ×3) · 0 probes · steps 3/3

---
_Test run created 2026-09-04T17:24:24.401Z · Source: `evals/results/test-runs/2026-09-04T17-24-24-400Z-the-composer-stays-editable-while-the-session-snapshot-query-is-in-error/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4392 --test-run 2026-09-04T17-24-24-400Z-the-composer-stays-editable-while-the-session-snapshot-query-is-in-error`_

PR #4390, opened 2026-09-04
fix(den-web): let a dashboard hold the same MCP App more than once with different launch input

yomgui:
## Problem

An organization dashboard could not hold two tiles of the same MCP App with different launch input — for example two `searchJiraIssuesUsingJql` tiles with different queries. After the first add, the **Add app** picker row turned into a terminal **✓ Added** badge: no Add button, no launch-input field, and any second add of the same app was silently dropped.

Root cause: the Den Web picker identified a tile by `serverName:toolName` only (`org-dashboard-detail-screen.tsx`), ignoring `launchArguments`. Nothing else in the stack has that limitation — the API accepts duplicates, elements are a plain JSON array, and Desktop's `grantedEntryId` already includes launch arguments (pinned by `granted-dashboard-store.test.ts`). The authoring UI simply disagreed with the rest of the feature.

Internal ref **#6** (customer report; identity kept in the private tracker).

## Change

- **Tile identity = App + canonical launch arguments** (`dashboardElementKey` in `dashboard-mcp-app-catalog.ts`), matching Desktop. Same tool + different input → two tiles; same tool + same input → one.
- **Picker row is no longer a dead-end**: after the first add it keeps the Add control (labelled *Add another*), the auto-run toggle, and the launch-input field, and shows an *Added* / *Added ×N* count hint.
- **Identical tile is refused visibly** ("This app is already on the dashboard with the same launch input.") instead of silently dropped.
- **Element rows show their launch input**, so same-App tiles are distinguishable.

No API, schema, or Desktop change. No migration: no saved dashboard can contain duplicates today because the UI prevented them; element identity is content-based on both sides and unchanged for existing tiles.

## Proof

Lane: **local** (MySQL local placement). Daytona is available but cannot reach this spec: on Daytona placements `server({ env })` is not forwarded to the shared Den sandbox, so `DEN_DASHBOARDS_ENABLED` never reaches den-api and Den Web redirects the dashboards route before the picker exists (pre-existing; `evals/packages/env/src/den.ts` `provisionDenSandbox` call, unchanged here). The spec is therefore registered in `daytona-e2e-regression-profile.json` under the existing `per-test-den-env` category rather than producing a spurious CI red. Forwarding per-test Den env to the Daytona sandbox is a separate harness change.

[...]

PR #4377, opened 2026-09-03
fix(den-api): accept RFC 3339 offsets for Calendar and expose query constraints in search_capabilities

benjaminshafii:
## Why

Two recurring agent failures against the Google Workspace native capabilities, pulled from real OpenWork session history:

1. **`Invalid ISO datetime` on Calendar** — hit in 7+ sessions since Jul 29 (`getCapabilitiesGoogleWorkspaceCalendarEvents` `timeMin`/`timeMax`, `postCapabilitiesGoogleWorkspaceCalendarEvents` `start`/`end`). The model sends a valid RFC 3339 timestamp with an offset (`2026-09-03T17:00:00+02:00`); the route used `z.string().datetime()`, whose default `offset: false` only accepts a trailing `Z`. Google Calendar accepts offsets, so we were stricter than upstream, and the description gave no hint.
2. **Gmail `maxResults` cap invisible to the model** — the route enforces `.max(25)` and even documents "capped at 25", but `search_capabilities` returned query hints as a *live Zod object* (`inputSchema: operation.inputSchema`) that `JSON.stringify` dumps as `{"def":{"type":"number","checks":[]}}` — constraints stripped, so the model asks for 50 and gets `too_big`.

## What

- `routes/org/google-workspace.ts`: `z.string().datetime({ offset: true })` on the four Calendar fields; descriptions now state "RFC 3339 date-time with a UTC offset or Z" with an example.
- `mcp/catalog.ts`: new `getQueryParameterSchema(operation)` — plain JSON Schema `{ type: "object", properties, required? }` built from the OpenAPI query parameters, which already carry `minimum`/`maximum`/`default`/`format`/`required`/`description`.
- `mcp/search.ts`, `mcp/native-capabilities.ts`: emit `querySchema` next to `bodySchema`; drop the leaked `inputSchema`.
- `mcp/agent.ts`: declare `querySchema` in the output schema; tool description lists it.

Out of scope (follow-up): `microsoft-365.ts` has the same `.datetime()` on `timeMin`/`timeMax`/`start`/`end`, but Graph's event-create `dateTimeTimeZone.dateTime` is offset-less, so widening it needs a conversion step rather than a validator change.

## Proof

Spec: `evals/specs/google-workspace-connector-retrieval.test.ts` (extended; wraps the den-api bun suite and pins test names). New den-api tests, each with its negative half:

[...]

PR #4399, opened 2026-09-04
ci(evals): fall back to the full E2E set when a PR only touches excluded specs

benjaminshafii:
## Why

Tracked in #3915. The three most recent `Daytona E2E Regression Suite` failures were `verdict` failures with an empty matrix, not product regressions:

- Runs [33841753494](https://github.com/different-ai/openwork/actions/runs/33841753494) and [33844367127](https://github.com/different-ai/openwork/actions/runs/33844367127) (PR #4390): the PR's only e2e change was `dashboard-duplicate-capability-tiles.e2e.test.ts`, which the same PR added to `daytona-e2e-regression-profile.json`'s excluded list. The plan step built `matched_tests` from `git diff` *before* removing excluded specs, so `matched` was non-empty, the "fall back to the full eligible set" branch from #4378 never ran, the filter yielded `[]`, the e2e job was skipped, and `verdict` failed.
- Run [33816055555](https://github.com/different-ai/openwork/actions/runs/33816055555) (PR #4377) ran under the pre-#4378 spec-impact workflow, which had no fallback at all; already superseded.

## What

`.github/workflows/daytona-e2e-regression-suite.yml`, `Build E2E test matrix` step only:

- Intersect the changed e2e specs with `eligible_tests` first, so the existing `($matched | length == 0)` fallback covers "changed specs are all excluded" the same way it covers "no spec changed".
- Replace the now-unreachable `Warden suggestion:` summary line with a truthful one printed when the fallback happens.

`evals/specs/daytona-e2e-regression-schedule.test.ts`:

- The second test used to assert the *exact shell text* of the `matched_tests=` line, so it went red on this refactor (and on any future one) while never catching the actual bug. It now extracts the `Build E2E test matrix` `run:` body from the workflow at test time and executes it against a throwaway git fixture repo, asserting on `GITHUB_OUTPUT` (`batches`, `has_tests`) and `GITHUB_STEP_SUMMARY` across four scenarios: only an excluded spec changed, one eligible spec changed, no e2e spec changed, and a scheduled run with `ONLY_FILTER`.

No changes to the `e2e` or `verdict` jobs, or to the profile.

## Verification

Rebased on `dev` (picks up #4404, which fixed the unrelated `spec-boundary-ratchet` failure on the previous CI run).

```
pnpm evals:pr specs/daytona-e2e-regression-schedule.test.ts
```
Placement: local, `pr` lane. Verdict: **Passed** — `Tests 2 passed (2)`. Every claim has an observable assertion (matrix contents, `has_tests`, summary line presence/absence).

Mutation check — same spec with `dev`'s workflow copied over this branch's:
`× PR-triggered selection falls back to the full eligible set when no eligible spec changed` → `expected [] to deeply equal [ 'alpha.e2e.test.ts', 'beta.e2e.test.ts' ]`. The spec reproduces the empty-matrix regression from PR #4390 and goes green only with this fix.

`node evals/scripts/spec-boundary-ratchet.mjs` → `170 specs checked` (spec is grandfathered in the baseline; baseline untouched). `tsc --noEmit -p evals` clean for the spec.

Note: because this PR touches `.github/`, the regression suite itself withholds on it ("PR changes trusted review machinery") — expected.

## Follow-up (not in this PR)

Run [33805238155](https://github.com/different-ai/openwork/actions/runs/33805238155) failed in `plan` with "Warden summary or pull request head does not match": a push landed while Warden was running. That is a benign race but exits 1 and posts an alert; it could withhold with `authorized=false` instead.

### DIFF
--- a/.github/workflows/daytona-e2e-regression-suite.yml
+++ b/.github/workflows/daytona-e2e-regression-suite.yml
@@ -172,7 +172,8 @@ jobs:
           )"
 
           if [ "$EVENT_NAME" = "workflow_run" ]; then
-            matched_tests="$(git diff --name-only --diff-filter=ACMR "$BASE_SHA...$HEAD_SHA" | { grep -E '^evals/specs/[^/]+\.e2e\.test\.ts$' || true; } | sed 's#^evals/specs/##' | jq -Rsc 'split("\n") | map(select(length > 0))')"
+            changed_tests="$(git diff --name-only --diff-filter=ACMR "$BASE_SHA...$HEAD_SHA" | { grep -E '^evals/specs/[^/]+\.e2e\.test\.ts$' || true; } | sed 's#^evals/specs/##' | jq -Rsc 'split("\n") | map(select(length > 0))')"
+            matched_tests="$(echo "$eligible_tests" | jq -c --argjson changed "$changed_tests" 'map(select(. as $test | ($changed | index($test)) != null))')"
             only_filter=""
           else
             matched_tests='[]'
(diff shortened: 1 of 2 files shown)


benjaminshafii:
Superseded by #4403, which deleted `daytona-e2e-regression-suite.yml` and the exclusion profile this PR patched. In the new `daytona-e2e.yml`, an empty PR matrix exits 0 with a Warden suggestion, so the empty-matrix verdict failure this fixed can no longer occur.

PR #4393, opened 2026-09-04
feat(app): make Cmd+K an intelligent search over actions, settings, and sessions

benjaminshafii:
## Summary

Cmd/Ctrl+K becomes a real search bar instead of a fixed list.

- **Fuzzy ranking** with `fuzzysort` (already a dependency) over title › aliases › breadcrumb › detail, a token-AND gate ("settings folders" → Permissions), and a recents boost. Results are grouped **Recent / Actions / Settings / Sessions**; the top hit's group is always first.
- **Every settings panel is searchable** with plain-language aliases: "authorized folders" → Permissions, "dark mode" → Appearance, "api key" → AI Providers, "logs" → Debug, "sign in" → Account, plus Library sections (Skills, MCP servers, Connections, Plugins, Agents, Commands) with a "Settings › Library" breadcrumb. Gated by developer mode, platform capabilities, and the memory flag like the settings sidebar.
- **Sessions inline** in the root view — type a session title and jump.
- **New core actions:** Toggle sidebar, Automations, Dashboard, New workspace…, Sign in to OpenWork Cloud.
- **Recents** persisted in `localStorage` (cap 8) and shown on empty query; `>` prefix restricts to actions.
- Drops the unused `cmdk` dependency.

### Why this shape
Surveyed opencode (custom registry, substring match), cal.com (`kbar`, Radix) and t3code (React 19 + Base UI + fuzzy logic module + settings search index). t3code's is the same stack as ours, so the palette keeps our existing Base UI `Command` primitives and adds a small pure ranking module rather than importing `kbar`/`cmdk` (both bring Radix and a second overlay).

## Verification

Unit (bun): `pnpm --filter @openwork/app test` — new `command-palette-{search,recents,settings}.test.ts`.
Typecheck: `pnpm --filter @openwork/app typecheck` ✓.

E2E: `evals/specs/command-palette-search.e2e.test.ts` — creates a session, opens the palette, asserts the session ranks first in the Sessions group, "folders" ranks `settings:permissions` first (and not Preferences, no Sessions group), clicking navigates to `/settings/permissions` on the same workspace, reopen shows exactly one Recent (Permissions, not Appearance) and the localStorage entry, "dark mode" → Appearance first, `>` hides Settings, Escape closes without navigating.

Run: `OPENWORK_EVAL_REF=<head> pnpm evals:e2e command-palette-search --daytona` — verdict posted below.

PR #4414, opened 2026-09-04
feat(app): pin and rename the open session from Cmd+K

benjaminshafii:
## Summary

Follow-up to #4393: from **⌘K / Ctrl+K** you can now act on the session you're looking at.

- **Pin session / Unpin session** (`session.pin.toggle`) — label follows the pin state; aliases: pin, unpin, favorite, star, keep on top, sidebar.
- **Rename session…** (`session.rename`) — opens the existing rename dialog (title prefilled) via a tiny window-event bus (`session-actions-bus.ts`), same pattern as the palette bus. Aliases: rename, title, name, edit title.

Both rows show the current session title as detail and only appear when a session is open (rename also needs the engine client, like the sidebar menu).

## Verification
- Unit: `bun test tests/command-palette-search.test.ts` — `pin` / `rename` rank these actions first (7/7). Typecheck ✓. `lint:layers` + spec ratchets clean.
- **E2E — Passed** (local cold boot, `evals/specs/command-palette-pin-rename.e2e.test.ts`, new `paletteSessionActions` world with a seeded session):
  `OPENWORK_EVAL_E2E_TESTS=1 pnpm --dir evals exec vitest run --config vitest.config.ts --project e2e specs/command-palette-pin-rename.e2e.test.ts`
  Claims: no "Pinned" group before; `pin` shows **Pin session** and not Unpin; Enter → sidebar "Pinned" group appears; reopening shows **Unpin session** and not Pin; `rename` + Enter opens the "Rename session" dialog with the "Session name" field; typing a new name + Save → sidebar shows the new title and the old title is gone.

Screenshots from the run are attached below.

### DIFF
--- a/apps/app/src/react-app/domains/session/chat/session-page.tsx
+++ b/apps/app/src/react-app/domains/session/chat/session-page.tsx
@@ -1,6 +1,6 @@
 /** @jsxImportSource react */
 import type { CSSProperties } from "react";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useEffectEvent, useMemo, useRef, useState } from "react";
 import { usePanelRef } from "react-resizable-panels";
 import { ArrowLeft, Cloud, FileText, Globe, Mic2, MoreHorizontal, PanelRight, TextSearch, X, Zap } from "lucide-react";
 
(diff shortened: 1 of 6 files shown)


benjaminshafii:
Screenshots from the passing E2E run:

pin shows Pin session for the open session

After pinning: Pinned group in the sidebar, palette now offers Unpin session

Rename session… opens the rename dialog

Sidebar shows the renamed session

PR #4436, opened 2026-09-04
fix(session): resume threads after sleep or connection loss

reachjalil:
## Problem

Close the laptop or lose Wi‑Fi while a thread is running and it sits on **Working** for minutes with nothing to do about it. The engine's model request does not fail — the socket just goes quiet — so the engine only notices through its own 5‑minute stream timeout, and while the machine is still offline its retry budget is spent on requests that cannot succeed. Background threads in other workspaces stay stale even longer, and a local engine keeps answering the status poll while offline, so the row keeps ticking confidently.

## What changes

Nothing in OpenCode. Everything below is how OpenWork connects to it.

- **Engine connection plugin** (`openwork-provider-connection`): through the supported `provider.options.fetch` seam, a new model request waits (bounded) for the machine to have a network route, and after a suspend/resume or network change an in‑flight request that receives nothing within a short grace window fails with a retryable *connection lost* error. The engine's existing retry/backoff then re‑asks right away and reports `retry` status as it already does. Applied to every provider in the effective config; env/auth‑only providers are untouched (a config entry would enable them without credentials).
- **Background threads heal**: a run that is still live when its retention lapses keeps its workspace stream until it settles, so unfocused threads and other workspaces converge without being opened.
- **Honest row while offline**: the browser's own connectivity is part of run health, so a local engine's *Working* row becomes the existing *Connection lost — reconnecting…* notice instead of ticking through the outage.

Long silent thinking without a disruption is never failed; only bytes received after a resume/network change count as proof the socket survived.

## Proof

Journey spec `evals/specs/thread-resumes-after-suspend.e2e.test.ts`: the engine is `SIGSTOP`ped for 40s and resumed while a witness model has gone quiet after its opening chunk. Asserts the model is re‑asked exactly once (`quiet → tool → final`), the reply lands, no duplicated user turn, no interrupted‑run card, and the retry notice is gone.

Witness/host additions for that journey: a declared `quietCompletions` fault on the shared mock agent, and a `Host.signal()` primitive so the suspend happens where the app runs (local or Daytona sandbox).

Colocated tests: `provider-connection-guard.test.ts` (monitor, guarded fetch, config hook), extended `session-sync-run-status.test.ts` (background retention, offline health), extended `mock-mcp.test.ts` (quiet completion).

Evidence is published below by `pnpm evals:e2e --publish`.


reachjalil:
## Test evidence — a-turn-that-goes-quiet-while-the-computer-sleeps-resumes-on-its-own — ✅ passed

✅ **0/1 screenshots passed** · 0 expectations passed · 0 failed · 0 pending

> screenshots not attached (gh < 2.99; run `brew upgrade gh`)

**[world]** mock(mcp) · den(daytona) · desktop(as admin) · workspace(/tmp/openwork-suspended-turn-1788553160070) · [seed:raw] evalIn(<expression>) ×3 · session("Suspended turn")
**[user]** type(composer, "Continue the deterministic task that spans a laptop sleep.") · click(Run task)
**[probe]** eventually(first model request is streaming and has gone quiet)
**[user]** see(text=/Working/) · notSee(text=The task finished after the computer resumed.)
**[probe]** eventually(engine re-asks the model after resuming and finishes the turn)
**[user]** see(text=The task finished after the computer resumed., timeoutMs=60000) · notSee(text=/Working/) · screenshot · notSee(text=/connection lost/i)
**steps** 1 ✅ the engine sleeps and wakes (41.5s)
**verdict** passed · 6 user observations (see ×2, notSee ×3, screenshot ×1) · 2 probes · steps 1/1

### ⚪ UNVALIDATED — 1. a-turn-that-goes-quiet-while-the-computer-sleeps-resumes-on-its-own artifact 1

- ⚪ **UNVALIDATED** — no visual expectations recorded.

---
_Test run created 2026-09-04T20:15:40.613Z · Source: `evals/results/test-runs/2026-09-04T20-15-40-612Z-a-turn-that-goes-quiet-while-the-computer-sleeps-resumes-on-its-own/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4436 --test-run 2026-09-04T20-15-40-612Z-a-turn-that-goes-quiet-while-the-computer-sleeps-resumes-on-its-own`_

PR #4438, opened 2026-09-04
fix(auth): prevent MCP sign-in loops caused by auth throttling

### LINKED ISSUE #4197: Rate limiting collapses to a single shared bucket behind a reverse proxy, and surfaces on `/mcp` as a misleading `invalid_mcp_token` 401
## Summary

On a self-hosted deployment behind a reverse proxy, Better Auth cannot resolve the client IP and
falls back to **one shared rate-limit bucket for the whole deployment**. Because the global limit is
20 requests / 60 s, a single active MCP client exhausts the bucket for everyone.

Worse, on the MCP transport the exhausted bucket is reported as:

```
HTTP/2 401
{"error":"invalid_mcp_token","oauthError":"invalid_token",
 "message":"The MCP bearer token is invalid or expired."}
```

The token is neither invalid nor expired. A well-behaved MCP client that receives this will do the
correct thing — discard the token and re-run the OAuth flow — which consumes *more* of the same
bucket. I lost about half an hour to this before reading the logs.

## Environment

- `ghcr.io/different-ai/openwork-den-api:0.18.37`
  (`sha256:5ae1fab94ae2badd49ae74649cfb38ca679a7afc7e9ae8926c927553a34caabd`)
- `packaging/docker/docker-compose.eval.yml` @ `9f8645e`, port bindings changed to expose it behind
  nginx (Nginx Proxy Manager), everything else per the eval stack
- `better-auth@1.7.0-beta.10`
- MySQL 8.4

## Root cause

`ee/apps/den-api` configures the headers to read, but not the proxies to trust:

```js
ipAddress: {
    ipAddressHeaders: ["x-forwarded-for", "x-real-ip", "cf-connecting-ip"],
    ipv6Subnet: 64,
},
```

`grep -c trustedProxies` in the built `auth.js`: **0**.

Since better-auth 1.7, `ipAddressHeaders` is only honoured when
`advanced.ipAddress.trustedProxies` is also set — correctly so, since trusting an unvalidated
`X-Forwarded-For` would let any client bypass the limit by forging it. Without it the library logs:

```
WARN [Better Auth]: Rate limiting could not determine a client IP and is falling back to a single
shared per-path bucket. Ensure your runtime forwards a trusted client IP header, then set
`advanced.ipAddress.ipAddressHeaders` or `advanced.ipAddress.trustedProxies` so the address can be
resolved.
```

The proxy *is* forwarding the header. Verified in the generated nginx config:

```nginx
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Real-IP       $remote_addr;
```

So the missing piece is entirely on the Den side.

## Reproduction

1. Deploy the eval stack behind any reverse proxy.
2. Complete the MCP OAuth flow and obtain a bearer token.
3. Issue ~25 MCP requests within 60 s.
4. Subsequent requests return `401 invalid_mcp_token`.

## Evidence that it is rate limiting, not the token

The JWT was decoded and was valid at the time of the 401 — `exp` 2611 s in the future, `aud`
`https://den-api.example.com/mcp/agent`, matching the endpoint. Re-running the entire OAuth flow
produced a fresh token that also 401'd immediately.

Waiting 30 s and retrying **with the same, untouched token**:

```
t+30s  http=200
```

The token recovered on its own. Only a time window can do that.

## Impact

1. Any multi-user deployment shares one bucket: one active client can lock out everyone else.
2. The error is actively misleading, and the natural client reaction makes it worse.
3. It is silent — nothing surfaces in the API response to suggest rate limiting; the only signal is a
   `WARN` at startup in the server log.

## Suggested fixes

1. **Set `advanced.ipAddress.trustedProxies`**, and make it configurable by environment variable
   (e.g. `DEN_TRUSTED_PROXIES`, comma-separated). There is currently no env var for it, so a
   self-hoster behind a proxy has no supported way to fix this — only patching the image.
2. **Return `429` with `Retry-After` instead of `401 invalid_token`** when an MCP request is
   rejected for rate limiting. The current mapping violates the client's reasonable assumption that
   `invalid_token` means "re-authenticate", and turns a transient condition into an auth storm.
3. Optionally, document the reverse-proxy requirement in the self-hosting guide, since every
   non-trivial deployment will be behind one.

## Workaround for other self-hosters

Until there is a supported setting, the only route is patching `auth.js` inside the container to add
`trustedProxies` — which does not survive an image update. The alternative, `OPENWORK_DEV_MODE=1`,
disables rate limiting altogether but also raises the signup limit from 5 to 100/hour and enables
private MCP URLs, so it is not appropriate for an internet-facing instance.

benjaminshafii:
MCP clients can lose access after roughly 20 requests because each JWT verification calls the public, rate-limited JWKS route. Its 429 response is then reported as an invalid token, encouraging unnecessary sign-in attempts.

Use Better Auth's internal signing-key API for MCP verification. Return a retryable 503 with Retry-After when signing keys cannot be loaded; malformed and expired tokens still return 401. Add the opt-in DEN_TRUSTED_PROXIES setting and document the network constraints required to resolve client addresses safely.

Fixes #4197.

Validation on the PR commit:
- `TSX_DISABLE_CACHE=1 OPENWORK_EVAL_DEN_RUNTIME_PREPARED=1 pnpm evals:pr specs/mcp-auth-rate-limit-recovery.test.ts` — local Den boundary journey; production rate limiting, separate client buckets, forged-prefix rejection, 25 successful signed MCP requests within the same minute, and invalid-token rejection.
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/mcp-access-token-contract.test.ts` — passed; simulated key outage returns 503 without an OAuth challenge, the same token recovers, and malformed/expired/wrong-audience tokens are rejected.
- `pnpm --filter @openwork-ee/den-api exec tsc --noEmit --pretty false` — passed.
- `node evals/scripts/spec-boundary-ratchet.mjs` and `git diff --check` — passed.

The HTTP journey uses a fresh isolated database and synthetic credentials. The signing-key outage is covered by the contract test, rather than a live database-outage test. Public test evidence is posted separately. No private report text, personal data, real tokens, or customer identifiers are included.


benjaminshafii:
## Test evidence — mcp-requests-keep-valid-credentials-after-the-public-auth-rate-limit-is-exhauste — ✅ passed

✅ **passed** · 1 expectations passed · 0 failed · 0 pending

### ℹ️ ASSERTION — 1. Public auth throttling isolates client addresses and does not invalidate MCP credentials

Observed 429 for the exhausted client, 200 for another client and 25 signed MCP requests, and 401 with invalid_token for a malformed token.

- ✅ **PASS** Public auth throttling isolates client addresses and does not invalidate MCP credentials — Observed 429 for the exhausted client, 200 for another client and 25 signed MCP requests, and 401 with invalid_token for a malformed token.

---
_Test run created 2026-09-04T20:26:11.237Z · Source: `evals/results/test-runs/2026-09-04T20-26-11-236Z-mcp-requests-keep-valid-credentials-after-the-public-auth-rate-limit-is-exhauste/test-run.json` · Repro: `pnpm --dir evals artifacts:publish -- --pr 4438 --test-run 2026-09-04T20-26-11-236Z-mcp-requests-keep-valid-credentials-after-the-public-auth-rate-limit-is-exhauste`_

PR #3185, opened 2026-07-27
fix(desktop): keep Enterprise activation isolated

reachjalil:
## Summary
- keep explicit bootstrap files supported while isolating them from unrelated legacy activation state
- make the immutable Enterprise artifact authoritative over bootstrap activation opt-outs in Electron and the renderer
- render Enterprise activation before sign-in and keep runtime/session commands blocked until activation completes
- keep the verified `openwork://connect` activation consumer available while locked
- preserve organization app name, logo, and icon from signed or HTTPS exchange activation claims
- persist branding and completed Enterprise activation atomically, apply the desktop icon, then start the runtime
- preserve Public and Cloud bootstrap behavior

## Verification
- `node --test apps/desktop/electron/desktop-distribution.test.mjs apps/desktop/electron/workspace-store.test.mjs` (35/35 passed)
- `bun test apps/app/tests/enterprise-activation.test.ts` (10/10 passed)
- `node --test apps/desktop/electron/connect-link.test.mjs apps/desktop/electron/desktop-distribution.test.mjs` (19/19 passed)
- `git diff --check`

## Order of operation
1. Read the configured bootstrap file and immutable packaged distribution.
2. Enterprise shows activation before sign-in.
3. The Den activation page opens a one-time `openwork://connect` exchange link.
4. Electron verifies the exact HTTPS Den claims while other desktop commands stay blocked.
5. The confirmation dialog shows the organization and its HTTPS logo/icon.
6. On approval, Electron persists server plus branding plus completed activation, applies the desktop icon, and starts the runtime.
7. The activation gate clears; only then can the sign-in gate render.

## Risk
Scoped to desktop activation precedence and the existing verified connect-link handoff. Default migration remains unchanged; explicitly configured bootstrap files remain readable; Public and Cloud defaults remain unchanged.

PR #4606, opened 2026-09-08
fix(app): render the enterprise activation gate instead of a white screen

tianlinzx:
## Problem

Since #4482 mounted `DesktopUpdaterProvider` at the outermost AppRoot layer, its `useUpdater()` hook reads the `Local` and `DesktopConfig` contexts during render. The pre-activation branch of `EnterpriseAwareAppProviders` (#3185) mounts neither, so an enterprise build crashes with `Local context is missing` (then `useDesktopConfig must be used within a DesktopConfigProvider`) before the activation gate can render.

Result: **a blank white window on first launch of an unactivated enterprise install** — the activation page ("Link this app to your organization") never appears, with no ErrorBoundary to recover. Reproduced in dev via `OPENWORK_DESKTOP_DISTRIBUTION=enterprise pnpm dev`: `#root` stays empty and the CDP console shows the crash stack `useLocal → useUpdater → DesktopUpdaterProvider`.

## Fix

Mount `DesktopConfigProvider` and `LocalProvider` in the pre-activation branch of `EnterpriseAwareAppProviders` (`apps/app/src/react-app/shell/providers.tsx`), so AppRoot-level consumers keep working before activation. `ConnectLinkProvider` stays outermost, preserving the branded connect-link activation consumer from #3185.

## Testing

- New behaviour test `apps/app/src/react-app/shell/providers.test.tsx`: renders the **real** `DesktopUpdaterProvider` inside the pre-activation branch (TDD — watched it fail with both crash messages before the fix, pass after). The fake window stub is restored in `finally` so it never leaks into other test files.
- Updated the providers source-shape assertion in `apps/app/tests/enterprise-activation.test.ts` to the new contract (`ConnectLinkProvider` + `DesktopConfigProvider` + `LocalProvider` in the pre-activation branch); the branded connect-confirm-dialog assertion is unchanged.
- Full `apps/app` suite: **1345 pass / 8 fail, identical failure set to the pre-change baseline** (verified by stashing the change and diffing the failure lists — the 8 are pre-existing).
- `pnpm --filter @openwork/app typecheck`: clean.
- Runtime-verified in dev: the enterprise activation gate now renders fully (workspace-address input + Continue), instead of the white screen.

### DIFF
--- a/apps/app/src/react-app/shell/providers.test.tsx
+++ b/apps/app/src/react-app/shell/providers.test.tsx
@@ -1,0 +1,87 @@
+/** @jsxImportSource react */
+declare const describe: (name: string, fn: () => void) => void;
+declare const test: (name: string, fn: () => void) => void;
+declare const expect: (value: unknown) => {
+  toBe: (expected: unknown) => void;
+  toContain: (expected: string) => void;
+  toEqual: (expected: unknown) => void;
+  not: { toThrow: () => void };
+};
+
+import { renderToStaticMarkup } from "react-dom/server";
+
+import { DenAuthProvider } from "@/react-app/domains/cloud/den-auth-provider";
+import { DesktopUpdaterProvider } from "@/react-app/domains/settings/state/desktop-updater-provider";
+import { EnterpriseAwareAppProviders } from "./providers";
+
+// Mirrors ENTERPRISE_DESKTOP_DISTRIBUTION from the desktop shell: the flavor
+// that gates every feature behind Den activation.
+const ENTERPRISE_DISTRIBUTION = {
+  flavor: "enterprise",
+  appName: "OpenWork Enterprise",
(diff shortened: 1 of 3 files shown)


benjaminshafii:
thx @tianlinzx we'll take a look tomorrow

yomgui:
Thank you @tianlinzx, your diagnosis and fix were exactly right. Your commit landed on `dev` in #4696 (authorship preserved, Co-authored-by on the squash), together with a packaged-build CI gate that boots the enterprise and cloud flavors on a fresh profile so this class of crash cannot ship again. Against the released 0.18.44 that gate fails in seconds naming `Local context is missing`.

Why it went through a second PR: fork PRs cannot receive our automated review clearance, and the new gate is red on `dev` without your fix, so they had to merge as one. The fix will ship in the next patch release. Sorry it took a release cycle to get to it.

PR #4662, opened 2026-09-09
fix(v1): preserve composer drafts and verify instant sends

benjaminshafii:
## Production fixes
Builds on the immediate-message implementation already merged in #4631; it does not duplicate that work.
- Keep immutable submitted draft A separate from editable continuation B while a new conversation is created and sent, including attachment preparation and explicit failure recovery.
- Do not classify a rejected send as a failed session transition; keep the error visible without permanently disabling the valid composer.
- Bound desktop-loopback snapshot read retries so a transient read does not remain paused indefinitely. Retries are cancellable reads, preserve ownership, and do not resend prompts.
- Preserve current admission-unknown, interruption, and queue protections from dev.

## Performance proof
Final tested head: `f5a009ef1e4214c399e6be05e819b69a33bde43c`.
`OPENWORK_WORLD_PLACE=local OPENWORK_EVAL_ENGINE=v1 OPENWORK_EVAL_MAX_WORKERS=1 pnpm evals:e2e workspace-new-task-hit-target --local`

**Passed: 1 test, 36/36 timing samples, 11/11 evidence assertion groups, 0 failures, 0 skips, 0 pending judgments.** Placement: `local (--local)`; isolated real Electron/v1 with a deterministic standalone model witness.

| Metric | Samples | Required | Observed max |
|---|---:|---:|---:|
| New task to focused, editable composer | 12 | <500ms | 103.1ms |
| First send to user row | 12 | <100ms | 14.8ms |
| Existing-session send to user row | 12 | <100ms | 24.2ms |

Timing runs from trusted renderer input to two consecutive qualifying animation frames, excluding CDP roundtrip. Creation/prompt requests are held before engine forwarding for at least 1500ms. Every sample is retained; these are renderer-frame opportunities, not hardware pixel-presentation telemetry.

The existing journey is extended rather than adding a new spec file. It checks no empty-session creation on opening, single submission, continuation preservation, known-rejection/Restore behavior, navigation isolation, SSE-before-HTTP response, and exact-once messages/model calls across reload. Its required local placement is declared in the journey catalog. Most of the diff is fixture instrumentation and evidence diagnostics.

## Additional checks
- App typecheck: passed.
- Focused native-session, composer-focus, and composer-state checks: 21 + 1 + 7 passed in separate invocations.
- Journey selection/reporting checks: 12 passed.
- Targeted strict TypeScript and browser callback checks: passed.

Final-head evidence is published below. Historical interrupted/red runs are not used as passing proof. This is a v1 local composer/performance slice, not certification of v2, live LLMs, hundreds of sessions, or the full safety baseline. No merge or release requested by this PR.

### DIFF
--- a/apps/app/src/app/lib/opencode-session-native.ts
+++ b/apps/app/src/app/lib/opencode-session-native.ts
@@ -9,6 +9,12 @@ import type { ResolvedWorkspaceEndpoint } from "./workspace-endpoint";
 type NativeSessionEndpoint = Pick<ResolvedWorkspaceEndpoint, "opencodeBaseUrl" | "token">;
 type RequestOptions = { signal?: AbortSignal };
 
+export type NativeSessionSnapshotTarget = {
+  owner: string;
+  endpoint: NativeSessionEndpoint;
+  sessionId: string;
+};
+
 export type NativeSessionOperations = {
   get: (sessionId: string, options?: RequestOptions) => Promise<FieldsResult<Session>>;
   messages: (sessionId: string, limit: number | undefined, options?: RequestOptions) => Promise<FieldsResult<Array<{ info: Message; parts: Part[] }>>>;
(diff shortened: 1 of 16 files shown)


benjaminshafii:
**Passed** · 1/1 tests · 2/2 assertions · 0 images

Commit `f5a009ef1e4214c399e6be05e819b69a33bde43c` · selected evidence

[Open review report](https://openwork-review-jqvsx0grh-prologe.vercel.app/r/8af6e4790d444ea2a93be3e9f5304371)

PR #4745, opened 2026-09-10
fix(app): re-enable Check now after an update install fails

yomgui:
## Root cause

Follow-on of triage cluster 7/K (`reports/e2e-red-2026-09-09/triage.md`, "Dismiss-then-`notSee` races") and the trace in #4724: once dismissal is synchronous, `background-desktop-update` test 1 fails later at the install-failure loop added by #4714 (merged as Verification — Incomplete, spec never executed): `click({role:"button", text:"Check now"})` finds the Settings button `visible=true, hitTestOk=false` for 30 s.

Reproduced with a probe (local Electron, cherry-pick of #4724 on top of `origin/dev`). Sequence after `Restart & update` fails:

1. `installUpdateAndRestart` sets `updateStatus = { state: "error", failedAction: "install" }` — the button is enabled at this point.
2. `world.openSettings()` mounts the Settings route → `useDenSession`'s `syncCurrentDenSettings` → `writeDenSettings` → `denSettingsChangedEvent` → `DesktopConfigProvider` bumps `settingsVersion` → signed-out branch applies `DEFAULT_DESKTOP_CONFIG`. The spec had injected custom branding through the dev bridge just before, so this re-apply has actions (`brandAppName`, `brandLogoUrl` cleared; `document.title` "Studio" → "OpenWork") and the `config` object identity changes.
3. `useElectronUpdaterState`'s auto-check effect depends on that identity (via `resolvePolicyReleaseChannel` → `runCheckForUpdates`) and re-runs, calling `check()`. `check()` only skipped `checking | downloading | ready`; the status was `error`, and >15 min of mocked time had passed since the last automatic check, so it ran `runCheckForUpdates(undefined, false)` → `available` → auto-download → `downloading` with the mock's `finishDownload` unresolved → `Check now` disabled (`updateState === "downloading"`).

The same happens for any real trigger (window focus/online/interval) after a failed install: the update is already downloaded, so a background re-check just re-downloads it and re-offers the same restart.

## What changed

- `electron-updater-state.ts`: the automatic check additionally skips `state === "error" && failedAction === "install"`. Download failures are unchanged; manual `Check now` and the native menu request still run in that state.
- `background-desktop-update.e2e.test.ts` (claims re-expressed, nothing loosened):
  - The install-failure loop now calls `world.returnToApp()` (focus + online, +16 min) before opening Settings, then asserts Settings shows `Couldn't install the update` and that `checks`/`downloads` did not move — the observable form of "a failed install does not restart the download loop; the person retries from Settings". This step is red on `dev` and green here.
  - The custom-branding claim (`customLogoLoaded`, `Restart Studio?`) moves to the final confirmed restart, where it sat before #4714. The eval injects branding through the dev-only desktop-config bridge for a signed-out app, and that injection does not survive the Den settings sync that opening Settings performs (step 2 above), so asserting `Restart Studio?` inside a loop that visits Settings was never satisfiable. The failure loop asserts the default `Restart OpenWork?` prompt instead; the brand claim is still asserted once, as before.
- Includes #4724's commit `a42676ec1` (`fix(app): close the restart prompt and command palette synchronously on dismiss`) as an unmodified cherry-pick (`b2527b337`): the spec cannot reach the failing step without it. Those two files are untouched here; once #4724 lands on `dev` the cherry-pick becomes a no-op in this diff.

## Proof

`pnpm evals:e2e background-desktop-update --daytona` with `OPENWORK_EVAL_REF=2df0042ea5ac11f36754b346086360d12da7c7d9` (sandbox log: `HEAD is now at 2df0042e test(evals): prove a failed install …`):

- placement: `{"command":"evals:e2e","lane":"e2e","daytona":true,"placement":"daytona","engine":"legacy","surface":"legacy","case":null,"vision":"defer","files":["background-desktop-update"],"passed":2,"failed":0,"skipped":0,"skips":[],"consented":["OPENWORK_EVAL_E2E_TESTS"],"verdict":"passed"}`
- ✓ `updates download outside Settings and offer a persistent, optional restart` (175 s) — trace shows both `see(text=Couldn't install the update)` → `click(text=Check now)` → `eventually(the user re-downloads after the failed install through Settings)` iterations passing, then `see(text=Restart Studio?)` and `eventually(restart only after confirmation)`.
- ✓ `a confirmed update relaunch resumes only the unfinished task on its original engine` (259 s)
- Verdict: **Passed** (2 passed / 0 failed / 0 skipped). All three sandboxes of the run were deleted by the testkit.

Also: `pnpm --dir apps/app typecheck` exit 0.

## Noticed, not touched

1. The automatic check's `check()` runs on every effect re-run (any dependency identity change), not only on mount/focus/online/interval. The 15-minute guard hides this in practice; it surfaced here because the world advances `Date.now`. Left as is.
2. Local lane on macOS cannot run this spec: `[data-sidebar-brand]`/`Notifications` are not rendered in the collapsed mac layout, the titlebar `Restart to update` is covered by `Toggle Sidebar`, and CDP clicks fire window `focus` events when the Electron window is not frontmost (which trigger automatic checks). Daytona is unaffected.
3. After a download failure (`failedAction: "download"`) the next focus/online/interval trigger still auto-retries and auto-downloads. Loop 1 of the spec passes only because no such trigger occurs between the failure and `Check now`; the spec does not claim otherwise.
4. Signed-out apps re-apply `DEFAULT_DESKTOP_CONFIG` on every Den settings sync, so a dev-bridge-injected desktop config (any eval using `__openworkApplyDesktopConfig` without a sign-in) is cleared when Settings opens.


yomgui:
## Resync with `dev` — merge commit `c6174cafdaa5dffd0be9650c1fe2efe6d72f3361`

`origin/dev` (`6909f4ea0`) merged into `e2e/settings-check-now-after-failed-install` as a merge commit (no rebase, no force-push). PR is MERGEABLE again.

**Conflict resolved — `evals/specs/background-desktop-update.e2e.test.ts`** (dev side: #4761). Both sides changed the install-failure loop of test 1; every claim from both is kept:
- **From dev (#4761):** `revokedTest` (auto-merged); `Keep working` dismissal via `probe.eventually`; before the loop, `Check automatically` is switched off (`automaticChecksEnabled: false`) and custom branding is applied, so the loop asserts the branded `Restart Studio?` and dismisses it via `probe.eventually("the failed restart dismisses its dialog")`.
- **From this PR:** inside each loop iteration, after the failed install, `world.returnToApp()` → `openSettings()` → `see("Couldn't install the update")` → `checks: index + 6, downloads: index + 3` unchanged → `Check now` re-downloads. The product change (`electron-updater-state.ts`: automatic check skips `error/install`) and the #4724 sync-dismiss cherry-pick are untouched; dev's pre-install policy re-validation from #4761 sits next to it in the same file (auto-merged).
- **Not kept from this PR:** moving `setCustomBranding()` after the loop and asserting the default `Restart OpenWork?` inside it. That move was a workaround for branding being cleared by the Settings-open Den sync at this PR's base; dev's world now feeds the branded config back through `__openworkSetDesktopConfigRefreshResult`, and dev's proof (and this run) show `Restart Studio?` holds inside the loop. Keeping dev's placement asserts the brand claim three times instead of once and drops the duplicated post-loop block. Net spec delta vs `dev` is exactly the 5 `returnToApp` lines above.

**⚠️ Flag — the spec no longer exercises this PR's product guard.** With dev's `Check automatically` switched off before the loop, `returnToApp()` (focus + online + 16 min) cannot trigger an automatic check regardless of the `error/install` skip added in `electron-updater-state.ts`. The assertions still hold and nothing was loosened, but "a failed install does not restart the download loop on its own" is now guaranteed by the disabled switch in the test, not observed against the guard. The guard still matters in the product (real users leave automatic checks on; dev's `installUpdateAndRestart` refresh also re-runs the auto-check effect). If reviewers want the guard proven, a follow-up can re-enable the switch inside the loop for one `returnToApp()` — I did not add that here to avoid inventing flow during a conflict resolution.

**Proof at the merge sha**
- Unit: `pnpm --dir apps/app typecheck` exit 0; `bun test --isolate tests/electron-updater-install-policy.test.ts tests/version-gate.test.ts` → **26 passed / 0 failed**; `pnpm --dir evals typecheck` reports no errors in the spec.
- Daytona: `OPENWORK_EVAL_REF=c6174cafdaa5dffd0be9650c1fe2efe6d72f3361 pnpm evals:e2e background-desktop-update --daytona`
  - `placement: daytona (--daytona)` · `[openwork/testkit] ==> source verified c6174cafdaa5dffd0be9650c1fe2efe6d72f3361`
  - **3 passed / 0 failed / 0 skipped** (`verdict: passed`): revoked-version 226 s, background/persistent-restart 162 s, relaunch-recovery 281 s.
  - Trace of test 1 (`results/test-runs/2026-09-10T16-35-35-611Z-…/test-run.json`, `gitSha c6174cafd`): `click(label=Check automatically)` ×1, `see(text=Restart Studio?)` ×3, `the failed restart dismisses its dialog` ×2, `see(text=Couldn't install the update)` ×2, `the user re-downloads after the failed install through Settings` ×2.
  - All five sandboxes deleted by the harness.

PR #4757, opened 2026-09-10
fix(app): keep a brand-new session sendable while its first snapshot loads

yomgui:
## Root cause

Triage: `reports/e2e-red-2026-09-09/triage.md` → Cluster 1 "Fresh-session `Run task` click swallowed" (Brief B).

1. **A brand-new session refused to send for its first ~0.9 s on Daytona.** `handleCreateTaskInWorkspaceWithOpenMode` navigated to the new session with an empty snapshot cache, so `deriveSessionRenderModel` reported `"switching"` (`!hasSnapshot && isFetching`) until the four-request snapshot round trip finished, and the composer's `disabled={model.transitionState !== "idle" …}` gate dropped any Run task click / Enter in that window with the draft intact (`triage-probe-run4-immediate-click.log`: disabled at t=386 ms, enabled at t=901 ms, nothing sent).
2. **A send that settled with a 5xx wedged the session.** Since #4635 a `prompt_async` answered 5xx/408 is an uncertain admission, but the resulting `admission_unknown` hold could never end: the acceptance check treated absence as no evidence, Stop refused while admission was unknown, and the response body was dropped, so the person saw a paused composer with no explanation (`session-error-technical-details` test 2: `runEnabled:false` after 15 s).
3. `readComposerState` reported an empty Lexical editor as `"\n"` (`streamed-markdown-answer` test 1 `draftText === ""`).

## What changed

- `seedCreatedSessionSnapshot` (session-sync) seeds the snapshot cache from the create-session response before navigating, stale (`updatedAt: 0`) so the real fetch still runs as a background refresh of a session already on screen. `hasSnapshot` is true from the first render → composer sendable immediately.
- `PromptAdmissionUnknownError` keeps the settled 5xx body as its `cause` (`promptAdmissionFailure`); the surface shows it as the session error card right away while the hold still prevents any resend.
- `readPromptAdmission` (accepted / absent / unknown): after the exact-message GET misses, list the idle conversation; when native lists it without the message, the settled POST did not admit it → new drain event `admission_rejected` halts like any definite failure (explicit retry only, never an automatic resend). Surface and global drainer probes dispatch it; on release the pending row is removed and the submitted composer is saved as an unsent message ("Restore unsent message"). Confirmed acceptance clears the stale failure card.
[...]

PR #4434, opened 2026-09-04
feat(browser): give each conversation its own built-in browser tabs

reachjalil:
## Problem

You read one conversation while another conversation's agent needs a web page. The browser pops open in the conversation you are reading, showing a page that has nothing to do with it, and the working conversation shows nothing. The built-in browser kept one global tab list with no owner, `openwork_execute` dropped the agent's session id before the UI command bridge, and the side panel mirrored every native tab into whichever conversation was mounted. Any navigation of a hidden tab also auto-selected it onto the screen.

## What changes

- **Tabs belong to the conversation that opened them.** The bridge stamps UI commands with the requesting session (`origin`, set from the tool context, never by the agent), the control bus hands that origin to the action, and the Electron browser keeps an owner per tab. Each side panel shows only its own tabs.
- **Background conversations browse silently.** Only the on-screen conversation's tabs may take the screen. A tab opened for a background conversation loads there, its panel state is marked open, and it is waiting when the user switches. The navigation auto-select is owner-aware; `panel-opened` / `panel-closed` name the owner.
- **Hidden tabs stay real pages for the agent.** Chromium only paints a page it considers visible, so a hidden tab keeps a one-pixel presence in the window corner (under the rounded-corner mask) while a main-process DevTools session emulates a 1280×800 viewport and focus. Reads, clicks, typing, and CDP screenshots work before anyone looks; both are undone when the tab returns to the screen and the existing viewport-reset path takes over.
- **New `@openwork/browser-tabs` package** holds the ownership, surfacing, and background-rendering policy, shared by the main process and the renderer and unit-tested on its own.

## What stays the same

Tabs opened without a conversation (no session on screen) stay shared and behave exactly as before. The visible conversation's own tabs surface as before. `browser_*` tools and the CDP handle contract are unchanged; `open_url` additionally returns `owner_session_id` and `visible`.

## Proof

New journey `evals/specs/browser-tabs-owned-by-thread.e2e.test.ts`: a background conversation opens a page — the on-screen conversation's tab, panel, and viewport are untouched; the hidden page lays out at 1280×800, reports focus, accepts a dispatched click and typed text, and screenshots at full size; switching to the owner shows its tab at the panel's viewport; switching back restores the first conversation's own tab only.

[...]

PR #4777, opened 2026-09-10
fix(app): preserve the v2 sessionless first prompt

yomgui:
## Current revision — CI advisory addressed

Current head: `c86bc3c72713ee3a0cd11f4459d43b1334966322`.

Addressed **V2L-CBB** (medium, spec-provenance-review): require explicit `OPENWORK_EVAL_ENGINE`, prefix each test verdict with v1/v2, and assert the live native engine-routing status before typing. No app or Warden-policy change in this follow-up commit.

- Omitted engine: named command returns **exit 2 / Incomplete**, 0 passed and 1 skipped, rather than a false v2 pass. This negative guard check is not counted as a passing journey.
- Final-head explicit v2 and v1 journeys: **1 passed / 0 failed / 0 skipped each**.
- Final-head v1 instant-send journey: **1 passed / 0 failed / 0 skipped**.
- App typecheck, 64 targeted unit tests, 13 catalog tests, 831 browser callback checks, and scoped eval typecheck passed again.
- Final bare Warden run `cdbda42d` (2026-09-10T14:54:50Z): **exit 0, all four expected skills completed, no findings/skipped files**. Before/after refs unchanged: HEAD `c86bc3c7…`, origin/dev `326ff094…`. Repeated because origin/dev moved during the preceding local review. Private logs remain private.
- The current combined test-evidence comment contains the three matching final-head records. Earlier proof below is retained as historical provenance.
- The review-machinery human-approval guard remains intact; no admin override, removal of coverage, or self-approval.

## Problem

With v2 selected, clicking an **enabled** Run task on `#/workspace/<id>/session` creates and opens a session and clears the composer, but never sends the prompt. The new thread stays at “Try one of these” and its native message list stays empty. This is **not** #4757's disabled first-snapshot send gate.

## Root cause and evidence

`createRouteSession` resolves the owning server's engine and creates a v2 session, but the hero previously keyed its submitted composer snapshot using the original workspace endpoint (`.../opencode`). `SessionSurface` looks for the scoped payload using its routed endpoint (`.../opencode2`). Those keys never match. The empty continuation is correctly transferred; the submitted snapshot is stranded under the wrong key, so auto-send never runs. This is an app handoff defect, not an engine defect.

[...]

PR #4779, opened 2026-09-10
fix(evals): saved-app creation opens its verification return tab as a person's tab, not agent browser control

yomgui:
## Root cause

Triage: `/Users/guillaume/OpenWork Chat/reports/e2e-red-2026-09-09/triage.md` — `saved-app-creation`, failure at `worlds/saved-apps.ts:163` (`returnVerification`) with
`Error invoking remote method 'openwork:browser:openUrl': Browser control requires a requesting conversation.` (dev nightly @ 52a028817, PR #4757 run @ 564e17c7d).

The spec's world channel, not the product, lacked the origin. `returnVerification` created its vehicle tab through the **agent** entry point — `window.__OPENWORK_ELECTRON__.browser.openUrl("about:blank", "builtin")`, the same IPC behind the `browser.open_url` affordance — without a requesting conversation. Since #4434 (1d2fc5b14) the browser task host correctly refuses that (`browser-task.mjs` `missing_session`), and it never accepted an `about:blank` destination either (`browserTaskUrl` only allows http(s)); stamping an origin would just have moved the failure to `invalid_url` plus a consent prompt.

The tab is only a stand-in for the person's own browser: containers have no OS protocol registration, so the world navigates the real `openwork://den-reauth` return link in an Electron tab and the main process intercepts it on `did-start-navigation` for **any** tab, owner or not. The honest channel for that is the person's new-tab path (`browser.createTab`, a null-owner "shared" tab) — already used by `worlds/browser-panel.ts` as "explicit human-created initial state, not an automation navigation". No product change; the requesting-conversation requirement is not bypassed or weakened.

## What changed

- `evals/worlds/saved-apps.ts` `returnVerification`: create the return tab with `browser.createTab("about:blank")` and find its CDP target by diffing page targets (that IPC returns only `tabId`). Navigation, deep-link interception and tab close are unchanged.
- Merged `origin/e2e/fresh-session-send-gate` (564e17c7d) so the spec gets past the send step (Cluster 1); those commits are not part of this fix.

## Proof

`pnpm evals:e2e saved-app-creation --daytona` with `OPENWORK_EVAL_REF=8ced4002cc82eb18fb24c66e844b4f67f988538b` (source verified `8ced4002c` in the sandbox log). Placement line: `selection: engine=legacy surface=legacy case=all; placement: daytona (--daytona)`.

Three runs (each provisioned its own Den + two desktop sandboxes and deleted all three):

| run | steps 1–11 (incl. both `returnVerification` calls in "share dashboard apps with a teammate") | step 12 "return from browser verification…" (3rd `returnVerification`) | outcome |
|---|---|---|---|
| 04:37Z | ok | failed on the Den **web** page: `Verify password` produced Better-Auth "Provider not found" (a social sign-in fired) 370 ms after a client-side route change | 0 passed / 1 failed / 0 skipped |
| 04:49Z | ok | ok | 0 passed / 1 failed / 0 skipped — failed at step 13 (below) |
| 04:59Z | ok | ok | 0 passed / 1 failed / 0 skipped — failed at step 13 (below) |

The cluster claim is proven: every `returnVerification` call now completes and the subsequent assertions (unrelated callback rejected, verified link shared 1 app, late callback left `#/dashboard`, browser-return grant gave the recipient 200) pass. The **whole spec is not green yet**: verdict `Failed` at later steps that have different root causes, so this PR does not claim `saved-app-creation` Passed.

## Noticed, not touched

1. **Step 13 "an admin cancels deletion in the app and confirms it on the dashboard" fails deterministically (2/2)** at `saved-app-creation.e2e.test.ts:523`: after `world.open("/dashboard")` (`go()` only sets the hash and returns) the `click("App options for Team briefing")` resolves in ~110 ms — most likely the still-mounted app-page header button — and then `Delete Team briefing` is never found; after 30 s the only visible buttons are the shell rail (`New session`, `Dashboard`, `Library`, …), i.e. the main content renders nothing. Looks like a product issue on route change with an open header menu, not just spec ordering; needs a screenshot/error-boundary look.
2. **Round-2 web verification flake (1/3)**: on `/reauth/desktop` with an existing cookie session, `type Password` + `click("Verify password")` within ~0.4 s of the client-side nonce change ended in "Provider not found" — a social button received the click. Testkit text click vs. dialog layout race, or the dialog should not shift while opening.
3. `pnpm --dir evals typecheck` is red on `dev` independent of this change (unbuilt `@openwork-ee/den-db`/`@openwork/email` outputs, `Promise.withResolvers` lib target); `worlds/saved-apps.ts` itself has no errors and `check:browser` passes.


yomgui:
## Resync onto `dev` after the #4757 conflict resolution

- Merged `origin/e2e/fresh-session-send-gate` (#4757 @ `4a2f43213`, AI's signed resolution) → `b53b16943`, then `origin/dev` (`0b219dc4e`, includes #4777, #4747, #4837) → **`1afdc2757`**. Both merges applied cleanly; **no conflicts** to resolve. Both merge commits are SSH-signed (`%G?` = `G`).
- `pnpm --filter @openwork/app typecheck` → exit 0. No `apps/app` files change vs `dev` on this branch.
- Net diff vs `dev` is still only `evals/worlds/saved-apps.ts` (`returnVerification` via `browser.createTab`); the squash diff will shrink to that once #4757 lands.

## Proof at the pushed head

`OPENWORK_EVAL_REF=1afdc2757bae3c517d03951dcdaa58860b2199da pnpm evals:e2e saved-app-creation --daytona` (`selection: engine=legacy surface=legacy case=all; placement: daytona (--daytona)`; `source verified 1afdc2757…` in all three sandboxes).

| test | result |
|---|---|
| **`create, preview, save and reopen an app without changing already-open results`** (this PR's spec) | **✓ passed (331.9 s)** — all steps ok, including every `returnVerification` call |
| `APP-ISOLATION embedded MCP Apps isolate siblings…` (added to this file by #4829 today) | × failed in world provisioning after 53 s, before any assertion: `Mock issuer gate failed: health reported https://3979-…, expected https://3979-…` |

`"passed":1,"failed":1,"skipped":0` → file verdict `Failed`, but the red test is not this PR's: `isolatedMcpApps` seeds **two** `seed.mock(...)` without explicit ports into one `appWeb` sandbox, so both get `daytonaPort: 3979` (`evals/packages/env/src/mock.ts:56`); `bootRemoteMocks` starts them sequentially on the same sandbox and the second mock's `pkill -f mock-oauth-mcp-server` + rebind on the same port makes the health gate see the first mock's issuer. Reproduced identically 2/2 at this sha (first run also hit a hung `mock process cleanup` exec on the Den sandbox → 600 s timeout; re-run per policy). #4829 proved that test with `--local --engine v1 --case APP-ISOLATION` only, never `--daytona`; the co-located Daytona path is not exercised locally. **Attribute to #4829** — give `second` a distinct `port` (e.g. `3980`). This PR does not touch that world/spec code.

Sandboxes: all created by run 2 deleted by the runner; run 1's leaked `openwork-server-20260910-213758-…` deleted manually.

PR #4786, opened 2026-09-10
fix(app): hide the child checklist when a stopped parent receives a follow-up

yomgui:
## Root cause

`parent-child-permission-approval` test 4 fails at spec line 354 (`notSee("Child checklist")`) after **Stop + follow-up**, identically on the `dev` baseline and on #4723's run; unowned in the [2026-09-09 E2E triage](file:///Users/guillaume/OpenWork%20Chat/reports/e2e-red-2026-09-09/triage.md) (cluster: `parent-child-permission-approval` test 4). The claim was added in #4649 without runtime proof.

OpenCode v1 executes tools through `EffectBridge.run.promise` (`session/tools.ts`), i.e. on a detached fiber. Aborting a session interrupts the prompt loop and marks the tool part `error: Tool execution aborted`, but the `question` tool's `Question.ask` keeps waiting: its request stays in `GET /question` until it is replied to or rejected, and no `question.rejected` event is emitted. Reproduced against the local engine (`opencode` 1.18.25, same code path as the pinned 1.18.18) with the eval mock agent: after `POST /session/:id/abort`, `/question` still listed the request 5 s later while the session was idle; `POST /question/:id/reject` then removed it, kept the session idle, and caused no new agent request.

So the parent faithfully kept showing the stopped child's pending question (and its "Child checklist"/"Child outline" buttons). This is a product gap, not intended retention: the child's tool call is already aborted, nothing can consume an answer, and a reload would resurrect the prompt from the engine list. The spec's claim stands unchanged.

## What changed

- `apps/app/src/app/lib/opencode-interruption.ts`: after `stopForegroundTree` confirms every stopped session idle, list pending questions and reject the ones owned by the stopped tree (root + interrupted descendants). Other roots are never touched. Best-effort: a failed withdrawal does not fail Stop (the tree is already stopped). Works through the shared `Client` on both engines (`question.list` / `question.reject`; the v2 adapter maps reject to form cancel).
- `apps/app/tests/opencode-session-native.test.ts`: the tree-stop test now serves a `/question` list with a stopped-tree question and an unrelated root's question, and asserts only the stopped one is rejected, after idle confirmation.

An earlier attempt (re-reading pending questions on terminal edges in `session-sync.ts`, `9e709825e`) was proven insufficient on Daytona at that sha (same failure at line 354, because the engine list still contains the orphan) and is reverted in `ed89128d4`.

This branch also merges `origin/e2e/delegated-task-shimmer` (#4723) and `origin/e2e/fresh-session-send-gate` (#4757) unchanged, only so test 4 is reachable; their files are not modified here.

## Proof

`pnpm evals:e2e parent-child-permission-approval --daytona` with `OPENWORK_EVAL_REF=783fd41fa34d3d4aab0a1dfb412df2c80d15b7c3` (source verified in-sandbox), placement line:

```
{"command":"evals:e2e","lane":"e2e","daytona":true,"placement":"daytona","engine":"legacy","surface":"legacy","case":null,"vision":"defer","files":["parent-child-permission-approval"],"passed":2,"failed":2,"skipped":0,"skips":[],"consented":["OPENWORK_EVAL_E2E_TESTS"],"verdict":"failed"}
```

| Test | Result at `783fd41fa` |
| --- | --- |
| 4 — a parent answers and stops real child questions, then finishes fresh work without settling an unrelated root question | **passed** (293 s); all 7 steps green, including `notSee(Child checklist)` (line 354) and `eventually(interruption removes only the stopped child's question)` |
| 1 — switching and creating threads hydrate permissions | timed out in world provisioning (no assertion reached) in this run; passed in the three earlier runs on this branch (`73b1d9397`, `9e709825e`, `783fd41fa` first run) |
| 2 — retry recovery and stopping a permission | failed at line 137 (`see("printf OTHER_PERMISSION_WITNESS")`) in every run — #4757's cluster, unchanged by this PR |
| 3 — a parent task surfaces and resolves its child permission | passed |

Baseline repro on the merged branch before the fix (`73b1d9397`): test 4 failed at line 354 with `Target remained visible`, tests 1 and 3 passed.

A first run at `783fd41fa` never reached Stop: a single `notSee` CDP call after `user.reload()` stalled for 938 s (normally 3 s) and the 20 min test timeout fired; re-run per the lane's known instability.

Unit: `bun test --isolate tests/opencode-session-native.test.ts tests/queued-drain-admission-wedge.test.ts` → 41 pass, 0 fail; `pnpm --filter @openwork/app typecheck` clean.

Overall verdict for this spec file: **Incomplete** (test 4 Passed; tests 1/2 red for reasons outside this PR, see above). All Daytona sandboxes from these runs were deleted.

## Noticed, not touched

- The same engine behaviour applies to **permissions**: `Permission.ask` also stays pending in `/permission` after abort in v1.18.x. Test 2's Stop step ("native interruption cleanup removes the pending permission", line 150) therefore cannot pass on v1 without an app-side withdrawal too — but its prose says "without a permission reply", so the #4757 owner should decide (reject-on-Stop vs. re-expressing the claim).
- Test 2 currently fails earlier, at line 137: in the on-screen dump the `bash` step ran without asking (`Ran command … Permission work finished.`), i.e. the `permission: { bash: "ask" }` policy did not take effect in the "other" session — different from the triage's line-138 failure.
- Test 1 timed out once in provisioning (600 s) in a run where Daytona was slow; three other runs passed.


yomgui:
## Resync against `origin/dev` `0b219dc4e` (merge commits, no rebase)

Two merge commits, in the prescribed order:

1. `f489b4542` — `Merge origin/e2e/fresh-session-send-gate into e2e/child-checklist-after-stop` (the **updated** #4757 head `4a2f43213`, which already carries `origin/dev` at `9a60a4b59`).
2. `8763acc6d` — `Merge origin/dev into e2e/child-checklist-after-stop` (auto-merged, no conflicts). Both signed `G`.

**Premise re-check (Jalil #4837 `7f8008d58`, Ben #4812 `cbf327cd4`).** #4837 adds `submitImmediateSessionTurn` / `foregroundDelegations` so an immediate follow-up interrupts foreground delegation before admission; #4812 fixes delegated child identity and activity timing. Neither withdraws the stopped tree's pending `question` requests — `dev` has no `question.reject` call in `apps/app/src` (only the v2 adapter's event mapping and `session-sync`'s consumer of `question.rejected`). The engine still leaves the aborted child's `Question.ask` pending in `GET /question`, so the parent keeps rendering the "Child checklist"/"Child outline" prompt after Stop. Still needed; complementary to #4837 (this PR acts after `stopForegroundTree` confirms idle, #4837 decides *when* to stop).

**Conflict** (from merge 1) — `apps/app/tests/opencode-session-native.test.ts`, the tree-stop test's request allow-list: this branch allows `${baseUrl}/question…` (the withdrawal calls), `dev` allows `${baseUrl}/path?…` (#4825's workspace-directory ownership check). Kept **both** (`/session/` || `/question` || `/path?`). `session-surface.tsx` had no conflict once the updated #4757 was in (its import line already carried `submitImmediateSessionTurn` alongside `promptAdmissionFailure`/`readPromptAdmission`).

**Proof at `8763acc6d`** (`OPENWORK_EVAL_REF` pinned; sandbox log `HEAD is now at 8763acc6`):

`OPENWORK_EVAL_REF=8763acc6dabe494501398da76c327b21046beec7 pnpm evals:e2e parent-child-permission-approval --daytona`

`{"placement":"daytona","engine":"legacy","files":["parent-child-permission-approval"],"passed":3,"failed":1,"skipped":0,"verdict":"failed"}`

| Test | Result at `8763acc6d` (with #4837 + #4812 present) |
| --- | --- |
| 4 — a parent answers and stops real child questions, then finishes fresh work without settling an unrelated root question | **passed** (328 s). All 7 steps green: `an unrelated root waits for its own answer`, `the parent displays only its delegated child's question`, `reloading the parent restores the same unanswered requests`, `answering in the parent resumes the original child and then the parent`, `the parent waits on another foreground child while the unrelated question stays pending`, **`Stop immediately followed by fresh work interrupts the child and completes only the new turn once`** (the #4837 path + this PR's withdrawal, incl. `notSee("Child checklist")`), `the unrelated root remains answerable without resuming the stopped child or duplicating the follow-up`. |
| 1 — switching and creating threads hydrate permissions | passed |
| 3 — a parent task surfaces and resolves its child permission | passed |
| 2 — retry recovery and stopping a permission | failed at spec line 137, step `two conversations own separate real approvals`: `Timed out after 45000ms seeing text=printf OTHER_PERMISSION_WITNESS` — byte-identical to the failure recorded in this PR's body before the resync ("#4757's cluster, unchanged by this PR"). Not attributable to #4837: same signature before #4837 existed on this branch. |

Verdict for this PR's claim: **Passed** on the merged head. File-level: **Incomplete** (test 2, #4757's cluster).

**Delegated-task journey (`task-activity-shimmer`, ACT-01) at `8763acc6d`** — `OPENWORK_EVAL_REF=8763acc6d… pnpm evals:e2e task-activity-shimmer --daytona` → 1 failed, died in **provisioning**, no assertion reached:

```
launch remote app-web runtime task-activity-web-25580-64b13465 failed with exit 1:
Error: EACCES: permission denied, mkdir '/private'   (at /tmp/task-activity-web-…-launch.mjs:6)
```

Root cause is in the world, not the product: `evals/worlds/task-activity-web.ts` (#4812) does `realpath(seed.tmpPath("task-activity-web"))` on the **host** and passes that as `workspacePath` to `seed.appWeb`; on a macOS host `/tmp/…` realpaths to `/private/tmp/…`, and the Linux sandbox's launch module then `mkdir`s that host path. The registered case's own example is `--local`. Deterministic, so I did not burn a second sandbox on it. **Attribute to #4812's world (harness portability), not to #4837 and not to this PR**; a Daytona verdict for ACT-01 from a macOS host needs that world to compute the workspace path the way `worlds/chat.ts` does for remote surfaces. Not run locally from inside this session (would compete with the hosting app's engine); happy to run `pnpm evals:e2e task-activity-shimmer --local --case ACT-01` on request.

Also at `8763acc6d`: `bun test --isolate tests/opencode-session-native.test.ts tests/queued-drain-admission-wedge.test.ts` 56 pass / 0 fail; `pnpm --filter @openwork/app typecheck` exit 0; `git diff --check origin/dev...HEAD` clean. All 7 sandboxes from both runs deleted (verified by name).

Teammate PRs compared: #4837, #4812 (and #4825 for the `/path?` allow-list line). Still stacked on #4723 and #4757 (both open); once they land this PR reduces to `opencode-interruption.ts` + the test hunk.

yomgui:
**#4747 revert check (#4870 on dev `858414c41`):** this branch does not contain `561b5c03a` (`git merge-base --is-ancestor 561b5c03a… HEAD` → false; its merged ancestors do not either) and `git diff origin/dev...HEAD -- apps/app/src/react-app/shell/desktop-runtime-boot.ts` is empty. No in-branch revert needed.

PR #4770, opened 2026-09-10
test(evals): fail loudly when a spec clicks a disabled control

yomgui:
**Merge order: land this AFTER #4757 (`e2e/fresh-session-send-gate`).** Until #4757 is in, the Cluster 1 specs that click `Run task` during the first ~0.9 s of a new session will fail *immediately* with `DisabledTargetError` instead of after a 90 s timeout — that is the intended behaviour of this PR, but it means their red turns from "timeout" into a named failure until the product fix lands.

## Root cause

Triage: `reports/e2e-red-2026-09-09/triage.md` → **Cluster 1** ("Fresh-session `Run task` click swallowed", 7 files) and **Brief B2**.

- `user.click` located the target with `waitForLocated` (visible + hit-test) and dispatched `Input.dispatchMouseEvent` at its centre without ever looking at `disabled` / `aria-disabled`. A fresh session's `Run task` button is `disabled` for ~0.9 s (`transitionState === "switching"`, see triage probe `run4`: `t=386ms clicked … buttons=[Run task:disabled] user=0`), so the click was silently swallowed and the spec died 60–120 s later on an unrelated `see(...)` timeout with an *enabled-looking* button in the dump.
- Related hardening flagged in #4728 and #4747 ("noticed, not touched"): the harness located, emitted the trace entry, then dispatched ~75–250 ms later without re-checking geometry, so a layout shift (Analytics tab strip moving between pages; async sections above the Advanced run-mode switch) could absorb the click just as silently.

## What changed (`evals/packages/cdp`, `evals/packages/testkit` only)

- `locate()` now reports `Located.disabled: string | null` — what disables the control (`disabled` via `:disabled`, so a disabled `<fieldset>` counts, and/or `aria-disabled="true"`), or `null`. The value is validated like every other located field (a non-string/non-null value is "invalid located-element geometry"). This matches how the app itself decides a control is actionable (`session-number-shortcuts.ts`: `!button.matches('[disabled], [aria-disabled="true"]')`).
- New `clickTarget(surface, target, { timeoutMs, mustHitTest, button, clickCount })`:
  1. `waitForLocated` as before (visible + hit-test);
  2. throw **`DisabledTargetError`** immediately if the control is disabled — `Refused to click disabled button "Run task" (disabled); the page would ignore the click.` No mouse event is dispatched. It does **not** wait for the control to become enabled: a spec that needs to wait must say so, otherwise the guard would hide exactly the bug Cluster 1 is about.
[...]

PR #4794, opened 2026-09-10
test(evals): saved-app world.open waits for the destination page before the spec acts on it

yomgui:
**Stacked on `origin/e2e/saved-app-browser-requesting-conversation` (#4779, which carries #4757) via merge commit `80bab88f4`; those commits are not part of this fix.** Net change over the merge: one hunk in `evals/worlds/saved-apps.ts`. The brief's working title was `fix(app): dashboard stays interactive after navigating with an app-page menu open`; the evidence below shows the dashboard was never broken, so no product code changes and the title says what actually changed.

## Root cause

Triage: `reports/e2e-red-2026-09-09/triage.md` → `saved-app-creation` (Cluster 1, then #4779 "Noticed, not touched" item 1: step 13 fails deterministically 2/2 at `saved-app-creation.e2e.test.ts:523`).

`world.open(path)` was `go(app, path)`, which only sets `window.location.hash` and returns. The page being left stays mounted until React Router commits the new location (~100–200 ms on Daytona), and the app page and the dashboard share control labels (`App options for Team briefing`) and preview text (`Weekly overview` / `Launch briefing`). So the spec's next observation could land on the page it had just left:

- **Step 13** (`world.open("/dashboard")` → `click("App options for Team briefing")`): the click resolved against the still-mounted app-page header button (trace target `1066,136`, identical to the pre-navigation click), its menu opened and unmounted with the page, and `Delete Team briefing` was then never found on the dashboard because no menu was open there.
- **Same race at line ~198** (`world.open(\`/dashboard${originalPath}\`)` → `eventually(previewText includes …)`): satisfied in 135 ms by the dashboard card's iframe before it unmounted; `showDetails()`/`previewText()` then read `''` from a vanished target (my run 2, below). #4779's three runs took 648–1340 ms here, i.e. they happened to wait for the real app-page iframe.

**The dashboard does not break.** A diagnostic run (sha `fead49b90`, temporary, removed in `c63458e38`) snapshotted the renderer around the route change with a raw hash set: right after `hash = #/dashboard` → `appHeader: true, dashboardPage: false` (app page still mounted); after the route committed → `dashboardPage: true, menus: 1, menuItems: ["Remove Team briefing from dashboard", "Delete Team briefing"]`, all dashboard buttons present (`Share`, `Add`, `Open Team briefing`, `App options for …`, `Refresh/Run Project updates`), `console.error`/`window.error`/`unhandledrejection` bucket empty, `body pointer-events: auto`, no `inert` / `aria-hidden` leftovers. The "main content renders no buttons" reading in #4779 was an artefact of `cdp/src/input.ts#locate` listing only the **first 8** visible buttons on a miss — the shell rail comes first in DOM order.

## What changed

`evals/worlds/saved-apps.ts` `open()`: after `go()`, wait (bounded, 30 s) for the destination's own root — `[data-app-header]` for `/dashboard/apps/…`, `[data-dashboard-page]` for `/dashboard` — before returning. Other paths are unchanged. No spec assertion was loosened or reordered and no sleeps were added; `open` now means "the destination page is rendered", which is what every call site already assumed.

## Proof

`pnpm evals:e2e saved-app-creation --daytona` with `OPENWORK_EVAL_REF` exported; placement line `selection: engine=legacy surface=legacy case=all; placement: daytona (--daytona)`; each run provisioned its own Den + admin desktop + web sandboxes and deleted all three.

| run | sha (source verified in sandbox log) | result |
|---|---|---|
| 2 | `3d139f004` (merge + diagnostics only, no fix) | 0 passed / 1 failed / 0 skipped — line 202 `expected '' to contain 'Hide details'` (the line-198 instance of the same race; step 13 not reached) |
| 3 | `fead49b90` (fix + step-13 diagnostics using a raw hash set) | **1 passed / 0 failed / 0 skipped**; diagnostics quoted above; the raw-hash click happened to land after the commit (`1055,190`, dashboard card) so `Z2 reproduced: false` in that run |
| 4 | **`c63458e387be1bb842dad885550813d90cad745b`** (final, diagnostics removed) | **1 passed / 0 failed / 0 skipped**, `"verdict":"passed"`. All 14 steps ok; **step 13 "an admin cancels deletion in the app and confirms it on the dashboard" ok (4.0 s)** — `Delete Team briefing` located at `912,260` on the app page and `470,298` on the dashboard card, then `Delete app`; **step 14 "Dashboard Add opens a creation conversation" ok (1.2 s)**. |

Also: `pnpm --dir evals run check:browser` passes (832 callbacks); `lint:layers` and `tsc -p evals` report only pre-existing violations/errors in files this PR does not touch (same as noted in #4770 / #4779).

## Noticed, not touched

- `locate()`'s not-found message caps "Visible button/link candidates" at the first 8 visible buttons, which are always the shell rail. It reads as "the page has no buttons" and misled #4779's diagnosis; listing candidates from the primary slot first, or raising the cap, would make misses self-describing.
- `inPreview` in `worlds/saved-apps.ts` picks the first sandbox iframe target; on the dashboard there can be several (company tile + personal card), so preview reads there are ambiguous. Not a problem for current call sites once `open` waits for the destination.
- The `go()` behaviour itself (hash set only) is shared by many worlds; a generic "route committed" signal would need a product marker for the active route. Kept this fix local to the saved-apps world.
- Run 1 was killed by my tool's shell timeout during Den provisioning; its leaked `openwork-server-…-32284-…` sandbox was deleted manually. No sandboxes from any of my runs remain.


yomgui:
## Resync onto `dev` after the #4757 conflict resolution

- Merged the **updated** `origin/e2e/saved-app-browser-requesting-conversation` (#4779 @ `1afdc2757`, which now carries AI's #4757 resolution `4a2f43213` + `dev`) → `74b1cfa22`, then `origin/dev` (`e8bc90f9f`) → **`05b752efc`**. Both merges applied cleanly; **no conflicts**. Both merge commits SSH-signed (`G`).
- `pnpm --filter @openwork/app typecheck` → exit 0. This PR's own change is still the single `evals/worlds/saved-apps.ts` `open()` hunk; the squash diff shrinks to that once #4779 → #4757 land.

## Proof at the pushed head

`OPENWORK_EVAL_REF=05b752efcbfb7bf9a14c2802daa8ce7c13d416e0 pnpm evals:e2e saved-app-creation --daytona` (`placement: daytona (--daytona)`, `source verified 05b752efc…` in all three sandboxes; fresh Den + admin desktop + web sandboxes, all deleted by the runner).

| test | result |
|---|---|
| **`create, preview, save and reopen an app without changing already-open results`** (this PR's spec, incl. step 13 "an admin cancels deletion in the app and confirms it on the dashboard") | **✓ passed (387.1 s)** |
| `APP-ISOLATION embedded MCP Apps isolate siblings…` (added to this file by #4829 today) | × failed in world provisioning (62 s), before any assertion: `Mock issuer gate failed: health reported https://3979-…, expected https://3979-…` |

`"passed":1,"failed":1,"skipped":0` → file verdict `Failed`, but the red test is #4829's, not this PR's: `isolatedMcpApps` seeds two `seed.mock(...)` without ports into one `appWeb` sandbox, both default to `daytonaPort: 3979` (`evals/packages/env/src/mock.ts:56`), and the second mock's `pkill` + rebind on the same port trips the issuer gate. Now reproduced 3/3 across #4779 (2 runs) and this branch; #4829 only proved that case with `--local`. See the same note on #4779. This PR does not touch that code.

PR #4638, opened 2026-09-08
Keep model defaults and saved settings consistent

reachjalil:
## Purpose

Keep provider Default as a real `null` choice, preserve saved same-model effort when metadata is missing or stale, and apply explicit effort edits in the full picker and Automation editor.

- Default stays selectable and checked, including forward/backward keyboard cycling.
- Runtime variant IDs remain exact, including custom keys and the distinct named `default` variant. Unknown saved values are explained, not silently rewritten.
- Explicit favorite/model switches carry only target-compatible effort, otherwise Default. Automation model switches reset effort; reselecting the same model preserves it.
- Full-picker callbacks update effort without replacing an unrelated default model. Automation edits preserve instructions and do not submit until Save.

## Independent dev target

Old head: `d3736d1ce759fe2051a5209e507886c17043cb8c`.
Current signed head: `14fffc4df6962562daf2f2521fcd27a9733ece20`.
Fresh base and verified merge base: `858414c41c63a22eecfe331d17f3f5dd348dfc6f`.

Rebased the single settings increment directly onto current dev. **No parent stack is included.** #4629 and its ancestors are not dependencies. Managed-only discovery, upgrade/allowance policy, free access, onboarding/sign-in changes and parent fixtures remain excluded. Provider-access restrictions, billing, budgets, catalogs and provider materialization are unchanged.

## Current-dev reconciliation

- Keep dev's opaque variant normalization and its regression assertions from #4853. Add explicit `null` Default alongside, not instead of, the exact runtime keys.
- Keep the complete native-v2 `MODEL-01` journey, its custom-effort round trip and its unsupported-model negative assertions. Resolve only the import conflict; do not replace its world or restore parent-stack fixtures.
- Fix an integration regression: injecting Default into the options list must not enable compact effort for a model with no advertised variants and no saved setting. Such models retain disabled **Unavailable** effort. A stale saved value still exposes Default so it can be cleared explicitly, without switching the model.
- Add one compact-picker component regression in the existing model-picker test file. It verifies the unavailable state, stale-value recovery to `null`, and no model-selection callback.

No review threads were present when reviewed. The branch commit is authored by Jalil, signed with the existing signer and verified `G`. Publication used an exact expected-head lease on the existing branch.

## Verification: Incomplete

Passed on the final source tree (exit 0), using frozen offline app dependencies:

```sh
[...]

PR #4633, opened 2026-09-08
fix(app): show Starting feedback immediately after Send

reachjalil:
## Summary

Distinguish a pending send from confirmed work without undoing immediate-message rendering, attachment preparation, draft continuity, interruption or admission protections.

- Show **Starting...** for text and attachment submissions while creating a conversation, through the immutable auto-send handoff, and while an idle-session submission is pending.
- Give observed busy/retry status precedence over a pending submission promise. Do not start a Working timer merely because a send is pending.
- Preserve waiting, compaction, delegated-task, visible-tool and reconnecting treatments rather than overlaying Starting.
- Reuse rejection/cancellation cleanup and draft recovery. No transport, queue, attachment-preparation or persistent-state changes.

## Current-dev review

Reviewed existing signed head `640bdf096512501ed7e18c7f51941c1b8ba934b3` against freshly fetched `dev@858414c41c63a22eecfe331d17f3f5dd348dfc6f`. Rebase reports up to date; merge base equals that exact dev SHA. All three branch commits verify with `G` and Jalil's author identity. No new commit or force push was necessary.

Current dev still starts the Working lifecycle for submitted state and prioritizes pending send over observed engine activity. This remains a useful, small increment. The branch preserves dev's progressive message footer, creation busy-button, connected-service preparation, immutable handoff and admission protections. Its fixture aligns live-status and snapshot seeds; no production assertion is weakened. No review threads were present at review time.

The native-session fixture exports and prior scan-limit adjustment are already in dev and are excluded. This PR does not change review policy configuration. The journey's hover re-arming repair remains useful.

## Verification: Incomplete

Passed again in a fresh isolated checkout of the current head, with frozen offline app dependencies (all commands exit 0):

```sh
corepack pnpm --filter @openwork/app exec bun test --isolate ./tests/message-list-loading.test.tsx -t 'message-list loading feedback|message-list reconnecting feedback|keeps delegated tasks'
# 17 passed, 0 failed, 7 filtered out; 76 assertions
corepack pnpm --filter @openwork/app exec bun test --isolate ./tests/composer-focus-continuity.test.tsx
# 1 passed, 0 failed; 476 assertions
git diff --check origin/dev...HEAD
```

These are supporting component checks, not real-engine runtime evidence. React duplicate-key warnings were emitted by the composer fixture; they did not fail its assertions and remain unresolved by this PR.

- Failed: no selected check failed in this review.
- Skipped/not run: the seven unrelated filtered component cases and broad suites.
[...]

PR #4579, opened 2026-09-06
fix(app): surface delegated questions in the parent conversation

reachjalil:
## Summary

Restore the ability to answer a delegated session's question from its parent conversation, so the child and the waiting parent can continue.

- Use the parent's existing descendant scope for questions as well as permissions. Unrelated roots remain isolated.
- Retain incoming child questions even before their transcripts are tracked.
- Reply using the original request ID and clear the actual owning child's queue and waiting state only after success.
- Preserve live questions during snapshot refresh and prevent a late list or replay from resurrecting a settled request.
- Extend the existing parent/child approval journey with real v2 delegation, native question forms, reload, original-request settlement, both sessions completing, and an unrelated pending-question control. No new test files or engine changes.

## What regressed

The working cross-session question handoff was introduced in #705. The React cutover in #1470 removed that store path. #1867 restored inline questions but subscribed only to the selected session's question cache. The descendant aggregation later restored for permissions in #4270 never reached questions.

There was a separate recent v2 transport omission in #4388: question listing was empty and native form events/replies were not bridged. #4550 fixed that transport path, but left the parent-only selector intact. A native child form could therefore be pending while the parent displayed no answer control. The parent waits for its child; the child waits for the unseen answer.

These findings are from source/history review and an isolated real-engine journey, not an inspection of any particular user's private session.

## Verification — Incomplete overall; targeted v2 handoff Passed

Head: `c6fe518c14a069e0f704cece3a48903b243c60e7`
Base: `89933074f80ad68fbe3f0fad725ab1f9ff5f19a4`

Passed on the final head:

- `OPENWORK_EVAL_REF=c6fe518c14a069e0f704cece3a48903b243c60e7 OPENWORK_EVAL_ENGINE=v2 pnpm evals:e2e parent-child-permission-approval` — exit 0; **2 passed, 0 failed, 0 skipped**. `placement: daytona (daytona CLI authenticated)`. Fresh desktop and Den resources checked out the exact head and were deleted after the run.
- `bun test --isolate apps/app/tests/session-sync-permissions.test.ts` — exit 0; **19 passed, 0 failed**. Covers early child events, owner isolation, rejection, stale snapshots, and waiting-state preservation.
- `pnpm --filter @openwork/app typecheck` — exit 0.
- `git diff --check origin/dev...HEAD` — passed.

[...]

PR #4798, opened 2026-09-10
fix(app): apply per-session bash 'ask' permission and withdraw orphaned permission requests after Stop

yomgui:
Makes test 2 of `parent-child-permission-approval` ("retry recovery and stopping a permission leave other requests and fresh work intact") green on Daytona; tests 1, 3, 4 stay green. Follow-up to #4786 (its "Noticed, not touched" section names both points fixed here). Cluster context: [2026-09-09 E2E triage](file:///Users/guillaume/OpenWork%20Chat/reports/e2e-red-2026-09-09/triage.md) — `parent-child-permission-approval` test 2 (listed under Cluster 1, but the line-137/138 failure is not the send gate).

**Stacked** on `origin/e2e/child-checklist-after-stop` (#4786, which carries #4723 and #4757) and on `origin/e2e/workspace-permission-reload` (#4747), both via merge commits; none of their files are modified here. #4747 is needed because a workspace created at its declared tmp path only exists in the server registry, and without its `desktop-runtime-boot` fix the spec's `user.reload()` would restart the embedded server/engine and drop every pending request.

## Root cause

**(a) spec line 137 — `bash` ran without asking although the world wrote `permission: { bash: "ask" }`.** Not a product bug in permission handling: it is the ancestor-`.opencode` shadowing trap from #4747. `splitPaneQuestions` called `seed.workspace()` without `create`, so it adopted the first-launch default workspace, which the dev profile places inside the repo checkout on Daytona. A throwaway diagnostic spec run at `0914332ff` showed:

- `GET /workspace/:id/permissions/effective` → `shell: allow`, `source: "engine"`, `files.workspace = /workspace/.openwork-daytona/profiles/…/openwork-dev-data/home/OpenWork Chat/opencode.jsonc` (not the `/tmp/openwork-permission-stop-recovery-…` path the world declared);
- engine `/config` → `permission: {"bash":"ask","*":"allow"}` — the `*: allow` comes from `/workspace/.opencode/opencode.json` (`"permission": "allow"`), merged **after** the workspace file;
- `openwork` agent ruleset → `[* allow, bash ask, * allow]`; last match wins, so bash was allowed and the tool part completed with no `/permission` entry.

In isolation (`opencode serve` 1.18.25, `/tmp` workspace with `{"permission":{"bash":"ask"}}`, OpenWork-style `OPENCODE_CONFIG`) the engine yields `bash * ask` for the same agent — so the engine and the app apply the rule; only the arrangement pointed at the wrong workspace.

**(b) spec line 150 — after Stop, the aborted tool's permission stayed in `/permission`.** Same OpenCode v1 engine gap #4786 fixed for `question`: tools run on detached fibers, so `session/abort` marks the tool part aborted but leaves `Permission.ask` pending; no event withdraws it. The spec prose ("Stop removes its cancelled approval without a permission reply") requires the app to withdraw it.

## What changed

- `evals/worlds/chat.ts` — `splitPaneQuestions` takes an opt-in `createWorkspace`; `permissionStopRecovery` creates its workspace at the declared tmp path (mirrors #4747's `emptySession` fix). Opt-in so the four other `splitPaneQuestions` worlds keep their arrangement (see below). No assertion changed.
- `apps/app/src/app/lib/opencode-interruption.ts` — #4786's post-idle withdrawal helper (now `withdrawRequests`) also lists pending permissions and rejects the ones owned by the stopped tree (root + interrupted descendants). Other roots are never touched; each kind is best-effort on its own so a failed question withdrawal does not skip permissions. Works through the shared `Client` on both engines (`permission.list` / `permission.reply`, the v2 adapter maps them per session).
- `apps/app/tests/opencode-session-native.test.ts` — the tree-stop test now also serves a `/permission` list with a stopped-tree entry and an unrelated root's entry and asserts only the stopped one receives `reply: "reject"`, after idle confirmation.

## Proof

`pnpm evals:e2e parent-child-permission-approval --daytona` with `OPENWORK_EVAL_REF=c2efbf936f24e557af5968783d89fcf0015fb2b8` exported (source verified in every sandbox: `==> source verified c2efbf936f…`), fresh sandbox per surface, placement line:

```
{"command":"evals:e2e","lane":"e2e","daytona":true,"placement":"daytona","engine":"legacy","surface":"legacy","files":["parent-child-permission-approval"],"passed":4,"failed":0,"skipped":0,"skips":[],"verdict":"passed"}
```

| Test | Result at `c2efbf936` |
| --- | --- |
| 1 — switching and creating threads hydrate permissions without waiting for unrelated roots | passed (125 s) |
| 3 — a parent task surfaces and resolves its child session permission request | passed (141 s) |
| **2 — retry recovery and stopping a permission leave other requests and fresh work intact** | **passed (285 s)** — all four steps, including line 137 `see(printf OTHER_PERMISSION_WITNESS)`, line 143/144 both approvals pending, line 150 `native interruption cleanup removes the pending permission`, no `final` mock call for the stopped prompt, reload keeps the unrelated approval, and the follow-up completes once |
| 4 — a parent answers and stops real child questions, then finishes fresh work without settling an unrelated root question | passed (318 s) |

Overall verdict for this spec file: **Passed** (4/4, 0 skipped).

Baseline on this branch before the fixes (`0914332ff`, #4786 stack only): the diagnostic run reproduced (a) as described above (`sawPermission: false`, tool part `completed`).

Unit: `bun test --isolate tests/opencode-session-native.test.ts` → 27 pass, 0 fail. `pnpm --filter @openwork/app typecheck` clean. All Daytona sandboxes from these runs were deleted.

## Noticed, not touched

- `splitPaneQuestions` consumers `delegatedQuestionHandoff`, `restartUpdateTaskWorld` (`background-desktop-update`), `newSplitPrimary`, `sessionSubmitErrorIsolation` still adopt the default in-repo workspace, so their `policy` blocks are shadowed by the repo's `.opencode/opencode.json` too. They happen to write policies the repo config already implies (`question: allow`, `bash: allow`), so their claims hold, but any future workspace-level `ask`/`deny` there will silently not apply. Flipping `createWorkspace` on for all of them is a one-line change once each spec is re-proven.
- #4747's "Noticed" item stands: `permissions/effective` attributes ancestor `.opencode/opencode.json` rules to `engine`, and a user whose workspace sits inside such a repo sees "Ask before actions" while the engine allows everything.
- `pnpm --dir evals typecheck` reports 389 pre-existing errors unrelated to these files (`Promise.withResolvers` lib target, `packages/email` jsx, app import extensions).


yomgui:
## Merged `origin/dev` — `8eb2d35be` (SSH-signed, `%G?` = `G`)

`dev` moved and conflicted in `evals/worlds/chat.ts` on the `splitPaneQuestions` signature. Resolved with a merge commit (no rebase/amend/force-push), keeping both intents — the same resolution already applied on the stacked #4810 (`b73422894`), so the two branches now carry byte-identical `splitPaneQuestions`:

- **dev** added a 5th parameter `surface: AppSurface = "electron"` plus the `appWeb` branch (`workspacePath` realpath'd for web, `agentMock` from either fixture). Kept verbatim.
- **this PR** added `options: { createWorkspace?: boolean }` so `permissionStopRecovery` creates its workspace at the declared tmp path instead of adopting the in-repo default. Carried as a **6th** parameter; the shared call is `seed.workspace(app, workspacePath, options.createWorkspace ? { create: true } : {})` with the explanatory comment, and `permissionStopRecovery` passes `"electron", { createWorkspace: true }`. Other callers unchanged (defaults).

Checks after the merge: `pnpm --dir evals run check:browser` → 843 callbacks clean; `tsc -p evals` shows only the three pre-existing `Promise.withResolvers` lib-target errors in `worlds/chat.ts` (also on `origin/dev`); `bun test --isolate tests/opencode-session-native.test.ts` (this PR's app unit test, `session-route.tsx` auto-merged nearby) → 27 pass, 0 fail.

**Proof at `8eb2d35be65556086ef9949e8d1f347e3c7980df`** — `pnpm evals:e2e parent-child-permission-approval --daytona` with `OPENWORK_EVAL_REF` exported (`==> source verified 8eb2d35be6…` in every sandbox), fresh sandbox per surface:

```
selection: engine=legacy surface=legacy case=all; placement: daytona (--daytona)
✓ switching and creating threads hydrate permissions without waiting for unrelated roots  124045ms
✓ a parent task surfaces and resolves its child session permission request  128929ms
✓ retry recovery and stopping a permission leave other requests and fresh work intact  303973ms
✓ a parent answers and stops real child questions, then finishes fresh work without settling an unrelated root question  289239ms
"passed":4,"failed":0,"skipped":0,"skips":[],"verdict":"passed"
```

All six sandboxes from the run were deleted by the runner. Verdict for this file: **Passed** (4/4, 0 skipped).

yomgui:
## Resync onto `dev` after the #4757 / #4786 conflict resolution — and the #4747 revert

**Merges (all SSH-signed `G`, pushed without force):**
1. `origin/e2e/child-checklist-after-stop` (#4786 @ `8763acc6d`, AJ's resync carrying #4757 `4a2f43213` + `dev`) → `df94b47ba`. One textual conflict in `apps/app/tests/opencode-session-native.test.ts` (tree-stop test's allowed request URLs: this PR adds `/permission`, upstream adds `/path?`) — kept both.
2. `origin/e2e/fresh-session-send-gate` → already up to date. `origin/dev` (`e8bc90f9f`) → `e68f9510a`.
3. After #4870 landed: `origin/dev` (`858414c41`) → **`614ebb795`** (current head).

**#4747 product commit `561b5c03a` (`reuse the running OpenWork server when the selected workspace is server-owned`) is gone from this branch.** Dev reverted #4747 in #4870 because bisect showed it causes the `active-session-workspace-storm` regression (stale per-workspace provider config after a switch). This branch carried `561b5c03a` via #4747's original commits; merging `origin/dev` with the revert inverted that hunk (`git revert --no-edit 561b5c03a` afterwards was a no-op: "nothing to commit"), and `git diff origin/dev...HEAD -- apps/app/src/react-app/shell/desktop-runtime-boot.ts` is **empty**. #4747's two evals-only commits (`de6b73890`, `a740eb41b`) remain. The product change is not re-added here.

**Composition with #4837 (`interrupt delegated work before immediate follow-ups`)**: `submitImmediateSessionTurn` → `interruptSessionTurn` → `stopForegroundTree` → authoritative idle → `withdrawRequests` (questions **and** permissions of the stopped tree, each best-effort). #4837's `foregroundDelegations` refactor and this PR's helper touch disjoint parts of `opencode-interruption.ts`; the merged `test.each([false, true])` tree-stop test asserts `withdrawn = ["que_nested", "per_nested:reject"]` for both the Stop and the immediate-follow-up path. `bun test --isolate tests/opencode-session-native.test.ts` → **42 pass / 0 fail**; `pnpm --filter @openwork/app typecheck` → exit 0.

## Proof at `614ebb795` (`OPENWORK_EVAL_REF` pinned, `placement: daytona`, `source verified 614ebb795…`)

`pnpm evals:e2e parent-child-permission-approval --daytona` → **3 passed / 1 failed / 0 skipped**:

| test | result |
|---|---|
| 1 switching and creating threads hydrate permissions… | ✓ 122 s |
| 3 a parent task surfaces and resolves its child session permission request | ✓ 130 s |
| 4 a parent answers and stops real child questions, then finishes fresh work… (delegated-question handoff + Stop, with #4837 present) | ✓ 292 s |
| **2 retry recovery and stopping a permission leave other requests and fresh work intact** | × — steps 1–2 ok; step 3 fails at line 160 |

Inside test 2, **this PR's two claims still hold**: line 137 (`bash` asked via the workspace's own `permission: { bash: "ask" }` — the `createWorkspace` fix) passed in step 2, and line 150 `native interruption cleanup removes the pending permission` passed in step 3 (Stop withdrew the aborted tool's permission without a permission reply; `notSee("Allow once")`, `notSee(Stop)` ok). The step then fails **right after `user.reload()`**: `pending(world.other.sessionId)` is `[]` instead of the unrelated approval — the reload restarted the embedded server/engine and dropped every pending request. That is precisely what `561b5c03a` prevented (this PR's body: "without its `desktop-runtime-boot` fix the spec's `user.reload()` would restart the embedded server/engine and drop every pending request"). So **the spec's reload-persistence claim depended on the reverted #4747 product change**; with #4870 on dev it is unproven on dev too, not something this PR should re-add. Owner call: either the spec's reload step waits for the composer-run-mode replacement fix, or the claim is re-expressed.

Earlier run at `e68f9510a` (before the revert landed; `561b5c03a` still present): 3/4 with test 2 passing steps 1–3 (including the post-reload check) and hanging in step 4 after `open(world.other)` — `see("Allow once")` never returned for **745 s**, screen dump `unavailable` (renderer main thread unresponsive after switching to the conversation with the pending approval). Per the audit note: **attribute to #4837 first** (Stop/interrupt path in `session-surface.tsx`), then #4838/#4844/#4846 (history/scroll). Not reproduced at `614ebb795` because step 3 now fails before it; noting it so it is not lost.

`pnpm evals:e2e task-activity-shimmer --daytona` (delegated-task journey) at `e68f9510a` → died in world provisioning: `launch remote app-web runtime … EACCES: permission denied, mkdir '/private'`. `taskActivityWeb` (rewritten by #4812/#4867, proven `--local` only) passes the host's macOS `seed.tmpPath(...)` as the `appWeb` `workspacePath`, and `startRemoteRuntime` mkdirs it inside the Linux sandbox. Deterministic harness defect on dev for every `appWeb` world under Daytona placement (the `APP-ISOLATION` case in `saved-app-creation` has the same shape); not this PR's. Not re-run.

All Daytona sandboxes from these runs were deleted by the runner. The squash diff of this PR shrinks to its own three commits once #4786 → #4757 land.

yomgui:
## Resync after #4786 squash-merged (`a7f0a806f`)

Merged `origin/dev` twice as dev kept moving: `a7f0a806f` → `b52b011e2`, then `f6ab5f271` → **`0912f4b5d`** (current head; both SSH-signed `G`, no rebase/force/stash).

**Conflicts (2 files) — resolved by keeping this PR's hunks**, since it is a superset of #4786's squash:
- `apps/app/src/app/lib/opencode-interruption.ts`: dev's `withdrawQuestions` (the #4786 squash) vs this PR's `withdrawRequests` (questions **+** permissions). Kept `withdrawRequests`; `withdrawQuestions` no longer appears anywhere (no duplication).
- `apps/app/tests/opencode-session-native.test.ts`: the tree-stop test's `/permission` list + `/permission/:id/reply` handler, `withdrawn = ["que_nested", "per_nested:reject"]`, and `/permission` in the allowed-URL set. Kept all three; #4837's `test.each([false,true])` immediate variants and the dev-side `/path?` allowance are intact.

**Verification** — `git diff origin/dev` (`f6ab5f271`) is exactly this PR's own extension and nothing else:
```
apps/app/src/app/lib/opencode-interruption.ts  | 17 +++++++++++++----   (withdrawQuestions → withdrawRequests + permission withdrawal)
apps/app/tests/opencode-session-native.test.ts | 19 ++++++++++++++++++-  (the three test hunks above)
evals/worlds/chat.ts                           | 12 ++++++++++--      (createWorkspace opt-in for permissionStopRecovery)
```
`apps/app/src/react-app/shell/desktop-runtime-boot.ts`: **no diff vs dev** — the #4747 (`561b5c03a`) in-branch revert survived. `pnpm --filter @openwork/app typecheck` → exit 0; `bun test --isolate tests/opencode-session-native.test.ts` → **42 pass / 0 fail**. PR is MERGEABLE.

## Proof at `0912f4b5d`
`OPENWORK_EVAL_REF=0912f4b5db02a8dc572c2e364d325a9a0e5a3ee4 pnpm evals:e2e parent-child-permission-approval --daytona` (`placement: daytona`, `source verified 0912f4b5d…` in all four sandboxes, all deleted by the runner) → **2 passed / 2 failed / 0 skipped**:

| test | result |
|---|---|
| 1 switching and creating threads hydrate permissions… | ✓ 137 s |
| 3 a parent task surfaces and resolves its child session permission request | ✓ 131 s |
| 2 retry recovery and stopping a permission… | × — **same as at `614ebb795`**: steps 1–2 ok, line 137 (`bash` asked) and line 150 (Stop withdraws the aborted tool's permission without a reply) pass; fails right after `user.reload()` because the reload restarts the embedded engine and drops the unrelated pending approval — the claim that depended on the reverted #4747 product commit. Not re-added. |
| 4 a parent answers and stops real child questions… | × **new since `614ebb795`** (where it passed 7/7 steps): step 3 "reloading the parent restores the same unanswered requests" — after `reload()` the child row shows `Choose delegated task format · General agent — Waiting for task result` but never its question (`Which format should the child task use?`, 45 s). This is before any Stop, so this PR's helper is not on the path. |

Attribution for test 4: the only product changes between `858414c41` (green) and `f6ab5f271` (red) are **#4875** (revert of #4858 — removes the visible-session-refs / owner-checked release logic in `session-sync.ts` and deletes `session-sync-permissions.test.ts`), #4633 (`message-list.tsx` "Starting…" status), #4862 (`mcp-app-frame.tsx`), #4638 (model settings). **Attribute to #4875 first**, then #4633. Test 4 exists on dev itself (#4579), so a clean control at `f6ab5f271` will settle it; not run here (another ~25 min sandbox cycle) — flagging instead.

PR #4808, opened 2026-09-10
test(evals): prove an admitted signed-in desktop turn runs its next bash tool while Den returns 503

yomgui:
## What

One Daytona-only `@openwork/testkit` E2E spec, `evals/specs/desktop-policy-unreachable.e2e.test.ts`, that proves the behaviour #4725 (desktop policy-engine rollback) was meant to guarantee:

> an admitted signed-in desktop turn executes its next bash tool while Den returns 503, without a managed-policy plugin

No product code changes.

## Why

The v0.18.45 stability soak never combined a **signed-in** desktop's **real agent tool execution** with an **unreachable Den/policy service at the tool boundary** — its tool journeys always had healthy Den transport and the restart fixture was signed out. So prior green runs could not tell whether the removed per-tool managed-policy hooks could still block tools. This spec closes that gap.

## How the journey works

1. Isolated Den (Daytona) with an org, admin, and member; a deterministic mock model (`mcpMock` agent workload) reachable independently of the Den path.
2. A fault proxy fronts Den for the desktop. Den Web's `/api/runtime-config` is rewritten to advertise the **proxy** API origin so the desktop cannot route around the fault (run 2 found this bypass).
3. The signed-in member's desktop admits a turn while Den is healthy. Its first real `bash` tool writes a baseline marker and waits on a fixture gate file.
4. The proxy then returns **503 for every `/api/den` request**. Calibration on the real desktop-server→Den path: a *separate* new session admission is rejected `403 policy_unavailable` (fail-closed admission is preserved, as #4725 documents).
5. The fixture releases the gate. The **next real `bash` tool completes and writes its result file** while Den still returns 503; an unrelated file and the signed-in identity are unchanged; the native session inventory is unchanged; the final reply is visible in the desktop.
6. The engine's runtime plugin registration list is read before and during the outage: known OpenWork plugins present, **no `managed-policy` / `managed-policy-next` registration**.

## Proof (Daytona lane, OpenCode v1)

| Command | Exit | P/F/S | Notes |
| --- | ---: | --- | --- |
| `OPENWORK_EVAL_REF=e4c5c3e3d49a75287c020111de989a93dbf30281 gtimeout -k 60 2400 pnpm evals:e2e desktop-policy-unreachable --daytona --engine v1` | 0 | **1/0/0** | 281 s; evidence `evals/results/test-runs/2026-09-10T16-05-21-402Z-…`; both sandboxes deleted by cleanup |
| `pnpm --dir evals run check:browser` | 0 | N/A | 838 browser callbacks, no type errors |
| `pnpm --dir evals exec tsc -p . --noEmit` (filtered to this spec) | — | N/A | no diagnostics in this spec; the pre-existing failures are unbuilt workspace deps in `apps/app`/`ee` |

Verdict: **Passed** for the claim in the test title.

### Evidence SHA note (deliberate, disclosed)
The green run's receipt is bound to `e4c5c3e3` (fetched `origin/dev` at run time) because the spec was still untracked when it ran. The PR head `716553e5b` differs from `e4c5c3e3` **only by adding this spec file**, whose content was byte-identical during the run (`git diff --stat e4c5c3e3 716553e5b` → 1 file, the spec). The sticky comment is therefore published with `--force` and carries the publisher's stale-SHA annotation. Reviewers who want a head-bound receipt can re-run the command above on this branch (≈5 min).

### Red runs on the way (all fixture-side; product behaved as documented)
- Run 1: fixture directory not created before the unrelated-file witness (ENOENT).
- Run 2: forwarded runtime-config advertised Den's direct API, so the desktop bypassed the fault proxy (admission returned 200 instead of 403) — fixed by rewriting `/api/runtime-config`.
- Run 3: null counts; the sandbox `source dependency install gate` took 1018 s (vs 26–29 s on the neighbouring runs of the same commit) and the launching session's outer budget ended the run. Classified as an environment slowness, not a harness defect. Its two leaked sandboxes were deleted manually.
- Run 4: 403 reached, but the probe read the denial code from a non-existent `error` field; the desktop server formats `ApiError` as `{ code, message, details }`. Fixed and tightened to `toBe("policy_unavailable")`.

## Not included
- **Negative control** (an old build / re-registered managed-policy plugin blocking the tool) — not cheap: needs a pre-#4725 build or plugin re-registration plus another Daytona run. The spec proves the positive claim only.
- The spec is a legacy-style `test()` spec importing `@openwork/behaviors` / `@openwork/hosts` like nine existing sibling E2E specs; it adds two `specs-use-testkit-only` entries to `lint:layers` output (the sandbox file witness needs `defaultDaytonaExec`, which testkit does not re-export). Happy to migrate to `spec.world()` in a follow-up if preferred.


yomgui:

Commit `cfb0a9d07cf594d3d483444a1ee7bb9ef2f942ca` · selected evidence

[Open review report](https://openwork-review-jqvsx0grh-prologe.vercel.app/r/e9c86ae0504b4aaa9fe87bfe00709346)

PR #4840, opened 2026-09-10
fix(app): clear stale waiting indicators when an aborted question never reports rejection

yomgui:
## Symptom

A session that finished hours ago still shows the orange "needs you" dot in the sidebar, and `session.archive` is refused ("could not be confirmed" → Stop-and-archive confirmation → "A task, approval, or message acceptance is still unresolved"). Found by the workspace audit on an "E2E fix" session: 12:02 the assistant calls `question`; 12:05 a superseding prompt (another agent's STOP) arrives while the question is pending; the question tool part ends with `state.status = "error"`; the follow-up turn finishes with `finish=stop`; engine idle for hours; app still says `waiting`.

## RCA (confirmed, with one correction)

**Engine (confirmed, not patchable here).** The engine is the upstream OpenCode binary (`constants.json` → `v1.18.18`, downloaded by `apps/desktop/scripts/prepare-sidecar.mjs`). In `packages/opencode/src/question/index.ts` (and identically `permission/index.ts`), `ask()` wraps `Deferred.await` in `Effect.ensuring(pending.delete(id))`; only an explicit `reply()`/`reject()` publishes `question.replied`/`question.rejected`. `SessionProcessor.cleanup()` waits 250 ms for open tool calls and then marks them `error: "Tool execution aborted"` **without interrupting the tool fiber**, so the question's deferred is never failed either. Read-only probe of the live engine while the affected session was idle:

- `GET /session/status` → the session is absent (idle)
- `GET /question` → **still lists** the session's request `que_…` with `tool: { messageID, callID }`
- DB: that `callID`'s tool part is `state.status = "error"`

So the RCA hypothesis 1 is confirmed *and stronger than stated*: not only is `question.rejected` never published, the abandoned request stays in the engine's pending list as a zombie. That rules out option (a) ("drop ids missing from the engine list") as a fix — the list is the source of the ghost.

**App (corrected).** Hypothesis 2 (a missed SSE idle plus the snapshot guards) is not what keeps the record waiting. `setRunStatus(idle)` does clear the ids, and `connect-reconcile` (`seedSessionRun` with `snapshotStartedAt`) heals a missed idle. The `record.status !== "idle"` guards in `seedWorkspaceSessions`/`seedSessionRun` are intentional: `waiting` + `runActive=false` is a legitimate state ("retains a pending question at the idle release deadline" test, v2 permissions), so they stay. What actually re-poisons the store is the zombie itself:

- `useSessionInteractions` reads `GET /question` whenever the session (or its parent) is open and `seedQuestionState` re-adds the zombie → `waiting` again with `runActive=false`; nothing ever settles it.
- `useSessionArchive.readWorking` reads the same list and counts the zombie as open work → archive refused forever.

## Fix (option (c): the transcript is authoritative)

A question/permission can only be answered while the tool call that asked it is still running. When that call's part is `completed`/`error`, nobody can consume an answer, so the request is settled:

- `sync/orphaned-interactions.ts` (new): `isTerminalToolPart`, `terminalToolCallIds` (raw parts), `terminalTranscriptToolCallIds` (UI transcript), `isOrphanedInteraction(tool, terminal)`.
- `session-sync.ts`: `settleOrphanedInteractions()` runs on live `message.part.updated` (before the tracked-session gate, so background sidebar rows heal too) and after each transcript snapshot; `seedQuestionState`/`seedPermissionState` treat orphaned requests as settled so a zombie list read cannot resurrect them. Settling goes through the existing `settle*State` path (settled set + waiting id removal), so the existing snapshot/live ordering guards are untouched.
- `use-session-archive.tsx`: `readWorking` ignores a pending request whose tool call already ended in the fetched messages.

Requests without a `tool` link and requests whose tool part is still `running` are never touched (covered by tests).

## Proof

- `apps/app`: `bun test --isolate tests/session-sync-orphaned-interactions.test.ts tests/session-sync-permissions.test.ts tests/session-sync-run-status.test.ts tests/session-sync-tool-parts.test.ts tests/session-sync-lifecycle.test.ts` → 107 pass / 0 fail (the 6 new tests were red before the fix: 5 fail / 1 pass; the existing "does not let an older idle snapshot stop a live run" guard stays green).
- `apps/app`: `pnpm typecheck` → clean.
- E2E: `evals/specs/abandoned-question-waiting-indicator.e2e.test.ts` (new world `abandonedQuestion` in `evals/worlds/chat.ts`): real question via mock model → `POST /session/:id/abort` + `prompt_async` follow-up (what another agent's STOP does) → asserts the engine tool part ended and the session is idle, the question panel closes, the sidebar dot stops saying "Waiting" within 15 s, and Archive succeeds with no Stop-and-archive confirmation. Lane and result: see the evidence comment below.

## Upstream engine item (not patchable in this repo)

Proposed issue for anomalyco/opencode:

> **Aborted/superseded question or permission is never rejected and stays in the pending list**
>
> When a session turn is aborted (`POST /session/:id/abort`) or superseded while a `question` (or permission) tool call is awaiting an answer, `SessionProcessor.cleanup()` marks the tool part `error: "Tool execution aborted"` but does not interrupt the tool fiber. `Question.ask`'s `Effect.ensuring(pending.delete(id))` therefore never runs: `GET /question` keeps listing the request after the session is idle, and no `question.rejected` event is published (only explicit `reject()` publishes one; the instance finalizer also fails deferreds silently). Clients cannot tell a live question from this leftover without cross-checking the tool part state. Same shape in `Permission.ask`. Suggested fix: on abort/supersede, interrupt or `reject()` the pending asks linked to the aborted tool calls so `question.rejected` / `permission.replied(reject)` is published and the pending map is cleaned. Observed on v1.18.18.

## Coordination

Sibling branch `fix/session-blocker-introspection` exposes pending question/permission state to agents. This PR does not touch `session-activity-store.ts`; it adds one helper module and small hooks in `session-sync.ts` (`settleOrphanedInteractions`, seed filters, `message.part.updated`, `seedSessionState`) and `use-session-archive.tsx`. If that branch reads the engine's pending lists, it should apply the same `isOrphanedInteraction` filter so agents do not see the zombie either.

### DIFF
--- a/apps/app/src/react-app/domains/session/sidebar/use-session-archive.tsx
+++ b/apps/app/src/react-app/domains/session/sidebar/use-session-archive.tsx
@@ -24,6 +24,7 @@ import { getComposerQueuedDrafts, useComposerStateStore } from "../surface/compo
 import { composerAutoSendScopeKey, consumeComposerAutoSend, hasComposerAutoSend } from "../surface/composer-auto-send";
 import { dispatchQueuedDrain, getQueuedDrainState, hasPendingQueuedAdmission } from "../surface/queued-drain-machine";
 import { clearQueuedSendContext } from "../sync/queued-send-context";
+import { isOrphanedInteraction, terminalToolCallIds } from "../sync/orphaned-interactions";
 import { applySessionArchived } from "../sync/session-sync";
 
 type ArchiveTarget = { workspace: RouteWorkspace; endpoint: ResolvedWorkspaceEndpoint; sessionId: string; title: string; draftScope: string | null };
@@ -188,8 +189,12 @@ export function useSessionArchive(input: {
           }
           if (idle) dispatchQueuedDrain(id, { type: "idle_reconciled", observedAt, terminalObserved });
           else dispatchQueuedDrain(id, { type: "busy_observed" });
-          return !idle || permissionV2 || permissions.some(request => request.sessionID === id)
-            || questions.some(request => request.sessionID === id) || localWork(id)
+          // A request whose tool call already ended was abandoned by the engine
+          // without a rejection; nobody can answer it, so it is not open work.
+          const terminal = terminalToolCallIds(messages);
+          const unanswered = (request: { sessionID: string; tool?: { messageID: string; callID: string } }) =>
+            request.sessionID === id && !isOrphanedInteraction(request.tool, terminal);
+          return !idle || permissionV2 || permissions.some(unanswered) || questions.some(unanswered) || localWork(id)
             || sessionHasPendingSubmission(baseUrl, id, messages) || sessionNeedsStop(baseUrl, id)
             || hasPendingQueuedAdmission(getQueuedDrainState(id))
             || [workspace.id, endpoint.workspaceId].some(workspaceId =>
--- a/apps/app/src/react-app/domains/session/sync/orphaned-interactions.ts
+++ b/apps/app/src/react-app/domains/session/sync/orphaned-interactions.ts
@@ -1,0 +1,37 @@
+import { isToolUIPart, type UIMessage } from "ai";
+import type { Part } from "@opencode-ai/sdk/v2/client";
+
+// OpenCode asks questions and permissions from inside a tool call. When that
+// turn is aborted or superseded by a newer prompt, the engine marks the tool
+// part terminal and drops the request without publishing question.rejected /
+// permission.replied; an abandoned (not interrupted) ask even stays listed by
+// GET /question. A request whose tool call already ended can no longer be
+// answered by anyone, so the transcript is the authority that settles it.
+
+type ToolLink = { messageID: string; callID: string };
+
+export function isTerminalToolPart(part: Part): part is Extract<Part, { type: "tool" }> {
+  return part.type === "tool" && (part.state.status === "completed" || part.state.status === "error");
+}
+
+export function terminalToolCallIds(messages: ReadonlyArray<{ parts: ReadonlyArray<Part> }>): Set<string> {
+  const ids = new Set<string>();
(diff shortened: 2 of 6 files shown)


yomgui:

Commit `5e7b16f491e7a25e26dd578bfa1dac424f76f95b` · selected evidence

[Open review report](https://openwork-review-jqvsx0grh-prologe.vercel.app/r/14d059cb35d843238eb89c3ea54c2007)

yomgui:
Coordinated with #4843: session attention reads and both question-proposal load/confirmation now cross-check a fresh owner transcript with the same isOrphanedInteraction/terminalToolCallIds helper. The helper is included at the same sync/orphaned-interactions.ts path and is byte-identical to 5e7b16f49 (blob 60a55c648a697ca6155a9fc8f6e14fcb2ba3aa90). No changes to session-sync.ts, session-activity-store.ts, or use-session-archive.tsx, and no changes to this PR's branch. Running/unlinked requests stay eligible; transcript failures fail closed. 32 focused cases pass; a new real v1 abort/supersede zombie E2E is being verified on the pushed #4843 head. #4843 remains draft for its separate native-v2 host-tool exposure blocker.

yomgui:
Follow-up proof on #4843 head 3fe27d100: the real v1 zombie E2E passed on Daytona. It requires the unchanged request to remain in GET /question after abort/supersede, verifies fresh session.attention filters it, then confirms an already-staged answer and observes rejected/sent=false. Normal v1 reply E2E also passed. Shared helper remains byte-identical to your 5e7b16f49; no seed/store guard changes from this feature. Current-head CI passed and Warden approved; #4843 stays draft for the separate native-v2 host-tool integration gap.

PR #4898, opened 2026-09-11
refactor(den-db): consolidate pending Gateway migrations with Drizzle

OmarMcAdam:
## Summary

Replace unapplied 0097–0099 with one **official Drizzle-generated** final Gateway schema migration from unchanged 0096 to current code. The operator confirms the eight intermediate inference tables contain no data, so conversion backfills are not required for this recovery.

- Raw SQL and snapshot produced by lockfile-pinned `drizzle-kit generate`, using its actual table/column rename prompts—not handwritten replacement DDL or a mutated baseline.
- Eight table renames, seven column renames, four final new tables; final model-universe and creator columns are included directly. No table or column drops, and no data backfills.
- All 186 historical artifacts and journal entries through 0096 remain unchanged. The journal ends at consolidated 0097 with its existing index/timestamp.
- Remove obsolete 0098/0099 files/snapshots and historical custom metadata helpers. Update local startup guards to refuse nonempty source tables and mismatched superseded receipts instead of silently restamping them.

## Why draft

This delivers the requested generated-schema consolidation. It is **not yet a production migration-runner fix**.

The existing production ORM migrator wraps DDL in a transaction; the reported Vitess error demonstrates that this execution path must be fixed separately. Drizzle also emits eight primary-key drop/re-add pairs that need target-engine validation. Those generated bytes were not hand-edited.

Do not merge until the execution path is resolved: migration changes reaching `dev` can trigger production application and scheduled retries. Do not apply this replacement to environments with old or partially applied 0097–0099. No production changes, migration applications, journal writes, or PlanetScale queries were performed for this PR. The empty-table premise is operator-confirmed, not a new database measurement.

## Offline verification

- 5 generated-artifact checks passed.
- 26 selected local-runner safety tests with synthetic executors passed.
- 7 selected parser/ownership fixture tests passed.
- Artifact check validated 97 ordered migrations and 41 foundation statements.
- Final generated snapshot matches current code; official Drizzle diff produces zero further statements.
- Verified raw generated bytes, unchanged history through 0096, and `git diff --check`.

No database replay, full application suite, or target-engine primary-key/DDL verification was run. Detailed generation provenance and rename choices are in `ee/packages/den-db/drizzle/0097_gateway_access_matrix.md`.


OmarMcAdam:
Update: the consolidated generated migration has now passed the normal db:migrate command against the operator-designated fresh PlanetScale/Vitess branch at this exact PR head (d99d7923890757b52e96adf9bbb872b1337846fe). This supersedes the earlier unverified transaction and primary-key compatibility concerns for this tested execution path. No custom migrator, per-statement workaround, journal stamping, or source-SQL edits were used.

Before: verified 0096 schema, empty application tables, and all 96 receipts exactly matching the supplied production export. Command: pnpm --dir ee/packages/den-db run db:migrate, with isolated runtime environment and the disposable branch URL. Exit 0; both Drizzle migration and subsequent schema repairs completed.

Fresh read-only post-verification: 110 application tables (111 including journal), 1281 columns, 472 indexes, and the expected CHECK match the generated final snapshot with zero normalized differences. All application tables remain empty. The original 96 receipts are unchanged; receipt 97 contains the generated SQL hash and original registered timestamp. All 12 Gateway tables exist; the eight old intermediate names are absent.

Production was not modified. The scope is the operator-confirmed empty-intermediate-table recovery, not a populated-data conversion. Older historical hash discrepancies remain untouched. CI/review requirements and the current production state still need to be respected before merging, which can trigger the production migration workflow.

PR #4904, opened 2026-09-11
feat(app): session.stop by id, attributed to the requesting agent (gated remainder of #4893)

yomgui:
Stacked on **#4903** (the un-gated half of #4893). This PR is exactly the part of #4893 that Warden flagged (**2AY-HF9**, previously NLX-3N2: `session.stop` can interrupt any loaded session without confirmation) and therefore the only part that needs a human approving review. Its tree is byte-identical to #4893's head `7bd7f33e0`; #4893 is closed as superseded by #4903 + this PR.

## What it adds (296 lines, one feature)

`session.stop { sessionId }` — the Stop button for any loaded session by id. User decision (2026-09-11 12:35): "an agent should be able to stop a session if a user asks him to do so", same trust level as `session.create` / `session.archive`, **no confirmation**.

- Aborts the turn, cancels subtasks and queued messages, no navigation. Idle → `{ok:true, alreadyIdle:true}`; success → `{ok:true, stopped:true, sessionId, title}`; unknown id → structured error. Not chained into archive.
- **Safeguard is attribution, not a gate**: "Session stopped: <title> — Requested by The agent in "<requester title>" <requester id>" as a toast and as a notification-center entry with a **View** action (new `open-session` notification action), plus the control-action narration.
- Reuses the archive path's interruption machinery (`stopOnly` in `use-session-archive.tsx`); v2 preview disabled with its own reason.
- The `target_working` hint and the `session.archive` description point agents at `session.stop` again ("call session.stop first, then archive"); docs step 2 "Stop when you ask for it" returns.

## What Ben is approving

The product decision that an agent may stop another session's work without a confirmation dialog, with attribution as the only safeguard. Everything else in #4893 merges independently through #4903. If the "who may stop whose work" question is re-opened, this PR is the one to park; #4903 does not depend on it.

## Verification (head `ba8b199e7`)

| Command | Lane | Result |
| --- | --- | --- |
| `bun test --isolate tests/session-archive-agent-contract.test.tsx tests/notification-store.test.ts` (apps/app) | Local | Exit 0: 18 passed (11 contract incl. 3 `session.stop` cases + 7 store) |
| `tsc --noEmit` (apps/app) | Local | Exit 0 |
| `pnpm evals:e2e session-archive-agent-requester` (`OPENWORK_EVAL_DAYTONA=1 OPENWORK_EVAL_REF=ba8b199e7…`) | Fresh Daytona | queued — published in the sticky comment below |

[...]

PR #4903, opened 2026-09-11
fix(app): agents see working sessions and archive warnings go back through the requester's channel (split of #4893, without session.stop)

yomgui:
Split of #4893 (2026-09-11): this PR carries everything Warden can self-clear. The `session.stop` control action (Warden 2AY-HF9, an explicit product decision that needs a human approver) is re-added in a stacked follow-up PR; #4893 is closed as superseded by the pair.

## Problem and RCA

On 2026-09-11 11:27 a session finished its work and, still inside its own assistant turn, called `session.archive` on itself. The app correctly saw the target as working and opened the "This session is still working" confirmation. Three things went wrong around that correct decision:

1. **The requester was discarded.** The OpenCode plugin attaches `origin: { sessionId }` to every bridged command and the control provider hands it to actions as `helpers.origin`, but `session.archive`'s `execute` ignored it. The modal could not tell that requester == target.
2. **Agents had no status.** `session.list_sessions` returned `{sessionId,title,workspace,updatedAt,pinned}` and `session.read` only the transcript; the activity store's idle/thinking/responding/waiting state was never exposed, so an agent could not check before archiving.
3. **The 5 s error was the bridge, not the archive.** `apps/server/src/ui-control.ts` resolves "The OpenWork window did not answer within 5 seconds" when the window has not replied; the archive promise settles only when the dialog closes. A human-gated dialog can never answer in 5 s, so the agent saw a generic transport error while the dialog stayed open.

Related: #4832, #4818, #4825, #4840, draft #4843.

## What changes

Principle (product decision, 2026-09-11): **the warning goes back through the channel the request came from.** The person gets a dialog; an agent gets a structured result in its own conversation, never a dialog.

- **Status.** `session.list_sessions` entries carry `status` (activity store: idle/thinking/responding/waiting/compacting/error) and `working`. `session.read` carries `status` (engine: idle/busy/retry/waiting) and `working`; an unreadable probe reports `working: true`, never a false "safe". Affordance descriptions tell agents to check `working` first.
[...]

PR #4911, opened 2026-09-11
feat(agent): session.send messages a session by id without moving the person's pane

yomgui:
## Problem

A person orchestrating many sessions from one chat saw the active pane jump to whichever session the orchestrating agent acted on, and relays were misrouted (five in one afternoon: twice into the orchestrator itself, once into a session of another workspace, once into a finished session). `composer.send` also answered `Action is disabled` whenever the visible session was mid-turn.

## RCA

The only way to talk to an existing session was `session.open` → `composer.set_text` → `composer.send`, and the composer actions are **focus-bound, not session-bound**:

- `apps/app/src/react-app/domains/session/surface/session-surface.tsx` registers `composer.set_text` / `composer.send` / `composer.stop` with `useControlAction(props.isControlTarget ? action : null)`; `session-page.tsx` passes `isControlTarget={activeWorkbenchPane === "primary" | "secondary"}`. One registration slot per id (`registerAction`, `shell/control/control-provider.tsx`) means the id resolves to whichever pane is focused when the call lands. Neither takes a `sessionId`.
- The navigation was therefore **incidental, not intrinsic**: the engine already accepts a prompt for any session (`POST /session/{id}/prompt_async`, used by `session.create`), but no affordance exposed it, so the agent navigated to make the target the focused composer. A focus change between `set_text` and `send` re-pointed both at the newly focused session; `composer.send` is disabled while the visible session's `transitionState !== "idle"`.
- `effects.ui` (`packages/types/src/openwork-affordance.ts`) is descriptive metadata for the model: nothing in `executeAction` / `executeCommand` reads it. It was also untruthful for `session.create_task` (`ui: none`, but it opens the new session in the focused pane).

`apps/app/tests/control-focus-bound-composer.test.tsx` reproduces the misroute deterministically (focus flips between `set_text` and `send`; B's handler receives the send meant for A).

## Principle

Control-plane actions are addressed by id and headless by default; a UI change is an explicit opt-in (`reveal: true`) or a separate "show me" action (`session.open`, `workbench.session.focus`). Same split as VS Code (`openTextDocument` vs `showTextDocument` + `preserveFocus`), Cursor Cloud Agents (`POST /v0/agents/{id}/followup`, `409 agent_busy`), Slack `chat.postMessage` to a channel id, tmux `send-keys -t`. Details: `docs/features/headless-session-control.md`.

## What changes

- **`session.send { sessionId, text, workspaceId?, reveal? }`** — server-executed command (`openwork-server` provider, same trust as `session.create`) in `apps/server/src/opencode-plugins/openwork-extensions-preview.ts`. Resolves the session across workspaces with the same ownership check as `session.read`, posts `prompt_async` with a generated `messageID`, returns `{ accepted: true, sessionId, workspaceId, title, messageId }`, and never touches the UI. `reveal: true` sends first, then issues one `session.open` through the UI mailbox stamped with the requester `origin`, reporting `revealed`; `effects.ui` is `navigate` only when the reveal actually happened.
- **Queue semantics** (opencode 1.18.18, verified in source and by the spec): `prompt_async` persists the user message and returns 204 at once; `SessionRunState.ensureRunning` makes a call against a running session await the existing run, whose loop picks the new user message up at its next step. Nothing is rejected or lost. Omitting `model` uses the session's stored model.
- **Truthful effects**: `session.create_task` now declares `ui: navigate`. `session.open` / `workbench.session.focus` remain the only navigate actions; `composer.set_text` keeps `ui: focus`.
- **Focus-bound actions say so**: `composer.set_text` / `composer.send` / `composer.stop` and `session.open` descriptions, plus the agent system instruction, point agents at `session.send` for another session.
- `session.rename` / `session.pin` / `session.archive` already act by id without navigating (archive closes the tab only when the target is on screen) — verified, no code change.
- Docs: `cross-chat-memory.mdx` gets "How agents manage other sessions"; `docs/mcp-ui-control-profile.md` no longer tells clients to open a session before reading it.

**Attribution:** `session.send` does not stamp the requester into the recipient's message; the sender's transcript holds the tool call and `messageId`, which is the trail (the recipient answers the user-visible text as written).

Related: #4903 (status/working in `session.read`), #4904 (`session.stop` by id, gated). Independent of both; expect a trivial both-added conflict in `cross-chat-memory.mdx` and `openwork-provider-adapters.ts` whichever lands second.

## Verification

Head: `ca629d183a1974e618aa3597ebc0296e255c0f74`. Lane: **local** for all checks. The PR-lane spec's world (`agentSessionSend`, like its sibling `agentUiContext`) boots a managed `openwork-server` + real engine in-process and does not take Daytona placement, so there is no Daytona variant of it; it runs in CI's `pnpm evals:pr`.

| Command | Result |
| --- | --- |
| `bun --conditions=development test src/opencode-plugins/openwork-extensions-preview.test.ts src/opencode-plugins/openwork-provider-adapters.test.ts` (apps/server) | Exit 0: 37 passed, 0 failed (3 new `session.send` cases + descriptor case) |
| `bun test --isolate tests/control-focus-bound-composer.test.tsx` (apps/app) | Exit 0: 2 passed, 0 failed |
| `pnpm evals:pr specs/session-send-headless-by-id.test.ts` | Exit 0: 1 passed, 0 failed, 0 skipped — evidence in the sticky comment |
| `tsc --noEmit` in apps/server, apps/app | Exit 0 |
| `git diff --check` | Exit 0 |

The spec drives the real engine with a scripted model and a fake window that polls the UI-control mailbox and refuses every command: session 1's `openwork_execute(session.send)` reaches session 2 by id (accepted, `messageId` matches the persisted user message, session 2 answers), session 3 has no messages, the mailbox saw **0 commands**; a second send while session 2 is busy on a held provider stream is accepted, already in the transcript before the model sees it, and answered after the held step; `reveal: true` yields exactly one `session.open` for session 2 with session 1 as origin, after the message was written; an unknown id is refused before anything reaches the engine or the window.

Verdict: **Passed** (local lane).

### DIFF
--- a/apps/app/src/react-app/domains/session/control/session-control-actions.ts
+++ b/apps/app/src/react-app/domains/session/control/session-control-actions.ts
@@ -80,7 +80,8 @@ export function useSessionControlActions(input: UseSessionControlActionsInput) {
   const createTaskControlAction = useMemo<OpenworkControlAction>(() => ({
     id: "session.create_task",
     label: "Create a new task",
-    description: "Create a new session in the selected workspace.",
+    description: "Create a new session in the selected workspace and open it in the person's focused pane. Use session.create to start sessions without changing what is on screen.",
+    effects: { data: "write", ui: "navigate", external: false },
     sideEffect: "mutation",
     disabled: !canCreateTask || !selectedWorkspaceId,
     execute: async () => {
@@ -110,7 +111,7 @@ export function useSessionControlActions(input: UseSessionControlActionsInput) {
   const openSessionControlAction = useMemo<OpenworkControlAction>(() => ({
     id: "session.open",
     label: "Open a session by ID",
-    description: "Focus a visible session or reuse its existing tab. Use list_sessions first to get the session ID.",
+    description: "Show a session to the person: focus it if visible, else open it in the focused pane. Only for when they should see it; use session.read to inspect and session.send to message a session without opening it.",
     effects: { data: "none", ui: "navigate", external: false },
     sideEffect: "navigation",
     requiresArgs: true,
--- a/apps/app/src/react-app/domains/session/surface/session-surface.tsx
+++ b/apps/app/src/react-app/domains/session/surface/session-surface.tsx
@@ -2631,7 +2631,7 @@ export function SessionSurface(props: SessionSurfaceProps) {
   const composerSetTextControlAction = useMemo<OpenworkControlAction>(() => ({
     id: "composer.set_text",
     label: "Type into the composer",
-    description: "Replace the current session draft and type the supplied text visibly.",
+    description: "Replace the draft of the composer the person currently has focused and type the supplied text visibly. Focus-bound: it targets whichever pane is focused when it runs, never a session by id. To message another session use session.send.",
     effects: { data: "none", ui: "focus", external: false },
     sideEffect: "none",
     disabled: archived || !archiveStateKnown || archiveHeld,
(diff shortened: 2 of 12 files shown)


yomgui:

Commit `ca629d183a1974e618aa3597ebc0296e255c0f74` · selected evidence

[Open review report](https://openwork-review-jqvsx0grh-prologe.vercel.app/r/a745d62238b74faeb22050a71d6cb826)

PR #4933, opened 2026-09-12
fix(app): redact JSON and header credentials in crash diagnostics and prove every sink uses it

yomgui:
## Changes
Redact quoted JSON credential keys (case-insensitive; single quotes and scalar values included), preserve keys, and fail closed when the 16,000-character work cutoff removes a closing quote. Handle escaped quotes. Bearer headers already redacted; explicit assertions now cover them.

All three requested sinks already route through the shared formatter: no routing bug or sink production change. Mounted recovery tests inspect revealed DOM, clipboard write arguments, and intercepted optional-report bodies, including identity/message/stack. The separate pre-boot sender (R40) remains out of scope.

Surface(s) touched: redaction / Rows changed: R06, R12, R15, R21, R22, R23 / Blanks introduced: none

- R06, R15: partial→filled.
- R12, R21–R23: blank→filled (bounded helper/component/handler scope).
- Private matrix report untouched and uncommitted.

## Verification — Passed
Local lane; Daytona API key unavailable. All commands exit 0, no skipped tests:
- From `apps/app`: `bun test tests/crash-diagnostics.test.ts` — 44 passed.
- `bun test tests/app-error-boundary.test.tsx` — 16 passed.
- `bun test tests/error-monitoring.test.ts` — 12 passed.
- `pnpm evals:pr specs/crash-redaction.test.ts` — 3 passed; publishes native test assertion output.

Intentional mutations all exit 1: remove JSON replacement → JSON-colon test fails; restore closing-quote-only pattern → both cutoff tests fail; cap-only or cap-before-sanitize → rewritten long-diagnostics test fails; remove quoted-assignment protection → all three sink tests fail. Every mutation restored; final signed head retested.

### DIFF
--- a/apps/app/src/app/lib/crash-diagnostics.ts
+++ b/apps/app/src/app/lib/crash-diagnostics.ts
@@ -1,7 +1,9 @@
 import { sanitizeDiagnosticString } from "./diagnostic-sanitizer";
 
 const URL_PATTERN = /\b[a-z][a-z0-9+.-]*:\/\/[^\s"'<>()]+/gi;
-const SECRET_PAIR_PATTERN = /(token|grant|code|secret|key)=(?:"[^"]*"|'[^']*'|[^&\s"'<>()]+)/gi;
+// Quoted values also end at the work cutoff; an absent closing quote must fail closed.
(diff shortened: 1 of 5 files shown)


yomgui:

Commit `e2f86bc0355ea13e203f84cee1bc74788355d177` · selected evidence

[Open review report](https://openwork-review-jqvsx0grh-prologe.vercel.app/r/7fb51a943da04e9b9c4b6fccf4e8bb4f)

PR #4939, opened 2026-09-12
fix(composer): restore IME input after adding attachment chips

### LINKED ISSUE #4932: [Bug]: Unable to type Chinese (IME broken) in the input area after selecting an image attachment
### Summary

When selecting an image attachment via the file picker before typing any text, Chinese input via IME (Input Method Editor) stops working completely—the IME composition/candidate box fails to appear. However, direct English (ASCII) input continues to work normally.

### To Reproduce

1. Open the chat / message composer.
2. Do **not** type any text yet.
3. Click the attachment button and pick an image via the system file dialog.
4. After the file is selected, try typing in Chinese using an IME (e.g., Pinyin).
5. **Actual Behavior:** No IME candidate window appears, and keystrokes are ignored or fail to compose. Switching to English input works fine.
6. **Expected Behavior:** IME composition should work properly, allowing Chinese input right after attaching a file.

### Expected behavior

The IME composition window should appear normally, and users should be able to type Chinese text after attaching an image.

### Actual behavior

The IME candidate window does not show up, and Chinese characters cannot be entered. Switching to an English input method works fine.

### Screenshots (optional)

_No response_

- OpenWork version：0.18.46
- OS： macOS Sequoia 15.7.1

Workaround: Clicking outside the input field to blur it, and then clicking back inside to refocus, immediately restores IME functionality.
Suspected Cause: Returning from the native OS file picker interrupts Chromium's IME context, preventing compositionstart events from firing properly.
Suggested Fix: In the file input's onChange callback, explicitly trigger an element blur() followed by a deferred focus() (wrapped in setTimeout(..., 50) or requestAnimationFrame) to reconnect the IME state machine.

autostu:
## Summary
- Fixes Chinese (and other CJK) IME input failure in the composer after adding an attachment or inline token chip.
- Appends a trailing whitespace `TextNode(" ")` and places the selection inside it when the paragraph ends with an inline token node, replacing the bare element-level selection.

## Why
- When an attachment chip (`ComposerAttachmentNode`) is appended as the last child of the paragraph, `SyncPlugin` previously fell back to an element-level selection via `lastParagraph.select(childCount, childCount)`.
- Because token chips have `contenteditable="false"` and `canInsertTextAfter(): false`, an element-level selection leaves the cursor directly against an uneditable element with no trailing text node.
- Chromium's IME subsystem requires an editable `TextNode` anchor to start composition; without it, `compositionstart` fails to trigger, completely breaking Chinese/CJK input until the editor is blurred and refocused.

## Issue
- Closes #4932

## Scope
- `apps/app/src/react-app/domains/session/surface/composer/editor.tsx` (`SyncPlugin` selection normalization logic).

## Out of scope
- Any changes to token rendering, keyboard navigation plugins, or file upload handlers.

### Ran
- `pnpm typecheck`

### Result
- pass/fail: pass

## CI status
- pass: expected to pass
- code-related failures: none

## Manual verification
1. Open the composer in a session with an empty draft.
2. Click the attachment button and select an image file via the system file picker.
3. Once the attachment chip appears, immediately type Chinese using an IME (e.g. Pinyin `nihao`).
4. Verified that the IME candidate popup opens normally and Chinese characters commit as expected.
5. Verified that English typing, slash commands, and skills continue to behave normally.

## Evidence
- Verified locally. IME candidate window pops up immediately upon typing after file selection. (Attach screenshot/GIF here if you have one, otherwise write `Verified locally`)

## Risk
- Low. Only affects the fallback selection behavior when the last node in the composer paragraph is an inline token.

## Rollback
- Revert this commit via `git revert <commit-hash>`.

### DIFF
--- a/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
+++ b/apps/app/src/react-app/domains/session/surface/composer/editor.tsx
@@ -990,13 +990,23 @@ function SyncPlugin(props: {
       setPrompt(props.value, props.mentions, props.pastedText, props.attachments);
       // $getRoot().selectEnd() doesn't work when the last node is a
       // token (chip) — Lexical can't position a cursor inside a token,
-      // so the selection collapses to position 0. Use element-level
-      // selection instead: place the cursor *after* the last child of
-      // the last paragraph.
+      // so the selection collapses to position 0. However, a bare
+      // element-level selection after a non-editable token breaks IME
+      // composition (e.g. Chinese/Japanese). Append a trailing space
+      // text node and place the cursor in it to ensure a valid TextNode exists.
       const lastParagraph = $getRoot().getLastChild();
       if ($isElementNode(lastParagraph)) {
-        const childCount = lastParagraph.getChildrenSize();
-        lastParagraph.select(childCount, childCount);
+        const lastChild = lastParagraph.getLastChild();
+        if (isComposerInlineTokenNode(lastChild)) {
+          const spaceNode = $createTextNode(" ");
+          lastParagraph.append(spaceNode);
+          spaceNode.select(1, 1);
+        } else if ($isTextNode(lastChild)) {
+          lastChild.select();
+        } else {
+          const childCount = lastParagraph.getChildrenSize();
+          lastParagraph.select(childCount, childCount);
+        }
       } else {
         $getRoot().selectEnd();
       }


apps/app/src/react-app/domains/session/surface/composer/editor.tsx
jlynn8599-sketch: Review

PR #4958, opened 2026-09-13
fix(chat): never bounce a typed follow-up on a slow complete-history read

reachjalil:
## Summary

After restarting OpenWork, reopening a thread whose turn was still working and typing the next message could show it as pending and then hand the text back to the composer with a failure.

Since #4838 every send awaited the uncapped complete-history read so a follow-up could tell whether it must interrupt a delegated task. On a cold engine after a restart, a long thread's complete history can outlast the renderer's 10 s request timeout (four attempts), and that failed read failed the send itself — the message was pulled from the transcript and returned to the composer. A second defect hid the cause: `unwrapSessionResult` dereferenced `result.response.status` when the SDK settles a thrown fetch with `response: undefined`, so the banner read `Cannot read properties of undefined (reading 'status')` instead of `Request timed out.`.

- `session-history.tsx`: new `readSendHistory()` — cached complete history when the thread has it, otherwise one bounded newest read (24 messages, which always holds the current turn's running delegation). A send never waits on the uncapped read. `ensureFullSnapshot()` stays for branching.
- `session-surface.tsx`: `sendDraft` uses it in place of `ensureFullSnapshot()`.
- `opencode-session-native.ts`: a transport failure keeps its own message when the SDK settles without a response.

## Proof

Verdict: **Passed** at `9f61bd25f`, local lane.

- `pnpm evals:e2e thread-continues-after-restart --local` → `placement: local (--local)`, 3 passed / 0 failed / 0 skipped. New journey `evals/specs/thread-continues-after-restart.e2e.test.ts` (world `evals/worlds/thread-restart.ts`: Den-less desktop, host-booted model witness, caller-owned profile relaunched on the same renderer origin). A running tool is interrupted by a **crash** (SIGKILL of the desktop process group: renderer, embedded server, engine), by an **ordinary quit**, and by a relaunch whose **complete-history read is delayed past the request timeout**. Each case requires the typed continuation to be accepted exactly once, answered, the composer left empty, earlier history intact, and the interrupted tool not replayed.
- Negative control: with only the two app source files reverted, the slow-read case reproduced the report (`draft: "continue"`, bounced, 0 continuations, "The rest of this conversation could not be loaded · Retry") while the crash and quit cases passed.
- Unit: `apps/app/tests/session-history.test.tsx` 45/45 (new: a send reads from cached complete history or one bounded newest read, never the uncapped read; a foreign newest read is refused); `apps/app/tests/opencode-session-native.test.ts` 51/51 (new: a transport failure keeps its own message). `pnpm --filter @openwork/app typecheck` clean; eval channel/boundary ratchets and layer lint clean for the new files.
- Not covered here: a running subagent delegation at restart, an engine unreachable at the archived pre-check, signed-in managed-model startup, Den boundary (Den-less by design).

## Observed, not changed

A SIGKILL of the desktop lost the last minute of renderer `localStorage` (default model reverted to its first committed value, active workspace, drafts, per-session model memory) on the same renderer origin; an ordinary quit lost nothing. The crash case therefore commits preferences with one ordinary quit before the interrupted turn, so it tests thread continuity rather than storage durability.

## UI/UX impact

None — existing UI/UX preserved. Only the failure path changes: the message no longer bounces back into the composer, and an engine transport failure shows its real cause instead of a TypeError.


reachjalil:

Commit `9f61bd25fd810a0c8c3311765598b2d0bb574195` · selected evidence

[Open review report](https://openwork-review-latest-prologe.vercel.app/r/754af0ffc2834111816a17046cf67019)

PR #5038, opened 2026-09-15
perf(dashboard): preserve app views and stabilize loading

reachjalil:
## Summary
Make dashboard MCP Apps feel steadier and load faster: tiles keep their view through refreshes, paint from the last saved version immediately, and no longer fail or go blank during ordinary startup.

**Keep views alive**
- Refresh eligible read-only Apps through the existing document and bridge, preserving guest-owned state when its result handler supports updates. A server-advertised resource digest guards reuse against HTML/CSP changes; leases keep their fixed expiry and every current authorization check.
- Route connection refreshes no longer rebuild workspace client objects for an unchanged server, so tiles stop retiring their own launch with a false "workspace reconnects" error. Genuine address or credential changes still invalidate.
- Transient Connect catalog discovery failures during warm-up retry like other startup errors and keep a still-valid saved view visible. Authority, policy and resource errors still remove it.
- The embedded view stays paintable under the loading overlay (inert until ready), so a slow first paint no longer shows an empty tile.

**Load from cache first**
- Remember width-, workspace-, app- and account-scoped tile geometry and reserve that space during loading.
- Render the last saved version through any matching workspace immediately while the live launch runs.
- Store each distinct App HTML once per scope instead of once per tile. Seven tiles from one App host previously overflowed the 3MB budget and evicted the oldest results; they now fit in roughly 500KB. Eviction keeps the newest results first and legacy caches still parse.

**Do less work**
- Hydrate result caches once per scope, batch persistence, and invalidate memory and pending writes on cache clearing.
- Fetch personal and managed dashboard definitions concurrently; bound launch network work with cancellation and priority handling.

Related to ENG-124 (bounded performance/continuity slice; does not close its policy/portability work).

## UI/UX impact
Dashboard tiles reserve remembered dimensions and show a loading overlay or a small "Updating…" status instead of restarting the visible App. Healthy tile chrome, masonry order, responsive columns, app options and inline chat presentation are unchanged. No height animation or offscreen iframe keep-alive.

## Validation
Passed on the final rebased head:
```sh
./node_modules/.bin/tsc -p apps/app/tsconfig.json --noEmit
./node_modules/.bin/tsc -p apps/server/tsconfig.json --noEmit
git diff --check
```
Focused checks that passed during development (Bun 1.3.4):
- `mcp-app-tile` loading/readiness regression: 2 cases.
- Cache round-trip: seven 480KB-HTML tiles cached under budget and restored; legacy flat cache format still parses.
- Direct Node assertion for scheduler capacity, priority, queued cancellation, rejection and slot release.

Manually verified on a local Electron dev build against a real seven-tile company dashboard: tiles that previously showed "Refresh failed" / blank frames now render, refresh in place, and paint from cache on reopen.

Added focused component/server regressions for notification delivery, document identity, guarded resource changes, authorization negatives, workspace-client identity, cache/geometry isolation, loading and queue behavior. Full suites rely on CI on this head; the existing saved-app-creation journey now asserts snapshot geometry persistence during delayed reload but has not been executed locally. Real-browser live-refresh proof of retained document/input/scroll remains outstanding, and no measured startup-speed or layout-shift figure is claimed.

## Scope and compatibility
- Servers without the advertised refresh guard use the replacement path; cached previews never restore live leases.
- Arbitrary guest state is not restored after a full navigation/reload. The 800px embedded-view ceiling remains.
- Expired-launch recovery (ENG-127) and first-open missing CSS (ENG-13) are not included.
- No database migration, deployment or merge requested.


reachjalil:
## Follow-up: keep saved views through refreshes and startup

Local testing of this branch found two ways tiles went blank or reported **Refresh failed** even though the apps were healthy:

- Route connection refreshes rebuilt workspace client objects for an unchanged server. Tiles read that as losing their workspace and retired their own launch ("This app view is unavailable until its workspace reconnects"). Client identities are now kept stable while the address and credentials are unchanged; genuine changes still invalidate.
- Connect catalog discovery can fail briefly while OpenWork Connect warms up. That is now retried like other transient startup errors, and a still-valid saved view stays visible instead of being cleared. Authority, policy and resource errors still remove the view.
- The embedded view now stays paintable under the loading overlay (inert until ready), so a slow first paint no longer shows an empty tile.

**Preload:** the last saved version and its remembered geometry now render immediately from cache through any matching workspace while the live launch runs, so dashboards paint before the network round-trip.

**Validation:** app typecheck and `git diff --check` passed on the new head; focused tile regression (2 cases) passed earlier on the visibility change. New workspace-client identity tests were added but the suites were not rerun on this head.

PR #5088, opened 2026-09-17
fix(chat): keep thread history loads independent and recoverable

reachjalil:
## Summary
Switching conversations should not discard useful history downloads or make the next conversation depend on a failed read. Opening failures also should not escalate into a larger full-history download.

- Keep opening reads scoped to their original conversation and runtime. A newly selected conversation starts independently; returning adopts an in-flight opening or its fresh cached result.
- Retain at most one abandoned opening per query client for up to five seconds. Excess abandoned work is cancelled; visible split panes remain protected. Cache freshness is two seconds with fifteen-second garbage collection.
- Revoke retained reads when runtime, endpoint, principal, or credentials change, including when the destination is already cached. Late responses cannot replace another conversation or resurrect invalidated data.
- Surface opening failures through the existing error/Retry presentation. Retry requests the bounded opening again. Only recognized missing/unsupported saved-message lookups recover through one latest page; operational failures no longer trigger uncapped history.
- Cancel sibling metadata/message requests when a history attempt fails. Permanent HTTP failures (including 401, 404, and 413), malformed data, ownership errors, and size errors do not retry automatically. Recognized transient transport failures and HTTP 408/429/5xx retain bounded backoff.

No background page chasing, new service, or generic visible-thread timeout is introduced. Active runs, event streams, explicit full-history operations, and existing pagination remain separate.

## UI/UX impact
Existing layout, styling, controls, and navigation are preserved. The existing history error/Retry state now appears for failed or interrupted opening reads instead of leaving an idle loading placeholder or escalating to full history. This reuses the existing presentation (DESIGN P1, P5) and preserves focus behavior (S5). Current-head screenshots below show the failure and recovery states (P10).

## Verification
Final head: `65d34d1a9`.

- `pnpm --package=bun@1.3.14 dlx bun test --cwd apps/app --isolate ./tests/session-history.test.tsx ./tests/opencode-session-native.test.ts ./tests/session-snapshot-owner-flip.test.tsx`: **210 passed, 0 failed**, 2,286 assertions. Covers held/failed source reads, cache reuse, bounded abandonment, split panes, runtime revocation, invalidation/retry, sibling cancellation, and permanent/transient failure classification. Two React act warnings remain.
- `pnpm --filter @openwork/app typecheck`: passed on the final production code.
- `pnpm --dir evals typecheck`: passed, 485 source files.
- `pnpm evals:check-browser`: passed during authoring; the added browser callbacks have not changed since that check.
- `git diff origin/dev...HEAD --check`: passed.
- An earlier focused run failed four owner-switch cases because their fixture expected automatic retry of a permanent 404. The fixture now uses a transient 503 for the retry/owner-switch scenario; separate native and history tests assert that 404 settles without retry. The complete focused selection passed afterward.
- `pnpm --dir evals lint:layers` was also attempted during authoring and reported 54 violations. No clean-base control was run, so their baseline status is not established; no unrelated layer changes are included.

The existing SWITCH-10 journey has an added held-opening witness for destination paint, source-cache reuse, foreign-content isolation, and absence of run-stop requests. It has **not been executed** on the real app. No Daytona, Electron, production-network, or latency result is claimed; required CI remains unchanged.

## UI evidence
Captured from the real appWeb renderer at `65d34d1a9ac6908f871fe90f8c89589eae4c419f`, native v1, 1280 × 900, using 150 synthetic stored notes. A declared HTTP 503 fault on the selected conversation's history GET produces the existing error/Retry state. After removing that fixture fault, clicking Retry retrieves the native stored history over HTTP 200 and shows the latest note. These local screenshots supplement the focused tests; they are not a SWITCH-10, Daytona, Electron, or latency verdict.

Opening history fails with a visible Retry action

Retry recovers the latest saved conversation history

## Scope and remaining limitations
- This fixes failed/cancelled history being mistaken for ongoing loading and avoids automatic amplification of permanent failures. A transport that stays open forever without completing or reporting an error is not automatically detected by this change.
- The five-second timer bounds abandoned cache-warming work only; it is not a loading deadline for a visible conversation.
- Message-count pagination is not a byte budget. No new payload-size cap was added, and no payload limit has been established as the cause of a reported slowdown. Existing HTTP 413/size errors now settle without repeated attempts or an uncapped fallback.
- Shared socket saturation, very large individual messages, and the legacy saved-ID fanout still need runtime-specific diagnosis. This change does not promise transport-level bandwidth priority or preemption of synchronous parsing.


reachjalil:

Commit `ff65451a76d27fe13a4f1df6e8e0ddc9d486c146` · selected evidence

[Open review report](https://openwork-review-latest-prologe.vercel.app/r/3a4a934da2b341839a40e89677ac6796)

PR #5117, opened 2026-09-17
fix(sessions): keep subagent progress live without navigation

reachjalil:
## Problem
A running sub-agent can stop showing fresh activity in its parent thread until the user opens the child chat. The child transcript has no query observer while only the parent is visible, so normal cache garbage collection can remove the baseline needed to apply subsequent progress deltas.

## Change
- Retain one disabled transcript query observer while a session is both active and tracked.
- Share retention across tracking owners and release it on terminal status, tracking cleanup, or workspace disposal.
- Preserve normal transcript garbage collection; add no polling or automatic history requests.
- Add seven regression cases covering navigation-free text/reasoning updates beyond the GC threshold, parent progress derivation, ownership, and cleanup.

Existing labels and layout are unchanged. This is a synchronization-only fix.

## Validation
- Red/green: both navigation-free child progress cases failed before the fix because the baseline was collected; all seven retention cases pass after the fix.
- Focused session sync, lifecycle, permissions, tool parts, cache GC, and sub-agent rendering tests: **223 passed, 0 failed** across seven files.
- App typecheck: passed.
- git diff --check: passed.
- Independent code review: no blockers.

## Scope and limitations
No running-app or desktop E2E was performed. Tests exercise real sync-event handling, TanStack GC, and parent progress derivation, with child tracking already established. This addresses active transcript eviction; missed initial child discovery and reconnect-history recovery are not changed.


reachjalil:
Superseded. The transcript-retention half of this fix landed on `dev` in #5184, which keeps session transcripts cached indefinitely (`gcTime: Infinity`) and leaves cleanup to the tracked-session lifecycle. The other half, keeping a sub-agent's progress live while you stay in the parent conversation, is now in #5304: it starts tracking child sessions as soon as they appear, with tests and an e2e proof.

PR #5128, opened 2026-09-17
fix(den): avoid redundant login-session lookups for MCP requests

reachjalil:
## Problem
MCP transport requests pass through global login-session hydration before their dedicated MCP verifier. MCP bearer tokens are not login-session tokens, so a session-cache miss can cause an unnecessary database query before the actual MCP authentication work.

## Change
- Skip ordinary session hydration only on the mounted `/mcp`, `/mcp/agent`, `/mcp/admin`, and `/mcp/agent/connections/:connectionId` transports.
- Match Hono path normalization, including encoded aliases, without broadly excluding neighboring routes.
- Keep dedicated MCP authentication, scope/resource checks, session/grant liveness, membership checks, and admin authorization unchanged.
- Keep REST session hydration, internal MCP-to-REST calls, and `/v1/mcp/token` outside the exclusion.

Handshake acknowledgments, catalog caching, and client retry behavior are intentionally out of scope. No UI changes.

## Validation
45 focused tests passed in separate Bun processes:
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/bearer-session.test.ts` — 20 passed
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/mcp-membership-revocation.test.ts` — 21 passed
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/mcp-grant-request-liveness.test.ts` — 4 passed

Mounted Hono-route tests use the real session middleware and MCP verifier with mocked persistence. A throwing session-lookup spy verifies that opaque MCP requests do not consult login-session hydration; existing scope and revocation failures remain enforced. Regression cases cover cookies/API keys, encoded route aliases, query strings, and neighboring routes.

Built the required `@openwork/mcp-apps` package after the first route-test attempt reported its missing generated export. All focused suites passed after that prerequisite build. `git diff --check` passed.

No production deployment, live-database test, or production performance benchmark was performed.


OmarMcAdam (changes requested):
Hold for full review
bypassing session validation and checks is not part of our security ethos or culture

PR #5130, opened 2026-09-17
fix(automations): skip session lookups for runner credentials

reachjalil:
## Problem
Automation runner endpoints authenticate dedicated signed bearer credentials, but global session middleware first tries to resolve those same credentials as ordinary user sessions. Malformed, tampered, and expired runner tokens can therefore cause unnecessary session-cache/database misses before the runner verifier rejects them. Cookies or API-key headers can add unrelated authentication work too.

## Change
Skip ordinary session/API-key hydration only for the exact runner-protocol methods and paths: work discovery, notification events, run claim/heartbeat/events/completion, and remote-session command claim/completion. Include HEAD on the GET routes to match Hono dispatch.

The existing runner verifier remains authoritative. Signed valid credentials still require active ownership on every request; invalid credentials retain the existing `401 runner_unauthorized` response. Token minting, runner presence, Automation management, neighboring paths, and unrelated HTTP methods retain ordinary session resolution.

## Relationship to existing work
Complementary to #4143, which handles rejected-runner rate limiting and desktop backoff. This change removes redundant work before route-level authentication and does not modify that PR or its files. It does not stop a client from sending requests, introduce negative-auth caching, or claim zero ingress/CPU cost. Inactive-owner checks remain authoritative rather than cached indefinitely.

## Validation
- `pnpm --filter @openwork/mcp-apps build` — prerequisite build passed.
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/bearer-session.test.ts` — 21 passed.
- `pnpm --filter @openwork-ee/den-api exec bun test --conditions development test/automation-runner-auth.test.ts` — 8 passed.
- `git diff --check` — passed.

Tests cover exact method/path boundaries; zero ordinary session/API-key lookups on runner routes despite attached cookies and API keys; real Hono polling routes rejecting missing, malformed, tampered, and expired credentials without session lookups; valid work discovery; and rejection after the owner becomes inactive. They use the real credential signer/verifier with mocked persistence and owner-service responses, not a live database or production load benchmark.

No UI, scheduling, credential-format, production configuration, or deployment changes.


OmarMcAdam (changes requested):
hold till review of the whole area
bypassing session validation stuff is not part of our security ethos and culture

PR #5156, opened 2026-09-19
fix(server): preserve draining tasks when activity probes fail

reachjalil:
## Problem
A failed or incomplete session-status probe could be interpreted as an idle engine. During configuration rollover, this could retire the old engine while a task was still running; a failed busy check could also select an unsafe in-place reload.

## Changes
- Distinguish complete activity snapshots from unknown or partial results, including HTTP errors, invalid payloads, disconnects, and timeouts.
- Preserve known session ownership and active sessions when probes are incomplete; retire as idle only after a complete empty snapshot or a confirmed dead process.
- Treat unknown reload activity conservatively as busy.
- Keep the existing activity-based inactivity deadline and forced cleanup, rather than allowing unknown activity to retain an engine indefinitely.
- Serialize drain-monitor ticks to prevent overlapping probes and duplicate cleanup.
- Extend the fake-engine regression test with partial multi-workspace failures, recovery, timeout serialization, ownership preservation, and dead-engine cleanup. No UI changes.

## Verification
Final head: `04f94aa0f6771f59080b7027275fdfb087946ce2`, rebased onto `dev` at `f69ca9c0e`.

| Check | Result |
| --- | --- |
| From apps/server: `pnpm exec bun --conditions=development test src/engine-drain-activity-survival.test.ts` | Passed: 1 test, 74 assertions, 0 failures; exit 0 |
| From apps/server: `pnpm run typecheck` | Passed; exit 0 |
| `git diff --check origin/dev...HEAD` | Passed; exit 0 |
| `pnpm --dir evals run typecheck` | Failed; exit 1, 16 diagnostics across 490 checked files |

Evals diagnostics are in unchanged specs: `catalog-fast-mode-wire.test.ts` (unknown model config), `models-analytics-upgrade.e2e.test.ts` (unsupported progressbar target role), and `route-session-list.test.ts` (missing workspace endpoint fields). First diagnostic: `catalog-fast-mode-wire.test.ts(103,11): TS2322`, config of type unknown is not assignable to the model specification. No clean-base control was run, so the failure is unresolved, not classified as pre-existing.

Prior pre-rebase validation supplied with this change: drain/pool/self-heal group 39 passed; cloud-sync/runtime-patch/eviction group 10 passed; server typecheck passed; focused managed-provider credential reload boundary test passed. These are historical checks, not final-head runtime proof.

## Remaining gaps
Draft: verification is incomplete and the evals typecheck is failing. Real-engine rollover and VM reproduction have not been run. No matching final-head testkit evidence report is available to publish; earlier evidence must not be presented as proof of this head. The focused run used Bun 1.3.4 rather than documented 1.3.14. Broad suites and environment setup were deliberately not expanded. No ready-for-review or merge-readiness claim.


benjaminshafii:
What error is this supposed to solve? I expected it to solve the one here. But that's a recording from the running instance.

I decided to merge #5155 which seemed to address this particular issue. If you're addressing a second concern here please let me know

https://github.com/user-attachments/assets/37872f9c-fcca-40b3-ab4d-4d84dfc49db6

benjaminshafii (changes requested):
See comments

benjaminshafii (approved):
See comments

PR #5160, opened 2026-09-19
fix(app): hide model ids while catalog loads

benjaminshafii:
## Summary
- prevent the model picker from flashing an internal gateway model ID during workspace startup
- keep the shared behavior consistent across web and desktop

## Verification
- `pnpm exec bun test apps/app/tests/composer-model-controls.test.ts`
- `git diff --check`


benjaminshafii:
Replaced by a clean PR based on current dev to avoid unrelated historical commits.

PR #5167, opened 2026-09-20
fix(app): preserve first-send continuity

benjaminshafii:
## Summary
- Render the submitted message immediately in the normal conversation layout while session creation runs internally.
- Keep one stable provisional sidebar row and show creation failures inline with retry instead of exposing a lifecycle interstitial.
- Add DESIGN.md P11, Continuity over lifecycle, plus loading-state and review-checklist guidance.

## Verification
- 
- bun test v1.3.8 (b64edcb4) — 22 passed
- Browser evidence covers successful first Send and creation-failure recovery.

## Design
Follows DESIGN.md P11, C3, T2, P5, and P10.

First Send preserves the normal conversation layout

Creation failure remains inline with retry


mdrayhan470598-ops (commented):
Feer fire

